PR #16, opened 2014-09-01
CHAP Response add retransmission.

hanfezh:
CHAP Response add retransmission: transmited 10 times utils received SUCCESS or FAILURE, in 3 seconds interval.

### DIFF
--- a/pppd/chap-new.c
+++ b/pppd/chap-new.c
@@ -73,21 +73,26 @@ static option_t chap_option_list[] = {
 };
 
 /*
+ * These limits apply to challenge and response packets we send.
+ * The +4 is the +1 that we actually need rounded up.
+ */
+#define CHAL_MAX_PKTLEN	(PPP_HDRLEN + CHAP_HDRLEN + 4 + MAX_CHALLENGE_LEN + MAXNAMELEN)
+#define RESP_MAX_PKTLEN	(PPP_HDRLEN + CHAP_HDRLEN + 4 + MAX_RESPONSE_LEN + MAXNAMELEN)
+
+/*
  * Internal state.
  */
 static struct chap_client_state {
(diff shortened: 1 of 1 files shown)


Neustradamus:
@nomis: What do you think about this PR?

nomis:
I've explained why retransmitting is a bad idea here: https://github.com/paulusmack/ppp/issues/87#issuecomment-595794742

This is a challenge/response authentication method and it is normal for the entire PPP session to be transparently transferred to a different LNS.

Neustradamus:
@hanfezh: What do you think about @nomis comment?

At the same time, can you rebase and squash your commits and sign-off please?

paulusmack:
> I've explained why retransmitting is a bad idea here: [#87 (comment)](https://github.com/paulusmack/ppp/issues/87#issuecomment-595794742)
> 
> This is a challenge/response authentication method and it is normal for the entire PPP session to be transparently transferred to a different LNS.

RFC1334 says "the authenticator MUST allow repeated Response packets after completing the Authentication phase". To be sure, the peer could be buggy, and we can provide ways to work around that, but we don't have to assume that the peer is buggy.

Neustradamus:
@hanfezh: What do you think about @nomis and @paulusmack comments?

Neustradamus:
@hanfezh: What do you think about @nomis and @paulusmack comments?

Neustradamus:
@hanfezh: Soon two years after my latest comment here.

What do you think?

Thanks in advance.

hanfezh:
> @hanfezh: Soon two years after my latest comment here.
> 
> What do you think?
> 
> Thanks in advance.

Hi @Neustradamus I think we can close this PR, since it's about 9 years and no one else has feedback.

PR #22, opened 2014-09-19
Add pkg-config support

lkundrak:
Hi,

review and eventually pull this. It will make it simpler for us to determine the location where to put the NetworkManger pppd plugin.

Thank you!
Lubo


lkundrak:
Replaced by #47

PR #24, opened 2014-09-25
Support using OpenSSL to provide DES functions.

agl:
The current code supports two sources of DES: libdes (very old) and
libcrypt (still quite common, thanks to POSIX). OpenSSL implements
libdes compatibility, but only as deprecated functions which may not be
availible.

This change adds an option to use the "modern" (13 year old) OpenSSL DES
functions.


pprindeville:
Looks good to me!

paulusmack:
Superseded by 3c7b86229f7b as far as I can see.

PR #25, opened 2014-09-30
PPPD: Fix crash when printing pppol2tp option

dedeckeh:
PPPD crashes (SEGV) when the dump or dryrun options are specified and an option
is internally defined as "o_special" with an option flag of "OPT_A2STRVAL".
As the option value is not saved when the parameter is processed and is later
referenced when the option is printed a crash occurs.
The crash has been observed when using xl2tpd and the l2tp_ppp kernel module.

Additionally the arg argument was missing to the printer call in the print_option
utility when the option flag OPT_A2STRVAL is set.

### DIFF
--- a/pppd/options.c
+++ b/pppd/options.c
@@ -974,7 +974,7 @@ print_option(opt, mainopt, printer, arg)
 			p = (char *) opt->addr2;
 			if ((opt->flags & OPT_STATIC) == 0)
 				p = *(char **)p;
-			printer("%q", p);
+			printer(arg, "%q", p);
 		} else if (opt->flags & OPT_A2LIST) {
 			struct option_value *ovp;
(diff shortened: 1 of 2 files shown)


paulusmack:
The code fixes look fine. This should be two separate commits since there are two separate bugs.

The commit message is misleading because it implies that the crash would occur whenever an o_special/A2STRVAL option is used along with dump or dryrun, which is not the case - options such as +ua or deflate work just fine. The bug is actually in the pppol2tp code - the setdevname_pppol2tp() function is not setting the string value for the option, which it is required to do since the A2STRVAL flag is set.

Neustradamus:
Merged: https://github.com/paulusmack/ppp/pull/72

PR #30, opened 2014-12-20
pppd: Fix sign-extension when displaying bytes in octal

pprindeville:
print_string() displays characters as \\%.03o but without first
casting it from "char" to "unsigned char" so it gets sign-extended
to an int. This causes output like \37777777630 instead of \230.

Signed-off-by: Philip A. Prindeville <philipp@redfish-solutions.com>

### DIFF
--- a/pppd/utils.c
+++ b/pppd/utils.c
@@ -625,7 +625,7 @@ print_string(p, len, printer, arg)
 		printer(arg, "\\t");
 		break;
 	    default:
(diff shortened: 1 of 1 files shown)


pprindeville:
Related issue: https://github.com/paulusmack/ppp/issues/43

pprindeville:
Thanks, @paulusmack.  Hard to believe we've been banging on PPP since what... 1992?

Issue #32, opened 2015-01-14
segfault on pppd parameters dump when using plugin pppol2tp


shooorf:
There is a bug in pppol2tp plugin (in this example - via xl2tpd, Ubuntu 14.10 server, linux core 3.16.0-29-generic):

Jan 14 00:56:27 iorange pppd[20230]: Plugin pppol2tp.so loaded.
Jan 14 00:56:27 iorange pppd[20230]: pppd options in effect:
Jan 14 00:56:27 iorange pppd[20230]: debug debug#011#011# (from /etc/ppp/options.xl2tpd)
Jan 14 00:56:27 iorange pppd[20230]: nodetach#011#011# (from command line)
Jan 14 00:56:27 iorange pppd[20230]: logfile /var/log/ppp0-dbg.log#011#011# (from /etc/ppp/options.xl2tpd)
Jan 14 00:56:27 iorange pppd[20230]: unit 0#011#011# (from /etc/ppp/options.xl2tpd)
Jan 14 00:56:27 iorange pppd[20230]: dump#011#011# (from /etc/ppp/options.xl2tpd)
Jan 14 00:56:27 iorange pppd[20230]: plugin pppol2tp.so#011#011# (from command line)
Jan 14 00:56:27 iorange pppd[20230]: noauth#011#011# (from /etc/ppp/options.xl2tpd)
Jan 14 00:56:27 iorange pppd[20230]: name shooorf#011#011# (from /etc/ppp/options.xl2tpd)
Jan 14 00:56:27 iorange pppd[20230]: remotename l2tp#011#011# (from /etc/ppp/options.xl2tpd)
Jan 14 00:56:27 iorange kernel: [266504.060596] pppd[20230]: segfault at 0 ip 0000000000424960 sp 00007fff609ac230 error 4 in pppd[400000+5a000]

Bug could be reproduced only when ppp option "dump" is active and pppol2tp is used.
Seg fault caused by function "print_option" in options.c

Stack trace

/usr/sbin/pppd(vslprintf+0x6a)[0x427454]
/usr/sbin/pppd(pr_log+0xba)[0x428ab7]
/usr/sbin/pppd[0x425603]
/usr/sbin/pppd[0x42576a]
/usr/sbin/pppd(print_options+0x87)[0x425814]
/usr/sbin/pppd(main+0x319)[0x40ac55]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf5)[0x7f52ba596ec5]
[...]

PR #37, opened 2015-04-04
ppp: fix pppol2tp option printing

nlhintz:
PPPD crashes (SEGV) when the 'dump' or 'dryrun' options are specified and
an option defined internally as "o_special" with an option flag of
"OPT_A2STRVAL" is used.  The crash occurs because the option value is not
saved when the parameter is processed, but is then referenced when printed.
Additionally, the "arg" parameter is missing from the call to the "printer"
utility.  This was encountered using xl2tpd and the l2tp_ppp kernel module;
the option PPPD crashes on is "pppol2tp 8".

Modify pppd to correctly save the option value, and to call the printer
utility with the correct parameters.

Signed-off-by: Nathan Hintz <nlhintz@hotmail.com>

---

This patch was originally submitted to OpenWRT in:

	https://lists.openwrt.org/pipermail/openwrt-devel/2014-July/026794.html

and was recently accepted in the following changesets:

	https://dev.openwrt.org/changeset/45263
	https://dev.openwrt.org/changeset/45264

Note: This patch resolves issue #32 (https://github.com/paulusmack/ppp/issues/32)

### DIFF
--- a/pppd/ipcp.c
+++ b/pppd/ipcp.c
@@ -175,10 +175,10 @@ static option_t ipcp_option_list[] = {
     { "noipdefault", o_bool, &disable_defaultip,
       "Don't use name for default IP adrs", 1 },
 
-    { "ms-dns", 1, (void *)setdnsaddr,
-      "DNS address for the peer's use" },
-    { "ms-wins", 1, (void *)setwinsaddr,
-      "Nameserver for SMB over TCP/IP for peer" },
+    { "ms-dns", o_special, (void *)setdnsaddr,
+      "DNS address for the peer's use", OPT_A2LIST },
+    { "ms-wins", o_special, (void *)setwinsaddr,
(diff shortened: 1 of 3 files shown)


nlhintz:
Pull request https://github.com/paulusmack/ppp/pull/25 is a partial duplicate to this request; that is, pull request #25 is missing some of the changes.

nlhintz:
This pull request and #25 provide the same changes; although #25 doesn't include the original authorship.

paulusmack:
As I just commented on pull request #25, I want this split into two commits and with a better patch description that identifies that the o_special/A2STRVAL bug is in the pppol2tp code not the option parsing code.

paulusmack:
The ms_dns/wins commit looks fine.

Neustradamus:
Merged: https://github.com/paulusmack/ppp/pull/72

PR #39, opened 2015-07-04
pppd: Add client CHAP authentication timeout

nomis:
If CHAP authentication is required with the peer but this is never
completed (either because the server never sends the challenge or
because the client doesn't receive the outcome) then the client
will wait forever, relying on the server to terminate the connection.

There are options for server side retries but a client side timeout
option is required to prevent the client from getting stuck if the
server won't terminate the connection. This is defaulted to 60 seconds.

### DIFF
--- a/pppd/chap-new.c
+++ b/pppd/chap-new.c
@@ -55,20 +55,23 @@ int (*chap_verify_hook)(char *name, char *ourname, int id,
 /*
  * Option variables.
  */
-int chap_timeout_time = 3;
+int chap_server_timeout_time = 3;
 int chap_max_transmits = 10;
 int chap_rechallenge_time = 0;
+int chap_client_timeout_time = 60;
 
 /*
  * Command-line options.
  */
 static option_t chap_option_list[] = {
-	{ "chap-restart", o_int, &chap_timeout_time,
-	  "Set timeout for CHAP", OPT_PRIO },
+	{ "chap-restart", o_int, &chap_server_timeout_time,
+	  "Set timeout for CHAP (as server)", OPT_PRIO },
 	{ "chap-max-challenge", o_int, &chap_max_transmits,
 	  "Set max #xmits for challenge", OPT_PRIO },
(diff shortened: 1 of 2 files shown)


lp0:
Updated with a fix for ```chap_lowerdown``` to read the ```cs->flags``` before clearing them.

lp0:
This fixes issue #87.

There are no timeout issues with the other authentication types.

paulusmack:
Hmmm, when the timeout occurs, I think it would be better to resend the CHAP-Response packet, instead of just giving up.

nomis:
>  I think it would be better to resend the CHAP-Response packet, instead of just giving up.

The client can't send a `CHAP Response` if it never received a `CHAP Challenge`. The server may not resend this or the retries may also be lost, so the client must timeout in this scenario.

The server may not accept a duplicate `CHAP Response` a full 60 seconds later, especially when it has already sent a `CHAP Success`.  This would add an assumption about how the other side of the link will behave. It looks like `ppp` would do this, but RFC 1334 does not require it. Only PAP explicitly specifies that retries will occur indefinitely.

nomis:
The server is also allowed to send multiple challenges, so the client may be retrying the response to an earlier challenge if there are two consecutive challenges and multiple lost packets.

In the UK the majority of DSL lines use BT Wholesale connections that perform the CHAP authentication using RADIUS and then hand-off to the PPP via L2TP to a remote server which then (usually) starts LCP again and does another CHAP authentication.

It's possible for the remote server to simply do a second `CHAP Challenge` (which is also lost). In that scenario the client would be retransmitting the wrong `CHAP Response` because the challenge value will be different.

It's possible for the remote server to start IPCP/IPV6CP without restarting LCP. In that scenario the client would be retransmitting the `CHAP Response` to a device that may not have the original challenge value and may not respond appropriately.

nomis:
Here's a comment on the implementation used by [FireBrick](https://www.firebrick.co.uk/):
> The system handles failure of a chap packet at any stage quite well with the appropriate resend. A whole minute may be too long though.

PR #41, opened 2015-08-04
add a sample ipv6-up script to add/replace default route for IPv6

rogers0:
This script is modified from a post in Debian Bug#477245
- http://bugs.debian.org/477245#45

That bug was considered to be sovled by upstream, that why it's posted here.

Signed-off-by: Roger Shimizu rogershimizu@gmail.com

### DIFF
--- a/scripts/ipv6-up.defaultroute
+++ b/scripts/ipv6-up.defaultroute
@@ -1,0 +1,29 @@
+#!/bin/sh
+# This script shows the example how to replace IPv6's default route when
+# PPP connection is estabished.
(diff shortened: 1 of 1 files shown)


sthibaul:
Err, why not just implementing a defaultroute6 option? That way be way more logical compared and convenient, compared with the defaultroute option.

rogers0:
@sthibaul maybe better to add defaultroute6 option, but after all these years, the issue is still pending.
adding this workaround is enough for most cases.

paulusmack:
Superseded by #77

PR #47, opened 2015-10-31
 Add pkg-config support

lkundrak:
Hii,

plese review this and consider merging it. It will make it simpler for us to determine the location where to put the NetworkManger pppd plugin.

Thank you!
Lubo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@
 *.so
 *.a
 *.cat8
+*.pc
 Makefile
--- a/chat/Makefile.linux
+++ b/chat/Makefile.linux
@@ -1,6 +1,6 @@
 #	$Id: Makefile.linux,v 1.15 2006/06/04 05:07:46 paulus Exp $
 
-DESTDIR = $(INSTROOT)@DESTDIR@
+DESTDIR = @DESTDIR@
 BINDIR = $(DESTDIR)/sbin
 MANDIR = $(DESTDIR)/share/man/man8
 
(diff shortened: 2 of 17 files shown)


paulusmack:
This has several problems in the "Determine some more variables at configure time" commit.

- In pppd/plugins/pppoatm/Makefile.linux you have changed the mode for the plugin from 4550 to 455, which is wrong. Arguably the 4000 (setuid) bit doesn't really mean anything for a plugin, but if you want to change that, it's a separate change that should be a separate patch, and then the mode should be 555 not 455.

- In pppd/plugins/pppol2tp/Makefile.linux you change the symbol for the plugin directory from LIBDIR to PLUGINDIR but then still install the plugin to LIBDIR.

- In pppd/plugins/rp-pppoe/Makefile.linux you set PPPDVERSION to the version number but then use $(VERSION) in defining PLUGINDIR.

lkundrak:
Thanks. Addressed the review.

I've also added one more change. That one is required so that the .pc file is "relocatable" -- the prefix is expanded later and thus can be overridden. That is useful when whoever consumes the file chooses an alternate installation prefix after the build is done (e.g. what autotools' "make distcheck" does when checking sanity of the installation).

Neustradamus:
@lkundrak: Can you rebase your PR?

Neustradamus:
@lkundrak: Can you rebase your PR?
Thanks in advance.

Neustradamus:
@lkundrak: Can you rebase your PR?

enaess:
I'd also vouch for this as maintenance of network-manager-sstp and sstp-client would be easier.

Neustradamus:
@enaess: I think the best solution, badly, is to create a new PR... :/

enaess:
@Neustradamus I'll have a look

vincele:
Hello, should this PR be closed now that pkg-config has been merged ? c.f. #270

enaess:
@lkundrak @paulusmack could you please close this PR?

PR #49, opened 2016-02-22
Possibility to rename interface in pre-up script from vyatta-ppp

voroniys:


paulusmack:
The commit description is inadequate. It should explain what this does and why we want it.

Also, it's not clear whether Stanislav Voroniy is the original author of the patch, and there is no Signed-off-by line to indicate that the original author has the right to contribute the patch.

Finally, I get these errors when compiling:

ipcp.c: In function ‘ipcp_up’:
ipcp.c:1905:12: warning: implicit declaration of function ‘if_nametoindex’ [-Wimplicit-function-declaration]
  ifindex = if_nametoindex(ifname);
            ^
ipcp.c:1911:7: warning: implicit declaration of function ‘if_indextoname’ [-Wimplicit-function-declaration]
  if (!if_indextoname(ifindex, ifname)) {
       ^

Are these functions available on Solaris?

Neustradamus:
@voroniys: Can you reply to @paulusmack?

And, currently it is not solved in master?

PR #55, opened 2016-05-09
Added /etc/ppp/ipv6-pre-up script

Tydus:
Vyatta based distros (EdgeOS, VyOS, etc.) are using ip-pre-up.d scripts to rename the ppp device to their format.
However the ip-pre-up script is only invoked by IPCP code, but not IPV6CP. Thus the vyatta scripts has no chance to cleanly do this.

### DIFF
--- a/pppd/ipv6cp.c
+++ b/pppd/ipv6cp.c
@@ -1146,6 +1146,7 @@ ipv6_demand_conf(u)
 	return 0;
     if (!sif6addr(u, wo->ourid, wo->hisid))
 	return 0;
+    ipv6cp_script(_PATH_IPV6PREUP, 1);
 #if !defined(__linux__) && !(defined(SVR4) && (defined(SNI) || defined(__USLC__)))
     if (!sifup(u))
 	return 0;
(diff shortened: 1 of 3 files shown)


novag:
@paulusmack Are there any objections or suggestions for improvement?

marcan:
Just a ping, is this PR good? It seems like a simple and logical change and, once it trickles down to Debian and Vyatta, will let people with IPv6-only PPPoE sessions (i.e. most of Japan) use those devices. Right now we have to use a manually patched build of pppd.

paulusmack:
The code change itself looks fine. It needs a signed-off-by from the original author. The commit message is inadequate. Please put a one-line summary starting with "pppd:" as the first line and then a description of what's being changed and why (doesn't need to be elaborate, just a sentence or two).

Tydus:
Thanks. I amended my commit. Does it look good?

paulusmack:
I committed this, and then reverted it, because it doesn't compile. ipv6cp_script() does not have the "wait" argument (and never has). It seems that you didn't even compile-test this.

PR #58, opened 2016-07-26
Fix bug in netif_set_mtu for Solaris

Low-power:
The argument mtu in this function is never used before.
This change makes for example 'pppd ... mtu 1400' works on Solaris.


paulusmack:
This needs a Signed-off-by line, plus a real name as the author.

Low-power:
I haven't used the signed commit before; can you tell me how to add the Signed-off-by line to an existing commit?

Low-power:
Replaced by #103.

PR #60, opened 2016-08-10
Add ppp_defs.h include to magic.[ch]

snickl:
magic.[ch] make use of the __P() macro provided by ppp_defs.h, therefore
they should include this header.

This omission did not cause compilation errors with glibc because it
also provides this legacy macro in the non-standard header sys/cdefs.h,
deprecated by musl.

Alternatively, I'd offer to prepare a patch removing the __P() macro from the entire ppp code base if that is acceptable.

### DIFF
--- a/pppd/magic.c
+++ b/pppd/magic.c
@@ -47,7 +47,7 @@
 #include <unistd.h>
 #include <sys/types.h>
 #include <sys/time.h>
-
+#include "net/ppp_defs.h"
 #include "pppd.h"
 #include "magic.h"
 
--- a/pppd/magic.h
+++ b/pppd/magic.h
@@ -42,6 +42,8 @@
  * $Id: magic.h,v 1.5 2003/06/11 23:56:26 paulus Exp $
  */
 
+#include "net/ppp_defs.h"
+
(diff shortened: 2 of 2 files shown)


paulusmack:
Since it's not causing compilation errors, and it is time we took out __P, I'm not inclined to merge this.

snickl:
I see how it's not entirely clear from my original comment, but this is one of the few things that __do__ prevent pppd from building against musl-libc.

However I fully agree that shedding __P is the better approach, on account of which I'd like to point you at https://github.com/snickl/lpppd/commit/3f764171541f85bbb6079ff74d87605fd7cf3d71 :)

Btw, would you be willing to merge changes from K&R-style function definitions to ANSI-style?

Neustradamus:
@snickl: It is solved in master?

snickl:
It's solved in snickl/lpppd, otherwise it seems ongoing with #164

Neustradamus:
@snickl: Have you tested PR164? Compilation...
It works?

snickl:
Works in my fork

neheb:
That PR is ready. The mentioned failure is unrelated.

PR #61, opened 2016-08-10
Add sys/types.h include to {net,linux}/ppp_defs.h

snickl:
struct ppp_idle makes use of the type time_t, which is defined in sys/types.h.
Therefore, ppp_defs.h should include this header.

This issue goes unnoticed with glibc, but causes a compilation error with musl
libc.

### DIFF
--- a/include/linux/ppp_defs.h
+++ b/include/linux/ppp_defs.h
@@ -50,6 +50,8 @@
 #ifndef _PPP_DEFS_H_
 #define _PPP_DEFS_H_
 
+#include <sys/types.h>
+
 /*
  * The basic PPP frame.
  */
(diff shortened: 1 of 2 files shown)


paulusmack:
It's not clear to me that <sys/types.h> is the canonical thing to include to get time_t. Wouldn't that be <time.h>?

snickl:
Hi Paul, very excited to see you working on pppd again!

I guess either will be fine, <time.h> seems a bit more common/general, but since the header itself only uses the time_t type, <sys/types.h> is sufficient. The standard [supports it](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html), and usually <time.h> includes <sys/types.h> for these types.

As an aside: What about unifying include/net/ppp_defs.h and include/linux/ppp_defs.h, possibly using C99 fixed integer types? It seems the only real difference is the ratio member of struct compstat.

PR #64, opened 2016-08-10
Fix multiple inclusion errors for ethernet structure

snickl:
Inclusion for the low-level ethernet headers in the rp-pppoe plugin is
quite a mess and does not work for musl libc as-is.  Since this code is
Linux-only and uses a static config.h, the patch gets rid of the
pointless flexibility, replacing it by a single inclusion of
linux/if_pppox.h in pppoe.h.

Note that linux/if_pppox.h was included unconditionally before and pulls
in the needed ethernet headers. This pragmatic approach may not honor
the include-what-you-use principle, but seems justified in this case.

### DIFF
--- a/pppd/plugins/rp-pppoe/config.h
+++ b/pppd/plugins/rp-pppoe/config.h
@@ -59,26 +59,14 @@
 /* Define if you have the <getopt.h> header file.  */
 #define HAVE_GETOPT_H 1
 
-/* Define if you have the <linux/if_ether.h> header file.  */
-#define HAVE_LINUX_IF_ETHER_H 1
-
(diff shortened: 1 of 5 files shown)


paulusmack:
Looks to me like the same effect could be achieved simply by changing config.h.

Also needs a signed-off-by.

Neustradamus:
@snickl: It is solved in master?

snickl:
It's solved in snickl/lpppd

Neustradamus:
@snickl: Can you create a new PR with the remark of @paulusmack?

snickl:
I don't think upstream is interested

PR #65, opened 2016-08-10
Fix up sys-linux.c ethernet/ppp includes for both glibc and musl

snickl:


pprindeville:
How did you test this?  What versions of glibc and MUSL have you compiled against?

snickl:
Hi, I haven't looked at this in a while but I must have tested this patch as part of my fork https://github.com/snickl/lpppd against musl 1.1.15 and glibc 2.24, as well as whatever uClibc was in buildroot around November last year.

PR #66, opened 2016-08-11
linux: support non-standard baudrates

schouleu:
When an unsupported baudrate is specify, we can force it to the kernel
by using the TCSETS2 ioctl with the BOTHER cflag.

TCSETS2 and struct termios2 are available in asm/termbits.h and
asm/ioctls.h but including these headers conflict with termios.h which
is needed by sys-linux.c.
For that reason, we put the force_baudrate function in a dedicated file.

Signed-off-by: Sylvain Chouleur sylvain.chouleur@gmail.com

### DIFF
--- a/pppd/Makefile.linux
+++ b/pppd/Makefile.linux
@@ -13,7 +13,8 @@ TARGETS = pppd
 
 PPPDSRCS = main.c magic.c fsm.c lcp.c ipcp.c upap.c chap-new.c md5.c ccp.c \
 	   ecp.c ipxcp.c auth.c options.c sys-linux.c md4.c chap_ms.c \
-	   demand.c utils.c tty.c eap.c chap-md5.c session.c
+	   demand.c utils.c tty.c eap.c chap-md5.c session.c \
+	   sys-linux-termios2.c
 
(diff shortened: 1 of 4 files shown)


Neustradamus:
@schouleu: Can you update your PR?

Neustradamus:
@schouleu: Can you update your PR?

pali:
I created a new pull request which address feature request to support arbitrary baud rate on Linux system: https://github.com/ppp-project/ppp/pull/278

> This won't compile on all architectures. For example, on powerpc the struct termios has the c_ispeed and c_ospeed members and there is no struct termios2.

I proved completely new implementation which does not have this issue yet. On architecture where ioctls are not provided, this new code is not compiled at all (it is hidden behind #ifdef).

> Does glibc really provide no official interface that hides the termios2 thing?

glibc still, even after 4 years since your comment, does not provide any interface for termios2. And does not look like this will change. Also `stty` program from coreutils does not provide a way for changing baudrate to arbitrary value.

pali:
Pull request #278 which replaces this one was now merged, so this pull request can be closed.

paulusmack:
Superseded by pull request 278.

Neustradamus:
Thanks to @pali and @paulusmack!

PR #72, opened 2016-12-10
Pppd print changes

nlhintz:
Paul:

I've updated the pull request based on your comments to the previous request.

Thank you,

Nathan


Neustradamus:
@nlhintz, @dedeckeh, @shooorf: Do you confirm that now it is perfect?

PR #69, opened 2016-10-13
ppp: fix building with linux-4.8

jackiehjm:
Fix a build error when using the linux-4.8 headers that results in:

In file included from pppoe.h:87:0,
                 from plugin.c:29:
../usr/include/netinet/in.h:211:8: note: originally defined here
 struct in6_addr
        ^~~~~~~~
In file included from ../usr/include/linux/if_pppol2tp.h:20:0,
                 from ../usr/include/linux/if_pppox.h:26,
                 from plugin.c:52:
../usr/include/linux/in6.h:49:8: error: redefinition of 'struct sockaddr_in6'
 struct sockaddr_in6 {
        ^~~~~~~~~~~~

Signed-off-by: Jackie Huang jackie.huang@windriver.com

PR #74, opened 2017-01-09
pppoe: include netinet/in.h before linux/in.h

lkundrak:
This fixes builds with newer kernels. Basically, <netinet/in.h> needs to be
included before <linux/in.h> otherwise the earlier, unaware of the latter,
tries to redefine symbols and structures. Also, <linux/if_pppox.h> doesn't work
alone anymore, since it pulls the headers in the wrong order, so we better
include <netinet/in.h> early.


lkundrak:
This seems to duplicat #69, but that one reportedly breaks on older kernels. I've tested this one with 2.6.32.

PR #79, opened 2017-03-24
Use systemd's sd_notify with option up_sdnotify

### LINKED ISSUE #70: Add systemd daemon support
nodetach and updetach are helpful, but it is difficult for systemd to know when pppd is up enough to start services that depend on it. Explicit systemd support would make dependencies easier to deal with.

cognifloyd:
This adds an up_sdnotify option so that systemd services of
Type=notify can have pppd send the READY=1 signal to systemd
once a network protocol (typically IP) is up.

To use up_sdnotify, pppd must be compiled with SYSTEMD=y.

Resolves #70

### DIFF
--- a/pppd/Makefile.linux
+++ b/pppd/Makefile.linux
@@ -60,6 +60,11 @@ HAVE_MULTILINK=y
 # Linux distributions: Please leave TDB ENABLED in your builds.
 USE_TDB=y
 
+# Uncomment the next line to enable Type=notify services in systemd
+# If enabled, and the user sets the up_sdnotify option, then
+# pppd will not detach and will notify systemd when up.
+#SYSTEMD=y
+
 HAS_SHADOW=y
 #USE_PAM=y
 HAVE_INET6=y
@@ -168,6 +173,11 @@ endif
 ifdef LOCKLIB
 LIBS     += -llock
 CFLAGS   += -DLOCKLIB=1
+endif
+
+ifdef SYSTEMD
+LIBS += -lsystemd
+CFLAGS   += -DSYSTEMD=1
 endif
 
 ifdef PLUGIN
--- a/pppd/auth.c
+++ b/pppd/auth.c
@@ -100,6 +100,10 @@
 #endif
 #include <time.h>
 
+#ifdef SYSTEMD
+#include <systemd/sd-daemon.h>
+#endif
+
 #include "pppd.h"
 #include "fsm.h"
 #include "lcp.h"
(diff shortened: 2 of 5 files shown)


paulusmack:
Mostly looks good.  Two comments:

1. Could you please add a Signed-off-by to the commit message?

2. What happens if a non-privileged user running pppd specifies the up_sdnotify option? Does sd_notify do anything bad if it is called when pppd was not run from systemd? Is there any way that this could let a non-privileged user attack systemd and possibly gain privileges somehow?

cognifloyd:
1. I'll add that sometime next week.

2. Systemd services can be running as --system or --user services. I have
not tested ruining ppp as a non root user, but it shouldn't matter how the
service is running (system vs user) or which user it is running as. Also,
systemd is very strict about listening for and acting on notifications from
processes that are allowed to do so. If systemd didn't start the process,
it ignores any notifications unless the signaling process was started by a
service that systemd is monitoring (directly or indirectly, such as a grand
child process in the same cgroup as a process that systemd started) AND
that service is Type=notify, AND that service is explicitly configured to
allow other processes to send a notification on behalf of that service by
setting NotifyAccess=all.

Side note: when a child process seems the notification instead of the
actual service process, there are some nasty race conditions that can cause
systemd to ignore the notification even if that child process is allowed to
send (eg the process excited before systems could look up its parents and
cgroup). Plus, if you run systemd-notify, a command line utility that seems
the notification, from within ip-up out one of ip-up's children, then the
notification never succeeds. First there's the race condition of those
scripts being too short lived. Then, ip-up's environment is so effectively
sanitized that even if I somehow set the environment variable that
systemd-notify expects, it still never succeeds in associating the
notification with service.

So, systemd is really good at ignoring notifications unless it expects
them, and the only really reliable way to send those signals is from the
actual non-daemonizing service process.

Also, the socket used is defined in an environment variable provided and
deleted by systemd, allowing system and user services to use a different
socket. I really don't think there's any way to use that socket (even via
the sd_notify api of their library) to gain elevated privileges.

On Sat, Mar 25, 2017, 21:57 Paul Mackerras <notifications@github.com> wrote:

Mostly looks good. Two comments:

   1.

   Could you please add a Signed-off-by to the commit message?
   2.

   What happens if a non-privileged user running pppd specifies the
   up_sdnotify option? Does sd_notify do anything bad if it is called when
   pppd was not run from systemd? Is there any way that this could let a
   non-privileged user attack systemd and possibly gain privileges somehow?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/paulusmack/ppp/pull/79#issuecomment-289254687>, or mute
the thread

paulusmack:
OK, sounds fair.

Please add a paragraph to the commit message explaining that you have made the new option a non-privileged option because ... (e.g., because a non-privileged user running pppd can't cause any disruption by setting it, and briefly why that is so).

cognifloyd:
OK. I finally got around to fixing the commit message.

PR #80, opened 2017-03-26
Include CLASS(25) attribute everywhere

anphsw:
Some radius servers will not accept accounting packets without CLASS attribute, this patch will include it.


paulusmack:
The patch itself looks OK. The commit description is quite inadequate. Even the headline is misleading -- "everywhere" - really? Throughout every program in the ppp package? And what's mikrotik? Don't answer in the comments, write an actual informative commit message, and include a signed-off-by.

Neustradamus:
A new PR has been done and merged: https://github.com/paulusmack/ppp/pull/100

PR #88, opened 2017-09-24
chat regex/glob pattern matching

elrafoon:
Hello,

please could you review and eventually merge my patches enabling regex and glob pattern matching in chat program?

I needed it to match multiple AT+CREG responses in and embedded project I'm working on.

I also re-indented the chat.c source, because I was unable to properly understand the code in original layout.

### DIFF
--- a/chat/chat.8
+++ b/chat/chat.8
@@ -53,6 +53,10 @@ to \fIstderr\fR.
 Enables environment variable substitution within chat scripts using the
 standard \fI$xxx\fR syntax.
 .TP
+.B \-R
+Enable posix extended regular expressions matching. Shell globs are used
+otherwise. Requires POSIX.1-2001 compliant system.
+.TP
 .B \-v
 Request that the \fIchat\fR script be executed in a verbose mode. The
 \fIchat\fR program will then log the execution state of the chat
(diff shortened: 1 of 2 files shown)


paulusmack:
I'm not going to pull these changes as they are.

* Thank you for putting a signed-off-by in each commit.

* Your commit messages are inadequate. You need to explain in more detail the broader consequences and motivation for the changes you are making.

* I don't like two-column indentation for C code, I prefer 8. Four-column is OK. I would accept a commit to convert the K&R-style argument declarations to ANSI C style.

* The main reason for not taking these commits is that they make an incompatible change that could break existing chat scripts -- if someone has a chat script that expects a string starting with '/', it won't work as it did before after this change.

* Another problem is that with these commits there is now a difference in behaviour between putting the expect/send strings on the command line vs. putting them in a file, where there wasn't before, which is undesirable.

* Have you considered using expect rather than chat? Expect is much more powerful and already has regexp matching built in.

elrafoon:
Thank you for review.

For now I'll keep it as it is, it works good for me. But if there is some interest from chat users, I can update the patches according to your points.

Neustradamus:
@elrafoon: Have you done improvements of your PR?

PR #90, opened 2017-11-29
inserted some code to export AC name to enviroment, so that AC name c…

bearmi:
I inserted some code to export AC name to enviroment, so that AC name can be collected and shown to end user

We want to show the access concentrator to the end user. So I added a variable PeerName in struct PPPoEConnectionStruct, and log the AC name to the variable when parsing PADO tags. 

And the value would be exported to env in function PPPOEConnectDevice. So that we can collect it and show it to user.

### DIFF
--- a/pppd/plugins/rp-pppoe/discovery.c
+++ b/pppd/plugins/rp-pppoe/discovery.c
@@ -149,6 +149,9 @@ parsePADOTags(UINT16_t type, UINT16_t len, unsigned char *data,
 	    !strncmp((char *) data, conn->acName, len)) {
 	    pc->acNameOK = 1;
 	}
+	conn->PeerName.type = htons(type);
+	conn->PeerName.length = htons(len);
+	memcpy(conn->PeerName.payload, data, len);
 	break;
     case TAG_SERVICE_NAME:
 	pc->seenServiceName = 1;
--- a/pppd/plugins/rp-pppoe/plugin.c
+++ b/pppd/plugins/rp-pppoe/plugin.c
@@ -219,6 +219,7 @@ PPPOEConnectDevice(void)
 	 conn->ifName);
 
     script_setenv("MACREMOTE", remote_number, 0);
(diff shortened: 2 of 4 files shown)


fhost:
I'm not sure using a `PPPoETag` structure to store the value is useful as it is not used as so later. Just using a string looks like a better way IMO.

I find the commit message a bit strange too, doesn't `export AC name to script environment` look enough?

paulusmack:
This seems like a useful feature, but this commit will need some more work before I can take it. I am not planning to put this in the 2.4.9 release, but it can go in after the release. This change intersects with the changes that @pali has submitted (#209) and should be reworked on top of those changes, I think.

I agree that using a string would be clearer and simpler than a PPPoETag structure.

The commit message needs a sign-off and should start with a headline like "plugins/pppoe: Export access concentrator name to environment", then a blank line, then a paragraph explaining why this is a useful thing to do.

Neustradamus:
@bearmi: Can you look the @paulusmack comment?

Neustradamus:
@bearmi: Can you look the @paulusmack comment?

Neustradamus:
@bearmi: Can you look the @fhost and @paulusmack comments?

Neustradamus:
@bearmi: Can you look the @fhost and @paulusmack comments?

The @pali PR has been merged.

Thanks in advance.

enaess:
I believe @pali has done work in this area. Could he please review this change?

enaess:
Change does look good to me though ...

pali:
This commit does not apply on top of the master branch... it needs to be rebased/reworked first.

Neustradamus:
@bearmi: Can you rebase your PR?

Neustradamus:
@enaess: I think it is needed to adapt the changes from this PR into a new... because @bearmi has not given a sign of life since a very very long time...

Neustradamus:
@bearmi: Can you look here please?
Thanks in advance.

paulusmack:
Superseded by #492.

PR #95, opened 2018-04-04
Honor LDFLAGS

yarda:
This allow distro to override LDFLAGS, e.g. https://bugzilla.redhat.com/show_bug.cgi?id=1563157

### DIFF
--- a/chat/Makefile.linux
+++ b/chat/Makefile.linux
@@ -18,7 +18,7 @@ INSTALL= install
 all:	chat
 
 chat:	chat.o
-	$(CC) -o chat chat.o
+	$(CC) $(LDFLAGS) -o chat chat.o
 
 chat.o:	chat.c
 	$(CC) -c $(CFLAGS) -o chat.o chat.c
--- a/pppd/plugins/Makefile.linux
+++ b/pppd/plugins/Makefile.linux
@@ -1,7 +1,7 @@
 #CC	= gcc
 COPTS	= -O2 -g
 CFLAGS	= $(COPTS) -I.. -I../../include -fPIC
-LDFLAGS	= -shared
+LDFLAGS_SHARED	= -shared
 INSTALL	= install
 
 DESTDIR = $(INSTROOT)@DESTDIR@
@@ -30,7 +30,7 @@ all:	$(PLUGINS)
 	for d in $(SUBDIRS); do $(MAKE) $(MFLAGS) -C $$d all || exit $$?; done
 
 %.so: %.c
-	$(CC) -o $@ $(LDFLAGS) $(CFLAGS) $^
+	$(CC) -o $@ $(LDFLAGS) $(LDFLAGS_SHARED) $(CFLAGS) $^
 
 VERSION = $(shell awk -F '"' '/VERSION/ { print $$2; }' ../patchlevel.h)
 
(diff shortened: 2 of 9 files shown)


chat/Makefile.linux
```diff
@@ -18,7 +18,7 @@ INSTALL= install
 all:	chat
 
 chat:	chat.o
```
teknoraver: can you try by removing the line below and leave only `chat:	chat.o`?

pppd/plugins/Makefile.linux
```diff
@@ -30,7 +30,7 @@ all:	$(PLUGINS)
 	for d in $(SUBDIRS); do $(MAKE) $(MFLAGS) -C $$d all || exit $$?; done
 
 %.so: %.c
-	$(CC) -o $@ $(LDFLAGS) $(CFLAGS) $^
+	$(CC) -o $@ $(LDFLAGS) $(LDFLAGS_SHARED) $(CFLAGS) $^
```
teknoraver: LDLIBS should be needed too.

chat/Makefile.linux
yarda: Did you mean to remove "all: chat"? I think somebody could call it with 'make all' and it seems to be used in other makefiles across this project. Also I think such change is not related to the purpose of this PR.

pppd/plugins/Makefile.linux
yarda: Did you mean LIBS? But it's for plugins (i.e. private non public DSOs), isn't enough to have LIBS in the final linkage of the executable?

yarda:
I updated the PR and added LDFLAGS to more places to fix problem which has been spotted by Fedora QA.

chat/Makefile.linux
teknoraver: No, I mean the whole `$(CC) $(LDFLAGS) -o chat chat.o` and the `chat:` target.
Make has some implicit rules which should be exactly the same

I didn't try but it should work without the target at all

yarda: Yes it works and also the rule for the chat.o can be removed. But this is not "let's write the shortest possible Makefile" contest. There are more Makefiles in this project which could be simplified with help of implicit rules. I think it should be changed consistently in all Makefiles or nowhere (personally I prefer explicit rules because it's more clear, but it's OT with regard to the original issue).

yarda:
Updated PR to add -Wl,-E for plugins.

pppd/plugins/Makefile.linux
teknoraver: LIBS is non standard, the correct one is LDLIBS according to the doc:

https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html

Neustradamus:
@yarda: It has been merged no?
- https://github.com/paulusmack/ppp/commit/a96702158e72c0d527a146940b983a366dbce464

yarda:
Yes, it's merged, thanks.

Neustradamus:
@yarda: Can you close this PR? :)

yarda:
Sorry, I missed it :)

Neustradamus:
No problem, some PRs have not been merged correctly in this repository, I think because there was before 2 repos in mirroring.

PR #98, opened 2018-04-30
Allow pppd to start as non-root.

micah-morton:
This patch adds #ifndef macros in 2 spots in order to allow pppd to be
spawned as a non-root user with only runtime capabilities (e.g.
CAP_NET_{RAW/ADMIN}) instead of giving pppd full root privileges. This
is helpful if pppd is itself spawned by a non-root user and the use of
file permissions (e.g. setuid-root) on the pppd binary is not a
desirable solution.

### DIFF
--- a/pppd/main.c
+++ b/pppd/main.c
@@ -371,6 +371,7 @@ main(argc, argv)
     if (debug)
 	setlogmask(LOG_UPTO(LOG_DEBUG));
 
+#ifndef ALLOW_START_AS_NON_ROOT
     /*
      * Check that we are running as root.
      */
@@ -379,6 +380,7 @@ main(argc, argv)
 		     argv[0]);
 	exit(EXIT_NOT_ROOT);
     }
+#endif
 
     if (!ppp_available()) {
 	option_error("%s", no_ppp_msg);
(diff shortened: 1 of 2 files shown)


paulusmack:
The basic idea seems fine, and the commit message is OK, though lacking a signed-off-by line.

With the ifndef in pppd/main.c, I'd prefer to see an explicit check that we have the capabilities we need if we're not running with euid=0, rather than simply not checking.

The ifndef in pppd/options.c doesn't look to be necessary. That code is about preventing a non-root user from overriding options set by the system administrator, it has basically nothing to do with the euid.

micah-morton:
Any idea regarding the set of runtime capabilities that pppd uses (and hence which capabilities we should check for in pppd/main.c)? The only capability I know for sure pppd needs is CAP_NET_ADMIN, but we are currently giving it CAP_NET_RAW and CAP_NET_BIND_SERVICE as well as part of a tree of processes. Then again I'm not familiar with the different functionalities of pppd to know if there are any other capabilities it may need in other use cases that I haven't mentioned here.

Neustradamus:
@micah-morton: Have you looked the PR from @a-andreyev?
- https://github.com/paulusmack/ppp/pull/136

What do you think?

@paulusmack: Can you look the update from today too?

PR #100, opened 2018-06-02
radius.c: Include CLASS attribute in radius_acct_interim and radius_a…

anphsw:
…cct_stop
Some radius servers will not accept accounting packets without CLASS attribute, this patch will include it.

Signed-off-by: Hauras Wigelstukas Alex <sysop@880.ru>


paulusmack:
Thanks. I applied this with a more informative commit message.

Neustradamus:
Merged here: https://github.com/paulusmack/ppp/commit/3215f9d68ee004b1797a54aad3d9989cb93f3344

PR #101, opened 2018-06-03
Support for interface renaming by pre-up scripts

dmbaturin:
When running, for example, both an L2TP/IPsec and a PPTP server on the same machine, it's nice to have them use different names for client interfaces , for example l2tpX and pptpX. It could be done with pre-ip-up scripts, but pppd assumes that interface name is not changed by them.

This is a patch made by Stephen Hemminger back in Vyatta times, but to my knowledge he never sent it to the maintainers. We are still using it in VyOS and it works well for us, but it would be nice if other people could benefit from it too (and we didn't have to maintain a custom version of pppd).

It looks POSIX-compliant to me, but I do admit I've never tested it on systems other than Linux. If there are any compatibility problems with it or other issues, I'm ready to work on them.

### DIFF
--- a/pppd/ipcp.c
+++ b/pppd/ipcp.c
@@ -55,6 +55,7 @@
 #include <sys/socket.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
+#include <net/if.h>
 
 #include "pppd.h"
 #include "fsm.h"
@@ -1788,6 +1789,7 @@ ipcp_up(f)
     ipcp_options *ho = &ipcp_hisoptions[f->unit];
     ipcp_options *go = &ipcp_gotoptions[f->unit];
     ipcp_options *wo = &ipcp_wantoptions[f->unit];
+    int ifindex;
 
     IPCPDEBUG(("ipcp: up"));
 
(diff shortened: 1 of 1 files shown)


paulusmack:
The change looks reasonable.

The commit message should be more descriptive. Include the reasons why you want to do this and why the ifname option is not adequate for your purposes.

I think there should be an error message (printed with error()) if we can't find the interface rather than just a debug message.

Please make it one commit rather than two.

Neustradamus:
@dmbaturin: Can you reply to @paulusmack and update your PR?

dmbaturin:
@Neustradamus @paulusmack I've updated the PR.

I feel like I need to provide some explanation of why I did it the way I did it initially and why I was under impression that I'm simply being redtaped.

>Please make it one commit rather than two.

There was a specific reason to keep that history at least inside the PR even if not in the mainline version: I've been adapting a change made by @shemminger to work with recent ppp. To make it one commit I had a choice between taking the credit for the change, or attributing my own (incredibly minor and technical) changes to Stephen.
In this update I solved the issue with a comment inside the commit message—I hope this is a satisfactory solution.

Also, one-click "squash and merge" has been available on GitHub for a long time, so I sort of internalized the "keeping or destroying history is repo owner's decision" principle.

>I think there should be an error message (printed with error()) if we can't find the interface rather than just a debug message.

I've done that now, but in fact the original change simply mirrored the "common practice" of that file: that pattern is everywhere there, e.g. https://github.com/paulusmack/ppp/blob/master/pppd/ipcp.c#L1916-L1919
I believe all those error can be improved indeed, but it seems out of the scope of this patch.

>The commit message should be more descriptive. Include the reasons why you want to do this and why the ifname option is not adequate for your purposes.

I've updated it too now. You cannot really use `ifname` for _client_ connections, and server operators may want more control than `ifname` alone could provide.

PR #102, opened 2018-06-03
Periodically close and re-open log file so that it can be rotated.

dmbaturin:
This was also made in Vyatta times, by Bob Gilligan in this case. Not sure how many people use log files instead of syslog or journald now, but see if you want this.

### DIFF
--- a/pppd/options.c
+++ b/pppd/options.c
@@ -1559,7 +1559,7 @@ setlogfile(argv)
 	option_error("unable to drop permissions to open %s: %m", *argv);
 	return 0;
     }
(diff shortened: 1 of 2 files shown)


paulusmack:
The commit message needs to be more informative.

There is an apparently unrelated change to the permissions for the log file (0644 changed to 0664) which is not explained in the commit message.

Wouldn't it be better to do this on a time basis rather than just every N bytes?

fhost:
Using the same approch of `tail -F` (following filename by watching parent directory) looks like a more resilent way to do so.

Neustradamus:
@dmbaturin: Can you reply to @paulusmack?

PR #106, opened 2018-07-24
pppd: use monotonic clock source for callouts

pkeroulas:
gettimeofday() isn't the best suitable function to compute duration.
It is prone to fail when the system time changes. Using the monotonic
clock as a time source makes it more resilient.

Signed-off-by: Patrick Keroulas <patrick.keroulas@savoirfairelinux.com>

### DIFF
--- a/pppd/main.c
+++ b/pppd/main.c
@@ -81,6 +81,7 @@
 #include <utmp.h>
 #include <pwd.h>
 #include <setjmp.h>
+#include <time.h>
 #include <sys/param.h>
 #include <sys/types.h>
 #include <sys/wait.h>
@@ -1257,7 +1258,7 @@ struct	callout {
 };
 
 static struct callout *callout = NULL;	/* Callout list */
-static struct timeval timenow;		/* Current time */
+static struct timespec timenow;		/* Current time from monotonic clock */
 
 /*
  * timeout - Schedule a timeout.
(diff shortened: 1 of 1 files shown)


fhost:
You wrote `cmonotonic clock` in the error message, is this a typo or we do say `cmonotonic`?

I think more places should need the update from `gettimeofday()` to `clock_gettime(MONOTONIC_CLOCK)`:
- pppd/tty.c:1075
- pppd/tty.c:1124
- pppd/main.c:525
- pppd/main.c:1233
- pppd/plugins/rp-pppoe/discovery.c:48
- pppd/plugins/rp-pppoe/discovery.c:364
- pppd/plugins/rp-pppoe/discovery.c:548

But anyway, this is a great update!

pkeroulas:
Correct, this is a typo.

This commit only addresses callout timings. The other `gettimeofday()` should be replaced in later changes so that eventual regression will be easier to identified. I'll see what I can do.

fhost:
You correct your typo only once, but it is present in all error messages you've changed.

thjc:
This week I ran into issues with clock stepping blocking connection changes. I have tested this patch and it is working for me. Is it able to be merged as is, or is it desired to have the other instances replaced before merging it?

themiron:
jfyi, https://github.com/paulusmack/ppp/pull/123
should be more general solution

paulusmack:
I pulled #123.

PR #116, opened 2019-03-13
PPP iface enhanced to ifnameN

EasyNetDev:
Hi,

Because I'm using this kind of patch for a while, I was thinking to see if you are agree with such change.
Is a small change in pppd and radius-plugin which allows you under Linux to use arbitrary names for the PPP daemon instead using standard naming pppX.
The previous patch in PPP is allowing you to use a fix name for the interface name, not allowing PPP to use unit number in the end.

The idea comes when I've tried to use a script to change the name, but because PPP is allocating by default pppX naming , radius plugin also had issues sending correct interims packets to radius server and the database couldn't be updated correctly and the clients were disconnected because of false inactivity in database.

In this case i've changed the PPP and I've improved the previous patch of fixed naming "ifname" to "ifnameN" with backwards compatibility with "ifname" format.

For example if I have 3 daemons running on same server, one is used for PPTP, other is PPPoE and other is L2TP, then for each daemon I'm creating a ppp configuration file and I'm setting like this:

PPTP:
```
ifname pptpN
```

PPPoE:
```
ifname pppoesN
```

L2TP:
```
ifname lt2pN
```

The daemon will rename the interfaces like:
pptp0, pptp1, pppoes2, l2tp3, pppoes3, pptp4 ..

Which will be a benefit also for the firewall if you want to classify the traffic in specific targets.

Kind regards,
Adrian

### DIFF
--- a/pppd/main.c
+++ b/pppd/main.c
@@ -730,7 +730,28 @@ set_ifunit(iskey)
     int iskey;
 {
     if (req_ifname[0] != '\0')
-	slprintf(ifname, sizeof(ifname), "%s", req_ifname);
+    {
+	char *ptr;
+	ptr = strchr(req_ifname, 'N');
+
+	if (ptr == NULL)
+	    slprintf(ifname, sizeof(ifname), "%s", req_ifname);
+
+	else {
+	    int index=0;
+	    char req_ifname_t[MAXIFNAMELEN];
+	    index = (int)(ptr - req_ifname);
+	    memset(req_ifname_t, 0, MAXIFNAMELEN); // force cleanup
+	    strncpy(req_ifname_t, req_ifname, index);
+
+	    // Copy the part of the req_ifname which doesn't contains "N"
+	    strncpy(req_ifname_t, req_ifname, index);
+
+	    // Add the ifunit to the end of the interface name
+	    slprintf(ifname, sizeof(ifname), "%s%d", req_ifname_t, ifunit);
+	    info("Using ifname in format ifnameN. Set to new name: '%s'", ifname);
+	}
+    }
     else
 	slprintf(ifname, sizeof(ifname), "%s%d", PPP_DRV_NAME, ifunit);
     info("Using interface %s", ifname);
--- a/pppd/options.c
+++ b/pppd/options.c
@@ -294,7 +294,7 @@ option_t general_options[] = {
       OPT_PRIO | OPT_LLIMIT, 0, 0 },
 
     { "ifname", o_string, req_ifname,
-      "Set PPP interface name",
+      "Set PPP interface name. If format is like ifnameN, N it will be replaced by unit number",
       OPT_PRIO | OPT_PRIV | OPT_STATIC, NULL, MAXIFNAMELEN },
 
     { "dump", o_bool, &dump_options,
--- a/pppd/plugins/radius/radius.c
+++ b/pppd/plugins/radius/radius.c
@@ -39,6 +39,7 @@ static char const RCSID[] =
 #include "ipcp.h"
 #include <syslog.h>
 #include <sys/types.h>
+#include <regex.h>
 #include <sys/time.h>
 #include <string.h>
 #include <netinet/in.h>
@@ -1314,7 +1315,25 @@ static int
 get_client_port(char *ifname)
 {
     int port;
-    if (sscanf(ifname, "ppp%d", &port) == 1) {
+    char cport[10];
+    int reti;
+    regex_t regex;
+    regmatch_t matches[2]; /* Array of matches */
+
+    /* Compile regular expression */
+    reti = regcomp(&regex, "^[[:alpha:]]\\+\\([0-9]\\+\\)$", 0);
+    /* Execute regular expression */
+    reti = regexec(&regex, ifname, 2, matches, 0);
+    regfree(&regex);
+
+    //info("radius.c: get_client_port: %s", ifname);
+
+    if (!reti) {
+        memset(cport, 0, 10); // force cleanup
+        strncpy(cport, ifname + matches[1].rm_so, (int)(matches[1].rm_eo-matches[1].rm_so));
+        port = atoi(cport);
+        //info("radius.c: get_client_port: %s", cport);
+        //info("radius.c: get_client_port: %d", port);
 	return port;
     }
     return rc_map2id(ifname);
--- a/pppd/sys-linux.c
+++ b/pppd/sys-linux.c
@@ -647,15 +647,49 @@ static int make_ppp_unit()
 	if (x == 0 && req_ifname[0] != '\0') {
 		struct ifreq ifr;
 		char t[MAXIFNAMELEN];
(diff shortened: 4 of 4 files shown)


Neustradamus:
@AdrianBan: Can you update your PR?

@paulusmack: What do you think about this one?

paulusmack:
The idea seems OK in principle.

I have some issues with the implementation, and the commit is not acceptable as-is because it doesn't have any description or Signed-off-by line.

The indentation of the added code is wonky.

For some reason you have two identical strncpy calls in a row. I don't like adding new calls to strncpy. You can in fact do the trimming of the interface name to the part up to the N using a %.*s format to slprintf, without needing to make a copy at all.

Just looking for an 'N' anywhere in the interface name seems dangerous. I would use some less usual character, for example '%', and require that it be the last character in the interface name.

It's not clear why radius now needs to use regular expressions. If that is really essential then it must be explained in the commit message. I would rather not use regcomp/regexec unless it's really necessary, because pppd gets used on embedded devices with small libcs that may not include regcomp/regexec.

Neustradamus:
@AdrianBan: Can you look the answer of @paulusmack?

AdrianBan:
Hi all,

It past sometime from when I've send this patch. I can review it again for the latest version of PPP.
Also I can change a little bit the code, adding some comments, removing double strncpy and search for another type of character. I was using this patch for few networks where I've implemented the PPPoE and they worked without issue for long time.

Kind regards,
Adrian

Neustradamus:
@AdrianBan: Thanks for your reply :)

Have you looked the code after the comment of @paulusmack?

At the end of 2020, @paulusmack has done various changes and some mergers.

It will be perfect if your PR can be merged before Debian 11 freeze.
The next Debian will be in 2023.

Neustradamus:
@AdrianBan: Tomorrow @paulusmack will create the 2.4.9...

AdrianBan:
Hi,

Sorry, but I had other project to finish on ARM devices.
I will try the best to change the PPP feature to a better code.

Soon as I have this new code, I will send to you.

Best regards,
Adrian

Neustradamus:
@EasyNetDev: Any news?

pali:
@EasyNetDev: PPPoE connections are bound to the ethernet interface (e.g eth3). I would like to see a way to specify that PPP interface name would contain also ethernet interface, not just PPP number, e.g. `eth3-pppoe10`.

What do you think about rather introducing %fmt syntax in `ifname`? E.g. `ifname %d-ppoe%n` With %fmt syntax it would be possible to extend it in future with backward compatibility. But now if you make any unescaped `N` character as a special then `N` would not be possible to use anymore. And also in future it would not be possible to extend it with backward compatibility.

Neustradamus:
@EasyNetDev: Have you looked?
Please read the @pali comment too.

Neustradamus:
@EasyNetDev: Do you received my emails?

Neustradamus:
@EasyNetDev: Have you looked?

Neustradamus:
@EasyNetDev: More 2 years since Mar 13, 2019 ^^

I would like a "success" for this PR ;)

EasyNetDev:
Hi,

I'm starting to adapt the code to latest PPP and also I will use regex and using antoher string for matching the ppp id.

EasyNetDev:
> 
> 
> @EasyNetDev: PPPoE connections are bound to the ethernet interface (e.g eth3). I would like to see a way to specify that PPP interface name would contain also ethernet interface, not just PPP number, e.g. `eth3-pppoe10`.
> 
> What do you think about rather introducing %fmt syntax in `ifname`? E.g. `ifname %d-ppoe%n` With %fmt syntax it would be possible to extend it in future with backward compatibility. But now if you make any unescaped `N` character as a special then `N` would not be possible to use anymore. And also in future it would not be possible to extend it with backward compatibility.

One thing we need to keep in mind: linux maximum interface name is 16 byte and only 15 are usable. In the case of old interface naming is ok, but new predictible interface naming is a big problem, example: 
enp1s0f0-pppoe10 -> 16 bytes at least! We can't use this type of method. 
Plus we need to introduce more checks in case of PPP is not coming on ethernet interface, example:

```
Jun 16 22:23:24 R01 pppd[2363060]: Renamed interface ppp0 to l2tp0
Jun 16 22:23:24 R01 pppd[2363060]: Found dynamic custom interface naming: l2tp0
Jun 16 22:23:24 R01 pppd[2363060]: Using interface l2tp0
Jun 16 22:23:24 R01 pppd[2363060]: Connect: l2tp0 <-->
Jun 16 22:23:24 R01 pppd[2363060]: Overriding mtu 1500 to 1300
Jun 16 22:23:24 R01 pppd[2363060]: PPPoL2TP options: lnsmode tid 14557 sid 30941 debugmask 0
Jun 16 22:23:24 R01 pppd[2363060]: Overriding mru 1500 to mtu value 1300
```
Connect part is not having an interface for L2TP, for example.

EasyNetDev:
My new code is using a format of ifname like:
```
ifname pppoe%s
```
Which %s is the dynamic ppp unit id.

pali:
@EasyNetDev: I see. But nobody is forced to use systemd predictible interface naming. You can even write your own udev rules and create your own naming. And I would like to see same for pppd naming. Just allows admins to configure template via %fmt syntax to use their own naming. If somebody is doing to use custom template naming then probably has also idea how to "overwrite" systemd predictible naming and use their own "naming convension".

And what would be default naming, just choose saming sane. "l2tp%d" is fine; also "pppoe%d" is fine.

> My opinion you can use for each ethernet interface a new PPP file which have different "ifname" parameter:

This is something which can be done already, but is not ideal. As for each PPPoE connection it is required to create a new config file. Not elegant.

Also something which may be interesting for PPPoE, allow to specify into template PPPoE session id (this is not ppp unit id!).

EasyNetDev:
@pali first let's see this patch in the main code
I was creating this pull request 3 years ago and this year I got an answer regarding this.
I thought ppp was not maintained anymore, honestly. 
3 years.. 
After this patch which I'm re-writing it we can see what we can implement more.

EasyNetDev:
@pali another idea is to change the iflink parameter of the new ppp interface and correlate with the interface where the connexion is comming.
It helps this?

pali:
Well, the best would be to finish your patch in the current form. And if you are going to use %fmt syntax then as a next step it is easy to extend code for a new / additional %arg. And it can be done by anybody in e.g. followup pull request.

EasyNetDev:
I got the final patch working on my system.
I think I will need to add some additional checks for BSD and Solaris. I don't which OSes are supporting interface renaming. I know some OSes, like FreeBSD is not supporting interface renaming. 
I would need some help here. We can add this additional check by simple eliminate the ifname directive for these OSes.
The code is backwards compatible with ppp naming in case you don't have the ifname directive.

pali:
I think you can start with just linux-only implementation and later this functionality can be extended also to other platforms... There are already some options under `#ifdef __linux__` code.

EasyNetDev:
New patch is added via #285. I can close this PR.

PR #122, opened 2019-06-06
pppd: Check tbd pointer before closing

alfonsosanchezbeato:
Check that pointer to the tdb is not NULL before calling tdb_close().
It is possible that the file could not be opened/created due to
permission issues. This change prevents the crash that happens in that
case.

### DIFF
--- a/pppd/main.c
+++ b/pppd/main.c
@@ -1566,7 +1566,8 @@ safe_fork(int infd, int outfd, int errfd)
 	/* Executing in the child */
 	sys_close();
 #ifdef USE_TDB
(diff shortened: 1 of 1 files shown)


alfonsosanchezbeato:
Added "Signed-off-by" line in commit description.

paulusmack:
Thanks. I cherry-picked the patch so I could fix the typo in the headline.

Neustradamus:
Merged here: https://github.com/paulusmack/ppp/commit/b6b4d28e0c38320ca6753af40845df991118cd11

PR #123, opened 2019-06-26
pppd: linux: use monotonic time if possible

themiron:
gettimeofday() suffers from time jumps due ntp or any manual change,
so duration measurements and scheduling can not be accurate.
let's instead use monotonic time source, if available.

it's known glibc (< 2.3.4) & old uclibc don't provide CLOCK_MONOTONIC
denine, but kernel may have it supported. so, use clock_gettime()
with fallback to gettimeofday() if first call has failed.

several gettimeofday()/time() calls still have to be preserved for
debug, pseudoterminal timestamping and string formatting. all the
rest calls are replaced to new get_time() call.

solaris kept with gettimeofday() as before, corresponding get_time()
system implementation can be updated/added in any future.

Signed-off-by: Vladislav Grishenko <themiron@mail.ru>

### DIFF
--- a/pppd/sys-linux.c
+++ b/pppd/sys-linux.c
@@ -2966,3 +2966,38 @@ ether_to_eui64(eui64_t *p_eui64)
     return 1;
 }
 #endif
+
+/********************************************************************
+ *
+ * get_time - Get current time, monotonic if possible.
+ */
+int
+get_time(struct timeval *tv)
+{
+/* Old glibc (< 2.3.4) does define CLOCK_MONOTONIC, but kernel may have it.
+ * Runtime checking makes it safe. */
+#ifndef CLOCK_MONOTONIC
+#define CLOCK_MONOTONIC 1
+#endif
+    static int monotonic = -1;
+    struct timespec ts;
+    int ret;
+
+    if (monotonic) {
+	ret = clock_gettime(CLOCK_MONOTONIC, &ts);
+	if (ret == 0) {
+	    monotonic = 1;
+	    if (tv) {
+		tv->tv_sec = ts.tv_sec;
+		tv->tv_usec = ts.tv_nsec / 1000;
+	    }
+	    return ret;
+	} else if (monotonic > 0)
+	    return ret;
+
+	monotonic = 0;
+	warn("Couldn't use monotonic clock source: %m");
+    }
+
+    return gettimeofday(tv, NULL);
+}
--- a/pppd/Makefile.linux
+++ b/pppd/Makefile.linux
@@ -33,7 +33,7 @@ endif
 # CC = gcc
 #
 COPTS = -O2 -pipe -Wall -g
-LIBS =
+LIBS = -lrt
 
 # Uncomment the next line to include support for Microsoft's
 # MS-CHAP authentication protocol.  Also, edit plugins/radius/Makefile.linux.
--- a/pppd/main.c
+++ b/pppd/main.c
@@ -522,7 +522,7 @@ main(argc, argv)
 	    info("Starting link");
 	}
 
-	gettimeofday(&start_time, NULL);
+	get_time(&start_time);
 	script_unsetenv("CONNECT_TIME");
 	script_unsetenv("BYTES_SENT");
 	script_unsetenv("BYTES_RCVD");
@@ -1216,7 +1216,7 @@ reset_link_stats(u)
 {
     if (!get_ppp_stats(u, &old_link_stats))
 	return;
-    gettimeofday(&start_time, NULL);
+    get_time(&start_time);
 }
 
 /*
@@ -1230,7 +1230,7 @@ update_link_stats(u)
     char numbuf[32];
 
     if (!get_ppp_stats(u, &link_stats)
-	|| gettimeofday(&now, NULL) < 0)
+	|| get_time(&now) < 0)
 	return;
     link_connect_time = now.tv_sec - start_time.tv_sec;
     link_stats_valid = 1;
@@ -1277,7 +1277,7 @@ timeout(func, arg, secs, usecs)
 	fatal("Out of memory in timeout()!");
     newp->c_arg = arg;
     newp->c_func = func;
-    gettimeofday(&timenow, NULL);
+    get_time(&timenow);
     newp->c_time.tv_sec = timenow.tv_sec + secs;
     newp->c_time.tv_usec = timenow.tv_usec + usecs;
     if (newp->c_time.tv_usec >= 1000000) {
@@ -1331,7 +1331,7 @@ calltimeout()
     while (callout != NULL) {
 	p = callout;
 
-	if (gettimeofday(&timenow, NULL) < 0)
+	if (get_time(&timenow) < 0)
 	    fatal("Failed to get time of day: %m");
 	if (!(p->c_time.tv_sec < timenow.tv_sec
 	      || (p->c_time.tv_sec == timenow.tv_sec
@@ -1356,7 +1356,7 @@ timeleft(tvp)
     if (callout == NULL)
 	return NULL;
 
-    gettimeofday(&timenow, NULL);
+    get_time(&timenow);
     tvp->tv_sec = callout->c_time.tv_sec - timenow.tv_sec;
     tvp->tv_usec = callout->c_time.tv_usec - timenow.tv_usec;
     if (tvp->tv_usec < 0) {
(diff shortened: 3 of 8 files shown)


pppd/sys-linux.c
```diff
@@ -2966,3 +2967,41 @@ ether_to_eui64(eui64_t *p_eui64)
...
+#ifndef CLOCK_MONOTONIC
+#define CLOCK_MONOTONIC 1
+#endif
+    static int monotonic = -1;
+    struct timespec ts;
+    int ret;
+
+    if (monotonic) {
+	/* Call clock_gettime() directly to skip librt linking */
+	ret = syscall(SYS_clock_gettime, CLOCK_MONOTONIC, &ts);
```
pali, line 2990: It is a bad idea to call syscall directly. First reason is that these time-based syscalls are supported by vdso, glibc wrappers use vdso when available and therefore it fully avoids to call (possible expensive) syscall.

Second reason is that since glibc 2.17 clock_gettime() function is available directly in libc.so, and not in librt.so. So also comment about skip linking is misleading/incorrect.

So this proposed patch is harmfull. It replaces fast vdso calls by slower syscalls. And plus it uses Linux-platform specific macros for direct issuing syscalls (which is insecure if you do not pass correct arguments) which makes code less portable.

Morover `clock_gettime` with `CLOCK_MONOTONIC` is in POSIX.1-2001, so there is no reason to use Linux-specific code when POSIX code is available (also on Linux).

themiron: > First reason is that these time-based syscalls are supported by vdso, glibc wrappers use vdso when available and therefore it fully avoids to call (possible expensive) syscall.

Reason is not generally valid, pppd is used in wild/embedded platforms where vendor SDKs have no glibc and vdso, as stated in initial description.

> Second reason is that since glibc 2.17 clock_gettime() function is available directly in libc.so, and not in librt.so. So also comment about skip linking is misleading/incorrect.

This is not pretty valid in terms of compatibility/portability either, pppd is used in wild with glibc < 2.17, as stated in initial description. Therefore librt have to be included for clock_gettime().

> So this proposed patch is harmfull. It replaces fast vdso calls by slower syscalls.

Sure there can be slowdown due context switches only for vdso case (not the only as I pointed earlier), but it does not really make sense with the way and frequency pppd gets the time.

> And plus it uses Linux-platform specific macros for direct issuing syscalls (which is insecure if you do not pass correct arguments) which makes code less portable.

Linux-specific macros is not the case, because it is sys-linux.c backend exactly for linux platform obviously.

> Morover clock_gettime with CLOCK_MONOTONIC is in POSIX.1-2001, so there is no reason to use Linux-specific code when POSIX code is available (also on Linux).

Old 2.3.4 globc doesn not provide it, old uclibc does not provide it, etc.
Defining only if when not defined makes no harm.

Although, you're conceptually right - there can be syscall number issue on 64-bit linux platforms and direct syscall use will be harmful after switching 32-bit targets to 64-bit time structures.
For example,  please refer https://bugs.busybox.net/show_bug.cgi?id=12091.
So, clock_gettime() seems the only reliable way for long-run with the costs of librt linking, thank you.

pali: Ok, if there are still important linux platforms without glibc or working clock_gettime() function then I think it is better to stick with gettimeofday() function. That timeswitching and using raw syscalls just cause more problems. I guess that non-monotonic gettimeofday() with possible jumps cause less problems on these platforms.

I'm thinking about following:
* glibc >= 2.17 - always use clock_gettime() as it is in libc.so
* glibc >= 2.34 && glibc < 2.17 - use clock_gettime() only if user decided to link with librt.so; otherwise use gettimeofday() so librt.so is not required
* glibc < 2.17 - always use gettimeofday()
* old uclibc or any libc which does not provide clock_gettime() - always use gettimeofday()

It is possible? Or are there any other problems?

Glibc detection can be done at compile time via macros, e.g.:
```c
#if defined(__GNU_LIBRARY__) && (__GLIBC__ > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 17))
/* Code for glibc >= 2.17 */
#endif
```

themiron: > * glibc >= 2.17 - always use clock_gettime() as it is in libc.so

librt is still available for glibc, right? so it will be linked-in  w/o effect, most likely shared librt will be already loaded into ram earlier/later by another app.
in case of gnu-ld we can omit linking with --as-needed, but we can't garantee there is gnu linker, detecting will increase complexity.
uclibc/uclibc-ng/musl either provide own librt or provide stub, no harm is expected I guess.

> * glibc >= 2.34 && glibc < 2.17 - use clock_gettime() only if user decided to link with librt.so; otherwise use gettimeofday() so librt.so is not required.

glibc/uclibc/whatever toolchain can have no CLOCK_MONOTONIC defined, but kernel can happily supports that, I don't think it's necessary to stick with exact glibc versions -> will require more exeptions in the future for non-glibc things.

> * glibc < 2.17 - always use gettimeofday()

same rationale, lots of platforms in wild has it < 2.17 and clock_gettime() will prefectly work even with customly defined CLOCK_MONOTONIC.

> * old uclibc or any libc which does not provide clock_gettime() - always use gettimeofday()

this is tricky... ppp has no relibale way to detect whether clock_gettime() is available and what shared library provides it unlike autotools/cmake.
what about to add ~USE_MONOTONIC option to Makefile.linux alike USE_LIBUTIL and enable it by default? will allow to disable librt/clock_gettime() off w/o patching if it causes compat/build issues on very ancient c libraries, please note it's linux-only case, solaris* is out of the scope.

pali: > librt is still available for glibc, right? so it will be linked-in w/o effect, most likely shared librt will be already loaded into ram earlier/later by another app.

librt.so is still available. But both libc.so and librt.so provides clock_gettime function. So when librt is linked then probably function librt.so would be used (instead of libc.so). But I do not think this cause any problems, this function is stateless. The only problem is "cosmetic", linking to librt.so is not needed.

> in case of gnu-ld we can omit linking with --as-needed

I'm not sure if this can really omit linking as function is still available in librt.so and linker could expect that user really want to call that function from librt.so (otherwise would not explicitly link to that library).

> glibc/uclibc/whatever toolchain can have no CLOCK_MONOTONIC defined, but kernel can happily supports that,

Maybe I misunderstood you. Does it mean that those old glibc versions *have* clock_gettime and *does not* have only CLOCK_MONOTONIC define? Or did you mean that those old glibc versions even do *not have* clock_gettime function and it is needed to use raw syscall?

> what about to add ~USE_MONOTONIC option to Makefile.linux

For me it looks a nice solution. Via compile time option can user choose what he wants. Either clock_gettime via additional librt because he uses desktop. Or plain gettimeofday as his embedded libc.so does not provide clock_gettime and do not have enough space for librt.so, etc...

themiron: > I'm not sure if this can really omit linking as function is still available in librt.so and linker could expect that user really want to call that function from librt.so (otherwise would not explicitly link to that library).

it can omit linking only if function is available in other but specified library.
for example, uclibc-ng has monolitic libc.so with clock_getttime(), so librt will be skipped as long as gnu linker finds function in libc. not sure about portability here...

> Does it mean that those old glibc versions have clock_gettime and does not have only CLOCK_MONOTONIC define? Or did you mean that those old glibc versions even do not have clock_gettime function and it is needed to use raw syscall?

yes, clock_gettime() is present with the only CLOCK_REALTIME (==0) defined, CLOCK_MONOTONIC (==1) has added a bit later.

> For me it looks a nice solution.

okie, will add soon, thank you

pali: > yes, clock_gettime() is present with the only CLOCK_REALTIME (==0) defined, CLOCK_MONOTONIC (==1) has added a bit later.

Ah, so then using clock_gettime() with 1 is safe and fine. Previously I understood it incorrectly and though that older glibc versions do not have clock_gettime at all (and raw syscall wrapper is needed).

> > For me it looks a nice solution.
> 
> okie, will add soon, thank you

Just to note, this is my opinion about it.

I'm not maintainer, so maybe @paulusmack as maintainer would have different opinition about it.

paulusmack:
I merged your commit, but I think there is a typo in the comment in get_time(). You say "Old glibc (< 2.3.4) does define CLOCK_MONOTONIC, but kernel may have it." I think you mean "does _not_ define".

themiron:
@paulusmack thank you, right "not" was lost in syscall->get_time change
previous comment:
https://github.com/paulusmack/ppp/commit/adbed1e0355ee2859f8c8f491a96d17a06a7f79c#diff-c9f5822a05913403f7af6dcb8f8245d6R2979
should I make new pr for typo or?

Neustradamus:
Merged here: https://github.com/paulusmack/ppp/commit/c78e3129d404f20d556f727ceee3704722de8cc7

@themiron: Please do a new PR if there is missing...

PR #129, opened 2019-11-20
pppd: free of memory not on the heap in multilink

flozilla:
On Linux, pppd may attempt to free memory that is statically allocated. This bug can be provoked by executing, for instance, "pppd login mp".

Freeing memory that has not been allocated on the heap can in the worst case lead to arbitrary code execution (https://cwe.mitre.org/data/definitions/590.html). However, because (i) the memory region that is attempted to be freed cannot be controlled by an adversary and (ii) the bug can only be triggered locally, I find this bug is not security sensitive.

To fix this bug, I propose that the free() operation is only executed on Solaris (not on Linux), where "p" is actually allocated on the heap.

### DIFF
--- a/pppd/multilink.c
+++ b/pppd/multilink.c
@@ -445,12 +445,16 @@ get_default_epdisc(ep)
 	if (p != 0 && get_if_hwaddr(ep->value, p) >= 0) {
 		ep->class = EPD_MAC;
 		ep->length = 6;
+		#ifndef __linux__
 		free(p);
+		#endif
 		return 1;
 	}
 
+	#ifndef __linux__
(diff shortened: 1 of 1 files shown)


paulusmack:
In fact the return value isn't dynamically allocated on Solaris either, so the ifdefs are not necessary.  Your commit lacks a signoff and a meaningful commit message. Since the fix is trivial, I just pushed the fix without the ifdefs and with reported-by you.

Neustradamus:
Linked to: https://github.com/paulusmack/ppp/commit/09f695f09e4ff6905388e15b106cd5669b05965e

PR #140, opened 2020-03-04
Adding EAP-TLS support

jjkeijser:
Added EAP-TLS support based on ppp-2.4.7-eaptls-mppe-1.200.patch with support for
- EAP-TLS authentication using X509 certificates on both client and server side
- TLS v1 - v1.3 (depending on OpenSSL library support)
- PKCS#11 hardware tokens
- pppd command-line options to specify certificate, private key, CA certificate file, CA certificate path
- with OpenSSL support now included, you can switch between using the OpenSSL provided MD5 & SHA1 functions or the code supplied by pppd itself using Makefile switches.

This patch also addresses CVE-2018-11574 and also build on Solaris (x86 tested).

Previous patch: https://www.nikhef.nl/~janjust/ppp/ppp-2.4.7-eaptls-mppe-1.102.patch

### DIFF
--- a/README.eap-tls
+++ b/README.eap-tls
@@ -1,0 +1,291 @@
+EAP-TLS authentication support for PPP
+======================================
+
+1. Intro
+
+    The Extensible Authentication Protocol (EAP; RFC 3748) is a
+    security protocol that can be used with PPP.  It provides a means
+    to plug in multiple optional authentication methods.
+
+    Transport Level Security (TLS; RFC 5216) provides for mutual 
+    authentication, integrity-protected ciphersuite negotiation and 
+    key exchange between two endpoints.  It also provides for optional
+    MPPE encryption.
+
+    EAP-TLS (RFC 2716) incapsulates the TLS messages in EAP packets,
+    allowing TLS mutual authentication to be used as a generic EAP
+    mechanism. It also provides optional encryption using the MPPE
+    protocol.
+
+    This patch provide EAP-TLS support to pppd.
+    This authentication method can be used in both client or server
+    mode.
+
+2. Building
+
+    To build pppd with EAP-TLS support, OpenSSL (http://www.openssl.org)
+    is required. Any version from 0.9.7 should work.
+    
+    Configure, compile, and install as usual. 
+
+3. Configuration
+
+    On the client side there are two ways to configure EAP-TLS:
+
+    1. supply the appropriate 'ca', 'cert' and 'key' command-line parameters
+
+    2. edit the /etc/ppp/eaptls-client file.
+    Insert a line for each system with which you use EAP-TLS.
+    The line is composed of this fields separated by tab:
+
+      - Client name 
+        The name used by the client for authentication, can be *
+      - Server name
+        The name of the server, can be *
+      - Client certificate file 
+        The file containing the certificate chain for the 
+        client in PEM format
+      - Server certificate file
+        If you want to specify the certificate that the 
+        server is allowed to use, put the certificate file name.
+        Else put a dash '-'.
+      - CA certificate file
+        The file containing the trusted CA certificates in PEM
+        format.
+      - Client private key file
+        The file containing the client private key in PEM format.
+
+
+    On the server side edit the /etc/ppp/eaptls-server file.
+    Insert a line for each system with which you use EAP-TLS.
+    The line is composed of this fields separated by tab:
+
+      - Client name
+        The name used by the client for authentication, can be *
+      - Server name
+        The name of the server, can be *
+      - Client certificate file
+        If you want to specify the certificate that the
+        client is allowed to use, put the certificate file name.
+        Else put a dash '-'.
+      - Server certificate file
+        The file containing the certificate chain for the
+        server in PEM format
+      - CA certificate file
+        The file containing the trusted CA certificates in PEM format.
+      - Client private key file
+        The file containing the server private key in PEM format.
+      - addresses
+        A list of IP addresses the client is allowed to use.
+
+
+    OpenSSL engine support is included starting with v0.95 of this patch. 
+    Currently the only engine tested is the 'pkcs11' engine (hardware token
+    support). To use the 'pksc11' engine:
+      - Use a special private key fileiname in the /etc/ppp/eaptls-client file:
+          <engine>:<identifier>
+        e.g.
+          pkcs11:123456
+
+      - The certificate can also be loaded from the 'pkcs11' engine using
+        a special client certificate filename in the /etc/ppp/eaptls-client file:
+          <engine>:<identifier>
+        e.g.
+          pkcs11:123456
+
+      - Create an /etc/ppp/openssl.cnf file to load the right OpenSSL engine prior
+        to starting 'pppd'. A sample openssl.cnf file is
+
+        openssl_conf = openssl_def
+
+        [ openssl_def ]
+        engines = engine_section
+
+        [ engine_section ]
+        pkcs11 = pkcs11_section
+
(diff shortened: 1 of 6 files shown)


xosevp:
> Added EAP-TLS support based on ppp-2.4.7-eaptls-mppe-1.200.patch with support for

There are newer patches for ppp-**2.4.8**: https://www.nikhef.nl/~janjust/ppp/download.html

paulusmack:
The commit here just adds a few files; it doesn't modify pppd/Makefile.linux or pppd/eap.c. So how is the new code ever going to even get compiled in, let alone called? It looks like you forgot to check in changes to existing files somehow.

Aside from that, there are some other problems with the commit. I would like some sort of commit message talking about what the commit achieves. It needs a Signed-off-by line as explained in Submitting-patches.md. And I would like the headline to start with "pppd: ".

I'm not sure we really need the changelog in README.eap-tls or the comments saying "Since version x.yy, this patch has ...". I think that file should just talk about the current state of the code. We have git to record how things change over time.

Finally, I would like some pointers to how I would test this code. Suppose I had two machines running pppd with EAP-TLS support, how would I create certificates, etc., and set things up so that one could authenticate itself to the other? Or fail to authenticate, so I can test the failure paths?

jjkeijser:
It seems the git commit only did a partial push. I am quite new to github so I suspect I could be doing something wrong, but I also suspect that github.com is overloaded at the moment.  Please let me know how I can send you a better version of the patch.

jjkeijser:
Hi Paul,

I've generated a new PR - as far as I can tell all files are now 
included. I've also updated the README.eap-tls file as requested.

cheers,

Jan Just Keijser

Sander80:
Hello, sorry for interrupting.
We are using this patch or its predecessor to access L2TP server from with the use os smart cards. Mostly people use Windows machines, but some people stay on Linux, and there was a lot of work to make things work properly.
So we've got a question to your patch.

Line 607 of eap-tls.c

>pkey = ENGINE_load_private_key(pkey_engine, pkey_identifier, NULL, &cb_data);

How is it supposed to work without a change to the lip11 library?
The corresponding code is
[https://github.com/OpenSC/libp11/blob/master/src/eng_back.c](https://github.com/OpenSC/libp11/blob/master/src/eng_back.c)

In the end the function 
`static int ctx_get_pin(ENGINE_CTX *ctx, const char* token_label, UI_METHOD *ui_method, void *callback_data)`
is called.

And it has 
```
 ui = UI_new_method(ui_method);
        if (!ui) {
                ctx_log(ctx, 0, "UI_new failed\n");
                return 0;
        }
```

It has no way to analyse cb_data in case no ui_method is set.

As a result one has both to patch the libp11 library to make things working, I can show the corresponding patch if needed. But could there perhaps be a way on pppd size to provide a non-null UI_method.

Sander80:
Seems like the proper way would be

 ```
       UI_METHOD* transfer_pin = UI_create_method("transfer_pin");
        int opener (UI* ui) {return 1;};
        int writer (UI *ui, UI_STRING *uis) {
            PW_CB_DATA* cb_data = (PW_CB_DATA*)UI_get0_user_data(ui);
            UI_set_result(ui, uis, cb_data->password);
            return 1;
        };
        int flusher (UI* ui) {return 1;};
        int reader (UI *ui, UI_STRING *uis) {return 1;};
        int closer (UI* ui) {return 1;};

        UI_method_set_opener(transfer_pin, opener);
        UI_method_set_writer(transfer_pin, writer);
        UI_method_set_flusher(transfer_pin, flusher);
        UI_method_set_reader(transfer_pin, reader);
        UI_method_set_closer(transfer_pin, closer);

        pkey = ENGINE_load_private_key(pkey_engine, pkey_identifier, transfer_pin, &cb_data);

```

What do you think about it?

**Comment by jjkeijser** [comment]
Hi,

On 10/05/20 16:24, Sander80 wrote:
>
> Hello, sorry for interrupting.
> We are using this patch or its predecessor to access L2TP server from 
> with the use os smart cards. Mostly people use Windows machines, but 
> some people stay on Linux, and there was a lot of work to make things 
> work properly.
> So we've got a question to your patch.
>
> Line 607 of eap-tls.c
>
>     pkey = ENGINE_load_private_key(pkey_engine, pkey_identifier, NULL,
>     &cb_data);
>
> How is it supposed to work without a change to the lip11 library?
> The corresponding code is
> https://github.com/OpenSC/libp11/blob/master/src/eng_back.c
>
> In the end the function
> |static int ctx_get_pin(ENGINE_CTX *ctx, const char* token_label, 
> UI_METHOD *ui_method, void *callback_data)|
> is called.
>
> And it has
>
> |ui = UI_new_method(ui_method); if (!ui) { ctx_log(ctx, 0, "UI_new 
> failed\n"); return 0; } |
>
> It has no way to analyse cb_data in case no ui_method is set.
>
> As a result one has both to patch the libp11 library to make things 
> working, I can show the corresponding patch if needed. But could there 
> perhaps be a way on pppd size to provide a non-null UI_method.
>
>
I've never had to patch libp11/engine_pkcs11 in the past for this. Also, 
note that I pass ui_method=NULL, which will then result in
   ui = UI_new_method(NULL);

which OpenSSL then will/should translate to

  23 UI *UI_new_method(const UI_METHOD *method)
[...]
  38
  39     if (method == NULL)
  40         method = UI_get_default_method();


(from OpenSSL 1.1.1g crypto/ui/ui_lib.c)
Thus, I don't see any reason why I cannot pass ui_method=NULL ; I will 
test my code later this week using a pkcs11 device just to make sure.

cheers,

JJK / Jan Just Keijser

**Comment by Sander80** [comment]
But what would be the reason of calling the default method? Up to my inderstanding in should open a window requesting a password. But I see two problems here:

1) pppd might be running in a way not being able to run GUI
2) You already should have the password up to my understanding!
```
(a few lines above in your code) 
EVP_PKEY   *pkey = NULL;
PW_CB_DATA  cb_data;
cb_data.password = passwd;
cb_data.prompt_info = pkey_identifier;
```
In the way we use it pppd first obtains the password with the passprompt.so plugin. So why would one ask for a password again?

Sander80:
From the openssl documentation:

> UI_add_user_data() adds a user data pointer for the method to use at any time. The builtin UI method doesn't care about this info. Note that several calls to this function doesn't add data, it replaces the previous blob with the one given as argument.

Hence if you already have a password, passing it via callback data and calling the default method makes no sense.

jjkeijser:
Hi,

On 11/05/20 12:55, Sander80 wrote:
>
> From the openssl documentation:
>
>     UI_add_user_data() adds a user data pointer for the method to use
>     at any time. The builtin UI method doesn't care about this info.
>     Note that several calls to this function doesn't add data, it
>     replaces the previous blob with the one given as argument.
>
> Hence if you already have a password, passing it via callback data and 
> calling the default method makes no sense.
>
>
I've managed to (re)produce the issue where the supplied password for a 
pkcs#11 device is not picked up from the 'cb_data' structure. This is 
most likely due a newer version of the libp11 library. The patch 
proposed by Sander80 does look sound and fixes the issue. I will update 
my code and patch shortly.

JJK / Jan Just Keijser


||

Sander80:
Nice )
I think I only forgot to call
`UI_destroy_method(transfer_pin)`
in the end

Sander80:
One more thought. My code should be invoked only if the passord is already there.
If not, the default method of obtaining the password should do.

Sander80:
But should perhaps this pull request be forgotten in favor of this?
https://github.com/paulusmack/ppp/pull/144

Neustradamus:

PR #143, opened 2020-04-04
Import lcp-echo-adaptive from openwrt.

alvardes:
Port Debians adaptive LCP echo patch to pppd, make it configurable and
enable it by default.

When adaptive LCP echo is enabled, LCP echo requests are only sent if the
link is idle, this avoids the common situation where a congested PPP link
(e.g. during torrenting) is falsely detected as disconnected because the
LCP replies are not received in time.

Signed-off-by: Marco d'Itri <md@linux.it>

### DIFF
--- a/pppd/lcp.c
+++ b/pppd/lcp.c
@@ -72,6 +72,7 @@ static void lcp_delayed_up __P((void *));
  */
 int	lcp_echo_interval = 0; 	/* Interval between LCP echo-requests */
 int	lcp_echo_fails = 0;	/* Tolerance to unanswered echo-requests */
+bool	lcp_echo_adaptive = 0;	/* request echo only if the link was idle */
 bool	lax_recv = 0;		/* accept control chars in asyncmap */
 bool	noendpoint = 0;		/* don't send/accept endpoint discriminator */
 
@@ -150,6 +151,8 @@ static option_t lcp_option_list[] = {
       OPT_PRIO },
     { "lcp-echo-interval", o_int, &lcp_echo_interval,
       "Set time in seconds between LCP echo requests", OPT_PRIO },
+    { "lcp-echo-adaptive", o_bool, &lcp_echo_adaptive,
+      "Suppress LCP echo requests if traffic was received", 1 },
     { "lcp-restart", o_int, &lcp_fsm[0].timeouttime,
       "Set time in seconds between LCP retransmissions", OPT_PRIO },
     { "lcp-max-terminate", o_int, &lcp_fsm[0].maxtermtransmits,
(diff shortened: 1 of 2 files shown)


gizero:
>Also bump the copyright year in the Makefile, remove a redundant maintainer
entry and fix the shell processing of the keepalive option when the two-
value syntax is used.

The commit does not seem to include the changes mentioned by the above paragraph of the commit message.

alvardes:
Hi @gizero I updated the commit message.

paulusmack:
No sign-off? Who wrote the patch originally?

alvardes:
@paulusmack 
>No sign-off? Who wrote the patch originally?

Hm.. This is a very good question. The debian patch doesn't seem to have authore information. 

By searching I found this: 

https://github.com/probonopd/Siedle_SG_650-0/blob/master/ptxdist-2016.01.0/patches/ppp-2.4.5/0003-adaptive_echos.patch

There the same patch begins with:

```
From: Marco d'Itri <md@linux.it>
Date: Wed, 25 Jan 2012 19:14:26 +0100
Subject: [PATCH] adaptive_echos

---
```

Maybe Marco d'Itri is the original author. I emailed him to try to find out any information about what mailing list was this and who was the original author.

gizero:
@alvardes Any feedback from Marco on this?

alvardes:
@gizero No, he has not responded yet.

rfc1036:
My apologies for the delay. This patch was added to the Debian package pre-git and I really have no memory of it. :-(

Let's examine the options here:
- I lifted it from some other distribution: apparently not, since I cannot find it elsewhere.
- Somebody sent it to the related [Debian bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=14331): no, it is not there.
- Somebody sent it to me by private mail: hard to believe, because it is not plausible that I would not credit the author either in the patch file or in the Debian changelog.
- I wrote it myself: the code is not complex and it looks like the kind of hyper-optimization that I would have cared for.

So I think that I can safely take credit for this code, please consider it to be licensed under the same terms of ppp.

`Signed-off-by: Marco d'Itri <md@linux.it>`

gizero:
Hi @alvardes Would you mind updating the PR with the above `Signed-off-by` line by @rfc1036 so it can move forward?

alvardes:
@gizero Sorry I am not really familiar with the process. I have updated the commit message.

gizero:
@paulusmack Looks like we do have attribution to the original patch author now...

paulusmack:
I merged it, and I changed the commit message to be more readable (and not mention openwrt, since this patch came from Debian in fact).

I put Marco as the author since he originally wrote the code.

Neustradamus:
Merged here: https://github.com/paulusmack/ppp/commit/ddd57c279f467e18c4fb3a661f21a069a5a37a1f

PR #144, opened 2020-04-22
pppd: Added EAP-TLS support, supporting up to TLSv1.3

jjkeijser:
pppd: Added EAP-TLS support based on ppp-2.4.7-eaptls-mppe-1.300.patch with support for

- EAP-TLS authentication using X509 certificates on both client and server side
- TLS v1 - v1.3 (depending on OpenSSL library support)
- PKCS#11 hardware tokens
- pppd command-line options to specify certificate, private key, CA certificate file, CA certificate path
- with OpenSSL support now included, you can switch between using the OpenSSL provided MD5 & SHA1 functions or the code supplied by pppd itself using Makefile switches.
- This patch also addresses CVE-2018-11574 and also build on Solaris (x86 tested).

For documentation on how to use and test this patch, see
[https://www.nikhef.nl/~janjust/ppp/documentation.html](url)
[https://www.nikhef.nl/~janjust/ppp/doc_pki_setup.html](url)
[https://www.nikhef.nl/~janjust/ppp/doc_pptp_serversetup.html](url)
[https://www.nikhef.nl/~janjust/ppp/doc_pptp_clientsetup.html](url)

### DIFF
--- a/README.eap-tls
+++ b/README.eap-tls
@@ -1,0 +1,230 @@
+EAP-TLS authentication support for PPP
+======================================
+
+1. Intro
+
+    The Extensible Authentication Protocol (EAP; RFC 3748) is a
+    security protocol that can be used with PPP.  It provides a means
+    to plug in multiple optional authentication methods.
+
+    Transport Level Security (TLS; RFC 5216) provides for mutual 
+    authentication, integrity-protected ciphersuite negotiation and 
+    key exchange between two endpoints.  It also provides for optional
+    MPPE encryption.
+
+    EAP-TLS (RFC 2716) incapsulates the TLS messages in EAP packets,
+    allowing TLS mutual authentication to be used as a generic EAP
+    mechanism. It also provides optional encryption using the MPPE
+    protocol.
+
+    This patch provide EAP-TLS support to pppd.
+    This authentication method can be used in both client or server
+    mode.
+
+2. Building
+
+    To build pppd with EAP-TLS support, OpenSSL (http://www.openssl.org)
+    is required. Any version from 0.9.7 should work.
+    
+    Configure, compile, and install as usual. 
+
+3. Configuration
+
+    On the client side there are two ways to configure EAP-TLS:
+
+    1. supply the appropriate 'ca', 'cert' and 'key' command-line parameters
+
+    2. edit the /etc/ppp/eaptls-client file.
+    Insert a line for each system with which you use EAP-TLS.
+    The line is composed of this fields separated by tab:
+
+      - Client name 
+        The name used by the client for authentication, can be *
+      - Server name
+        The name of the server, can be *
+      - Client certificate file 
+        The file containing the certificate chain for the 
+        client in PEM format
+      - Server certificate file
+        If you want to specify the certificate that the 
+        server is allowed to use, put the certificate file name.
+        Else put a dash '-'.
+      - CA certificate file
+        The file containing the trusted CA certificates in PEM
+        format.
+      - Client private key file
+        The file containing the client private key in PEM format.
+
+
+    On the server side edit the /etc/ppp/eaptls-server file.
+    Insert a line for each system with which you use EAP-TLS.
+    The line is composed of this fields separated by tab:
+
+      - Client name
+        The name used by the client for authentication, can be *
+      - Server name
+        The name of the server, can be *
+      - Client certificate file
+        If you want to specify the certificate that the
+        client is allowed to use, put the certificate file name.
+        Else put a dash '-'.
+      - Server certificate file
+        The file containing the certificate chain for the
+        server in PEM format
+      - CA certificate file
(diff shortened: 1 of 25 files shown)


paulusmack:
Still no Signed-off-by?

paulusmack:
Some more comments on this PR...

We seem to end up with -lcrypto appearing 4 times on the command line for linking pppd. Does it really need to be given 4 times?

I'm not a fan of ifdefs in code, and this patch seems to have a lot of them, particularly in eap.c.

There is a change to sys-solaris.c, adding a netif_get_mtu() function, which seems to be entirely unused elsewhere in the patch. Is this because we are now forcing MPPE on in the Solaris builds, and we weren't before?

There are changes to the way SHA1 is done which seem like they should be a separate patch, not mixed in with the EAP-TLS implementation.

paulusmack:
Why are there changes to code that uses MD5? Does the use of our own copy of MD5 create some sort of conflict or other problem when linking in openssl?

jjkeijser:
On 25/05/20 05:43, Paul Mackerras wrote:
>
> Still no Signed-off-by?
>
>
hmmm strange, I will try to fix that on the next attempt.

JJK

jjkeijser:
On 25/05/20 07:33, Paul Mackerras wrote:
>
> Some more comments on this PR...
>
> We seem to end up with -lcrypto appearing 4 times on the command line 
> for linking pppd. Does it really need to be given 4 times?
>
ah , nice catch, I will restructure the Makefile stuff to only add 
-lcrypto and -lssl once
>
> I'm not a fan of ifdefs in code, and this patch seems to have a lot of 
> them, particularly in eap.c.
>
I fully understand, but the patch merely follows what was done for the 
SRP patch (#ifdef USE_SRP). If you compare the two:

$  grep -c "ifdef.*USE_SRP" eap.c
20

$ grep -c "ifdef.*USE_EAPTLS" eap.c
21

But if you'd rather not see #ifdefs then I can happily remove them

> There is a change to sys-solaris.c, adding a netif_get_mtu() function, 
> which seems to be entirely unused elsewhere in the patch. Is this 
> because we are now forcing MPPE on in the Solaris builds, and we 
> weren't before?
>

correct, the sys-solaris.c change is needed only to allow the code to 
build on Solaris with MPPE enabled; note that my patch does not build 
with MPPE *dis*abled.
>
> There are changes to the way SHA1 is done which seem like they should 
> be a separate patch, not mixed in with the EAP-TLS implementation.
>
>
Because the EAP-TLS patch so heavily depends on OpenSSL, I added the 
option to use ALL SHA1 (and MD5) functions from the OpenSSL library; 
this also explains why I had to change the MD5 code slightly, as there 
are some minor differences between the function calls in the ppp code 
versus what OpenSSL expects.

cheers,

JJK

jjkeijser:
On 25/05/20 10:15, Paul Mackerras wrote:
>
> Why are there changes to code that uses MD5? Does the use of our own 
> copy of MD5 create some sort of conflict or other problem when linking 
> in openssl?
>
>

My code does not use MD5 directly; this patch allows a user/developer to 
use either the internal MD5 code or OpenSSL's MD5 functions.

Let me break up my PR into pieces:
1) use OpenSSL MD5+SHA1 routines versus internal routines
2) add missing function so that MPPE builds on Solaris
3) EAP-TLS support

Note that patch 3) will require  patches (1)+(2) to compile

I will generate new PR's for each of them shortly.

cheers,

JJK / Jan Just Keijser

paulusmack:
> On 25/05/20 10:15, Paul Mackerras wrote: Why are there changes to code that uses MD5? Does the use of our own copy of MD5 create some sort of conflict or other problem when linking in openssl?
> My code does not use MD5 directly; this patch allows a user/developer to use either the internal MD5 code or OpenSSL's MD5 functions.

My concern is that using the openssl function means we have the risk that the openssl project will make some decision like "applications shouldn't call MD5 functions directly" and stop exporting them, expecting applications to be recoded to use some higher level interface which can redirect to other hash functions. Which might be fine for many applications which just need some sort of suitable hash function, but not for pppd where we specifically need MD5. Since MD5 is not a lot of code, having our own implementation costs very little and insulates us from openssl interface changes.

> Let me break up my PR into pieces: 1) use OpenSSL MD5+SHA1 routines versus internal routines 2) add missing function so that MPPE builds on Solaris 3) EAP-TLS support

That would be good.

> Note that patch 3) will require  patches (1)+(2) to compile I will generate new PR's for each of them shortly. 

You don't need three PRs; you can do one pull request for three commits on a single branch, which then automatically preserves the order of the commits.

> cheers, JJK / Jan Just Keijser

jjkeijser:
Please close this pull request as my git repo is a total mess.; I cannot do three new signed-off commits without recreating the entire PR.

Neustradamus:

PR #145, opened 2020-04-25
Add ipv6cp-accept-remote option

sthibaul:
From Fedora

### DIFF
--- a/pppd/ipv6cp.c
+++ b/pppd/ipv6cp.c
@@ -245,6 +245,8 @@ static option_t ipv6cp_option_list[] = {
 
     { "ipv6cp-accept-local", o_bool, &ipv6cp_allowoptions[0].accept_local,
       "Accept peer's interface identifier for us", 1 },
+    { "ipv6cp-accept-remote", o_bool, &ipv6cp_allowoptions[0].accept_remote,
+      "Accept peer's interface identifier for itself", 1 },
 
     { "defaultroute6", o_bool, &ipv6cp_wantoptions[0].default_route,
       "Add default IPv6 route", OPT_ENABLE|1, &ipv6cp_allowoptions[0].default_route },
@@ -444,6 +446,7 @@ ipv6cp_init(unit)
     memset(ao, 0, sizeof(*ao));
 
     wo->accept_local = 0;
+    wo->accept_remote = 0;
     wo->neg_ifaceid = 1;
     ao->neg_ifaceid = 1;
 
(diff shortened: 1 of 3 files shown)


paulusmack:
The change looks fine, but I can't merge this commit as-is because it lacks a Signed-off-by line. Are you the author of the code? If not, who is?

sthibaul:
I am not the author. I digged back in the fedora source:

https://apps.fedoraproject.org/packages/ppp/sources/spec/

* Fri Nov  4 2005 David Woodhouse <dwmw2@redhat.com> 2.4.3-5
- Implement ipv6cp-accept-remote option

I added a From: line accordingly.

sthibaul:
Ah right I can commit under his name, now done so.

pali:
@sthibaul You enabled this `accept_remote` option by default. So how it can be turned off, to mimic behavior prior applying your patch?

sthibaul:
@pali I was about to say "well, just like ipv6cp-accept-local", but I don't see how that one would be working either indeed. I'll turn them into no- option

pali:
For IPv4 is `ipcp-accept-local` turned off by default. But for IPv6 in this patch you have `ipv6cp-accept-local` enabled by default and I do not see any mechanism either for IPv4 or IPv6 to turn *-accept-local option off.

So, why you are turning `ipv6cp-accept-local` to on by default at all?

sthibaul:
The patch in this PR is *not* touching `ipv6cp-accept-local`: it seems it has been enabled by default for as much that git can easily remember.

> So, why you are turning ipv6cp-accept-local to on by default at all?

I'm not.

But possibly we actually want to change the `ipv6cp-accept-local` default actually, to behave like ipv4 and ipx.

pali:
> The patch in this PR is not touching `ipv6cp-accept-local`:

Sorry, I mean `*-remote` options.

> `ipv6cp-accept-local`: it seems it has been enabled by default for as much that git can easily remember.

Yea, looks like this option is somehow broken. It is always enabled despite if user enabled it or not.

> But possibly we actually want to change the `ipv6cp-accept-local` default actually, to behave like ipv4 and ipx.

Looks like a good idea. So IPv6 accept local option would behave in same way as IPv4 and IPX.

sthibaul:
done so.

sthibaul:
I'm however thinking that ipv6cp.c is missing the bits that set accept_local/remote in ipv6cp_resetci when the address is not specified by the user. Which is possibly why the original authors just set it to 1 up front.

sthibaul:
There it is.

pali:
I have some patches which address these problems. But they needs working `ipv6cp-accept-remote` option, basically functionality from this pull request. I will create pull request after this one is merged.

pali:
> There it is.

Great! If it is working fine, I think it is ready for merge.

Neustradamus:
@sthibaul, @pali: Good job!

@paulusmack: Good for you?

Neustradamus:
@paulusmack: Thanks a lot for the merging and all other guys who have participated on it!

PR #150, opened 2020-05-25
Disable asking password again when prompt return 128 (quit connection)

Sander80:
Greetings!
There is a suggestion coming from our group.
There's the password prompt plugin, and it might be asking for a password up to 10 times.
But what if the one using the callback app hits cancel or whatever and does not want to enter the password?
That would be nice to let the plugin know, that no more requests are needed.
For this we proposed the 128 return code but it can be whatever you prefer.

What do you think of that?

### DIFF
--- a/pppd/plugins/passprompt.c
+++ b/pppd/plugins/passprompt.c
@@ -32,7 +32,7 @@ static int promptpass(char *user, char *passwd)
     int readgood, wstat;
     ssize_t red;
 
-    if (promptprog[0] == 0 || access(promptprog, X_OK) < 0)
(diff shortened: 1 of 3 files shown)


paulusmack:
The idea is fine.

I don't like the way it sets the asked_to_quit variable. The passprompt plugin should just have an internal variable that records the fact that the promptprog program returned a 128 exit status, and return -1 from promptpass() if it is set.

The commit has no description, no sign-off and multiple spelling errors, so I will not merge it.

Sander80:
Thank you for the answer, I tried to fix it and make a new request from a new fork (the old one was dropped)
https://github.com/paulusmack/ppp/pull/161

PR #151, opened 2020-05-27
Add support for EAP-TLS (including experimental TLS v1.3 support)

jjkeijser:
Add support for EAP-TLS (including experimental TLS v1.3 support).
The Extensible Authentication Protocol (EAP; RFC 3748) is a security protocol that can be used with PPP. It provides a means to plug in multiple optional authentication methods.

Transport Level Security (TLS; TLSv1.3 RFC 8446 and TLSv1.2 RFC 5246) provides for mutual authentication, integrity-protected ciphersuite negotiation and key exchange between two endpoints. It also provides for optional MPPE encryption.

EAP-TLS (RFC 5216 obsoleting RFC2716) encapsulates the TLS messages in EAP packets, allowing TLS mutual authentication to be used as a generic EAP mechanism.

TLSv1.3 support is based on these drafts
- [https://tools.ietf.org/html/draft-ietf-emu-eap-tls13-09](url)
- [https://tools.ietf.org/html/draft-dekok-emu-tls-eap-types-00](url)

and was tested using a Linux client and server only.

This patch is based on the latest patch
  [https://www.nikhef.nl/~janjust/ppp/ppp-2.4.8-eaptls-mppe-1.301.patch]
and is stripped down to support only OpenSSL 1.0.1+ builds. 

To test EAP-TLS support, read & follow the documentation at

- [https://www.nikhef.nl/~janjust/ppp/doc_pki_setup.html](url)
- [https://www.nikhef.nl/~janjust/ppp/doc_pptp_serversetup.html](url)
- [https://www.nikhef.nl/~janjust/ppp/doc_pptp_clientsetup.html](url)
- [https://www.nikhef.nl/~janjust/ppp/doc_pptp_winclientsetup.html](url)

Signed-off-by: Jan Just Keijser <jan.just.keijser@gmail.com>

### DIFF
--- a/pppd/sys-linux.c
+++ b/pppd/sys-linux.c
@@ -2757,7 +2757,7 @@ int sif6addr (int unit, eui64_t our_eui64, eui64_t his_eui64)
     memset(&ifr6, 0, sizeof(ifr6));
     IN6_LLADDR_FROM_EUI64(ifr6.ifr6_addr, our_eui64);
     ifr6.ifr6_ifindex = ifr.ifr_ifindex;
-    ifr6.ifr6_prefixlen = 10;
+    ifr6.ifr6_prefixlen = 128;
 
     if (ioctl(sock6_fd, SIOCSIFADDR, &ifr6) < 0) {
 	error("sif6addr: ioctl(SIOCSIFADDR): %m (line %d)", __LINE__);
@@ -2768,7 +2768,7 @@ int sif6addr (int unit, eui64_t our_eui64, eui64_t his_eui64)
     memset(&rt6, 0, sizeof(rt6));
     IN6_LLADDR_FROM_EUI64(rt6.rtmsg_dst, his_eui64);
     rt6.rtmsg_flags = RTF_UP;
-    rt6.rtmsg_dst_len = 10;
+    rt6.rtmsg_dst_len = 128;
     rt6.rtmsg_ifindex = ifr.ifr_ifindex;
     rt6.rtmsg_metric = 1;
 
@@ -2805,7 +2805,7 @@ int cif6addr (int unit, eui64_t our_eui64, eui64_t his_eui64)
     memset(&ifr6, 0, sizeof(ifr6));
     IN6_LLADDR_FROM_EUI64(ifr6.ifr6_addr, our_eui64);
     ifr6.ifr6_ifindex = ifr.ifr_ifindex;
-    ifr6.ifr6_prefixlen = 10;
+    ifr6.ifr6_prefixlen = 128;
 
     if (ioctl(sock6_fd, SIOCDIFADDR, &ifr6) < 0) {
 	if (errno != EADDRNOTAVAIL) {


Sander80:
Greetings. We are using the code of jjkeijser to connect to our servers with the use of pkcs11 hardware tokens (vpn-based lt2p ipsec solution). It works originally on Windows, but with this code and our gui tool [https://github.com/Sander80/l2tp-ipsec-vpn](https://github.com/Sander80/l2tp-ipsec-vpn) it can be used on Linux as well. After recent changes the code of jjkeijser can be used without any extra changes. Thus we are interested in getting the code into the official ppp repository too.

paulusmack:
I'm curious why the EAP-TLS code is tied to the use of PPTP. Is there any essential connection, or is it just that PPTP is one case where you might want EAP-TLS rather than something else?

jjkeijser:
the EAP-TLS code was originally developed to allow the user of certificates for PPTP VPNs; however, the code is not tied to PPTP itself and one does not need to use PPTP at all. @Sander80 uses it for L2TP VPNs, for example but you can also use it without any other VPN software, e.g. as follows
    ./pppd file options-pppossh-client-eaptls pty "ssh -t root@<server> /usr/sbin/pppd file /root/options-pppossh-server-eaptls"

with attached options-pppossh files.
[options-pppossh-client-eaptls.txt](https://github.com/paulusmack/ppp/files/4721686/options-pppossh-client-eaptls.txt)
[options-pppossh-server-eaptls.txt](https://github.com/paulusmack/ppp/files/4721687/options-pppossh-server-eaptls.txt)

lo1ol:
EAP-TLS should be embedded for the following reasons:
1.	New code doesn’t break the structure of the old code and can be disabled. Even if pppd is complied with EAP-TLS support, EAP-TLS could still be refused via configure file.
2.	EAP-TLS is more secure. For example, in the old algorithms, such as CHAP, the password for auth may be broken via brute force. Otherwise, cryptography used in EAP-TLS doesn’t have any known practical vulnerabilities. 
3.	EAP-TLS is more convenient. You don’t need to remember strong password for your user, you can just put your file containing the key in directory with the proper permissions and use it.  Furthermore, you can keep your certificates and keys on your smart card and use them everywhere.
4.	Other VPN implementations have been using PKI infrastructure for authentication for a long time (OpenVPN, IPsec/IKE and so on)

paulusmack:
In reviewing this I notice that none of the commits have a Signed-off-by line. This is a show-stopper.

Please re-do the commits with a Signed-off-by on each one (as described in Submitting-patches.md).

Sander80:
Greetings and thank you for feedback. While having made a few commits, I am not the author of the pull request or the mane code being proposed. But I would like to help somehow if it is possible, at least by clarifying some things.

I searched for merged commints, found an example (https://github.com/paulusmack/ppp/pull/148). If I am not mistaken, this signed-off part is only a part of the commit message, right?

If so do you think it makes sense to rewrite history by forcing an overwrite of rewritten commits or should it be a new series of commits in a new branch?

jjkeijser:
I've done a 'git commit -s -amend' on each of the earlier commits but only the last one shows up as "signed-off-by"

paulusmack:
Amending a commit changes its SHA1 ID, so the commits that were previously descendants are no longer descendants. You would need to do git cherry-pick -e rather than git commit --amend on all except the oldest commit.

jjkeijser:
On 30/06/20 03:51, Paul Mackerras wrote:
>
> Amending a commit changes its SHA1 ID, so the commits that were 
> previously descendants are no longer descendants. You would need to do 
> git cherry-pick -e rather than git commit --amend on all except the 
> oldest commit.
>
>
Just tried it and cannot get it to work. This is taking too much of my 
time already.

If no one objects I will cancel this PR and re-submit a new PR based on 
the current commits in my branch, combined with the patches from 
@Sander80 and Petr Mikhalitsy. This will alter the history of the 
commits but it seems the easiest way forward.

Any objections?

Sander80:
No objections from my side. You can author our commits as well, they are
anyway small compared to your code.

вт, 30 июн. 2020 г., 17:19 jjkeijser <notifications@github.com>:

jjkeijser:
I think I managed to sign off all the commits while keeping the history.

paulusmack:
I have some more questions and comments on the code, having looked through it again.

* The top commit seems to be a revert of the second-top commit. Why did you bother to include them when you were rebasing? Could we just drop both of those top two commits?

* In auth.c you add options that are not privileged. I think that is safe because those options are only used on the client side of the authentication. Is that your reasoning?

* Why would you want to require the peer to authenticate the local machine? What advantage does that bring? It would be good to explain the reason in the section on need-peer-eap in pppd.8.

* There is some code which looks strange to me in have_eaptls_secret_server():

```
+    if (client != NULL && client[0] == 0)
+		client = NULL;
+    else if (server != NULL && server[0] == 0)
+		server = NULL;
```

Why is that "else" in there? I realize it is copied from have_chap_secret(), and the code there goes back to 1993, and I don't remember why it was done like that. I think at the time it wasn't possible to have both client and server be the null string, but in your code that is possible as far as I can see.

* In some places you use strlcpy, which is good, but in others you are still using strncpy, which makes me nervous. For instance in password_callback() you do strncpy() and then do strlen() on the destination, which is dangerous because strncpy doesn't guarantee the destination is null-terminated. You add 5 instances of strncpy and it's not clear to me that they are all safe (in fact the password_callback() one isn't). Could we change them all to strlcpy?

* Finally, the indentation looks all over the place to me. Are you using some tab width other than the standard 8? Please use 8, since that is what is used in all the rest of the code.

None of these things are necessarily an impediment to merging, but some might need to be fixed immediately after merging.

jjkeijser:
Hi Paul,

thanks for taking your time to review my patch.
See comments inline

On 12/07/20 12:39, Paul Mackerras wrote:
>
> I have some more questions and comments on the code, having looked 
> through it again.
>
>  *
>
>     The top commit seems to be a revert of the second-top commit. Why
>     did you bother to include them when you were rebasing? Could we
>     just drop both of those top two commits?
>
there's a minor difference between the top commits; I'd much rather 
re-submit the patch with the commits "collapsed".
>
>  *
>
>     In auth.c you add options that are not privileged. I think that is
>     safe because those options are only used on the client side of the
>     authentication. Is that your reasoning?
>
Correct; the options which are listed:

  253 #ifdef USE_EAPTLS
  254 char *cacert_file  = NULL;  /* CA certificate file (pem format) */
  255 char *ca_path      = NULL;  /* directory with CA certificates */
  256 char *cert_file    = NULL;  /* client certificate file (pem format) */
  257 char *privkey_file = NULL;  /* client private key file (pem format) */
  258 char *crl_dir      = NULL;  /* directory containing CRL files */
  259 char *crl_file     = NULL;  /* Certificate Revocation List (CRL) 
file (pem format) */
  260 char *max_tls_version = NULL;   /* Maximum TLS protocol version 
(default=1.2) */
  261 bool need_peer_eap = 0;         /* Require peer to authenticate us */
  262 #endif

are all unprivileged and are used (mostly) on the client side, although 
it is possible to use some of them server-side as well
>
>  *
>
>     Why would you want to require the peer to authenticate the local
>     machine? What advantage does that bring? It would be good to
>     explain the reason in the section on need-peer-eap in
>
this is an odd one and I don't recall the particular reason for it; the 
purpose of this option is to *ensure* that the other side is doing EAP 
verification - i.e. if the client does not offer EAP authentication then 
reject the connection. However, if a user would want that, {s}he could 
also simple disable the other options using refuse_* , of course.

I admit that I have hardly ever used this option (though I have tested it).
OTOH, the code impact is small ...

>
>  *
>
>     There is some code which looks strange to me in
>     have_eaptls_secret_server():
>
> |+ if (client != NULL && client[0] == 0) + client = NULL; + else if 
> (server != NULL && server[0] == 0) + server = NULL; |
>
> Why is that "else" in there? I realize it is copied from 
> have_chap_secret(), and the code there goes back to 1993, and I don't 
> remember why it was done like that. I think at the time it wasn't 
> possible to have both client and server be the null string, but in 
> your code that is possible as far as I can see.
>
you are right - this is a straight copy+paste action from 
have_chap_secret() and have_srp_secret(). The 'else' can go.

>  *
>
>     In some places you use strlcpy, which is good, but in others you
>     are still using strncpy, which makes me nervous. For instance in
>     password_callback() you do strncpy() and then do strlen() on the
>     destination, which is dangerous because strncpy doesn't guarantee
>     the destination is null-terminated. You add 5 instances of strncpy
>     and it's not clear to me that they are all safe (in fact the
>     password_callback() one isn't). Could we change them all to strlcpy?
>
Yes, strlcpy is the way to go - in a future submitted patch (if any) I 
will convert them all to strlcpy
>
>  *
>
>     Finally, the indentation looks all over the place to me. Are you
>     using some tab width other than the standard 8? Please use 8,
>     since that is what is used in all the rest of the code.
>
>

In eap-tls.[ch] I use TS=4 which might throw off the indentation; I will 
convert all tabs to spacse on the next patch submit.

jjkeijser:
As a follow-up:

- I've updated my tree by removing the tabs and replacing strncpy with strlcpy ; I have no idea what the best way is to submit this new version/tree into the main line, at the same time getting rid of the commit-junk
- to come back to the 'need-peer-eap' option: this option is useful on the client-side only and causes the client to not connect if the server does NOT want to do EAP auth.  It would be more logical to do something like
  require-eap
on the client side but that does not work, unfortunately. Adding 'require-eap'  more or less pushes pppd into "server mode" (not just for EAP-TLS, I believe).  Thus, the 'need-peer-eap' option was added. I admit it is of limited use, but it *does* have a purpose. 

What's your opinion?

paulusmack:
> As a follow-up:
> 
>     * I've updated my tree by removing the tabs and replacing strncpy with strlcpy ; I have no idea what the best way is to submit this new version/tree into the main line, at the same time getting rid of the commit-junk

You should be able to use git push -f to update your github tree to whatever you have locally, throwing away the commits on github that were previously there.

>     * to come back to the 'need-peer-eap' option: this option is useful on the client-side only and causes the client to not connect if the server does NOT want to do EAP auth.  It would be more logical to do something like
>       require-eap
>       on the client side but that does not work, unfortunately. Adding 'require-eap'  more or less pushes pppd into "server mode" (not just for EAP-TLS, I believe).  Thus, the 'need-peer-eap' option was added. I admit it is of limited use, but it _does_ have a purpose.
> 
> 
> What's your opinion?

I believe you that it has a purpose, but it seems to me that the intention is not primarily to authenticate the local machine to the remote machine.  There must be some side effect of the authentication process which is what we're really after. I imagine it could be something like the implicit authentication of the remote machine by the fact that it has presented a certificate that we accept. Or it could be that the authentication process provides key material for other protocols. Whatever it is, it would be useful to have it made explicit and explained.

jjkeijser:
I will try the "git push -f" once I am back from vacation (in a few weeks).

As for "need-peer-eap":  I will add an explanation to the docs/man page. All that flag effectively does is ensure that when a client connects to a server that dos not advertise that it wants to EAP negotiaton then the client will refuse to connect.  
If I cannot explain it any better than this in the docs/man page then I will remove the flag completely.

Neustradamus:
@jjkeijser: Any news?

Neustradamus:
@jjkeijser: Why have you closed the ticket?

@paulusmack: Can you look?

jjkeijser:
This was an attempt to collapse the commit log into a single message. Clearly, that failed.
I will resubmit the PR from scratch next week.

Neustradamus:
@paulusmack: Can you reopen this PR?

jjkeijser:
I have just created a new signed-off PR (#172)

Neustradamus:

@jjkeijser: Please edit your old ticket because all your links do not work correctly: https://github.com/paulusmack/ppp/pull/151#issue-423774618

PR #153, opened 2020-07-05
Document PPPoE options in pppd.8 manpage

pali:


paulusmack:
I think it's not quite correct to say that the ethernet interface name has to be specified with "nic-" in front. It can be, but it doesn't have to be, as long as the interface name doesn't look like any other pppd option.

pali:
I thought that it is better to have simple, clear and unambiguous documentation. People tend to do mistakes and requirement _"nic-" prefix is not required when interface name does not look like another option_ seems to be too complicated. Moreover specifying "nic-" prefix in documentation does not break anything.

paulusmack:
If it was a user guide then I would agree, but it is a man page which should be as complete and accurate as is reasonably possible.

I am also still pondering whether each plugin should have its own man page rather than trying to put them all in pppd.8, which is already rather large.

pali:
Ok! I updated documentation and put there information that `nic-` may be avoided.

pali:
@paulusmack It is fine for you now?

Neustradamus:
@paulusmack: I think it can be merged, no?

Issue #87, opened 2017-07-13
Lost CHAP authentication packets do not result in a timeout by the client


nomis:
If CHAP authentication is required with the peer but this is never completed (either because the server never sends the challenge or because the client doesn't receive the outcome) then the client will wait forever, relying on the server to terminate the connection.

This can easily happen if the CHAP challenge or outcome packet is lost for any reason.

nomis:
Comments copied from my PR that fixes the issue but has been closed:

----

> Hmmm, when the timeout occurs, I think it would be better to resend the CHAP-Response packet, instead of just giving up.

The client can't send a `CHAP Response` if it never received a `CHAP Challenge`. The server may not resend this or the retries may also be lost, so the client must timeout in this scenario.

The server may not accept a duplicate `CHAP Response` a full 60 seconds later, especially when it has already sent a `CHAP Success`.  This would add an assumption about how the other side of the link will behave. It looks like `ppp` would do this, but RFC 1334 does not require it. Only PAP explicitly specifies that retries will occur indefinitely.

----

The server is also allowed to send multiple challenges, so the client may be retrying the response to an earlier challenge if there are two consecutive challenges and multiple lost packets.

In the UK the majority of DSL lines use BT Wholesale connections that perform the CHAP authentication using RADIUS and then hand-off to the PPP via L2TP to a remote server which then (usually) starts LCP again and does another CHAP authentication.

It's possible for the remote server to simply do a second `CHAP Challenge` (which is also lost). In that scenario the client would be retransmitting the wrong `CHAP Response` because the challenge value will be different.

It's possible for the remote server to start IPCP/IPV6CP without restarting LCP. In that scenario the client would be retransmitting the `CHAP Response` to a device that may not have the original challenge value and may not respond appropriately.

----

Here's a comment on the implementation used by [FireBrick](https://www.firebrick.co.uk/):
> The system handles failure of a chap packet at any stage quite well with the appropriate resend. A whole minute may be too long though.

paulusmack:
Fixed by a72f809b1dfe.

PR #160, opened 2020-08-06
pppd: Add client CHAP authentication timeout

nomis:
If CHAP authentication is required with the peer but this is never
completed (either because the server never sends the challenge or
because the client doesn't receive the outcome) then the client
will wait forever, relying on the server to terminate the connection.

There are options for server side retries but a client side timeout
option is required to prevent the client from getting stuck if the
server won't terminate the connection. This is defaulted to 60 seconds.

----

>  I think it would be better to resend the CHAP-Response packet, instead of just giving up.

The client can't send a `CHAP Response` if it never received a `CHAP Challenge`. The server may not resend this or the retries may also be lost, so the client must timeout in this scenario.

The server may not accept a duplicate `CHAP Response` a full 60 seconds later, especially when it has already sent a `CHAP Success`.  This would add an assumption about how the other side of the link will behave. It looks like `ppp` would do this, but RFC 1334 does not require it. Only PAP explicitly specifies that retries will occur indefinitely.

----

I'd really like to get this merged to fix the problem where pppd hangs indefinitely when a single `CHAP Challenge` or `CHAP Success` packet is lost (#87).

### DIFF
--- a/pppd/chap-new.c
+++ b/pppd/chap-new.c
@@ -55,20 +55,23 @@ int (*chap_verify_hook)(char *name, char *ourname, int id,
 /*
  * Option variables.
  */
-int chap_timeout_time = 3;
+int chap_server_timeout_time = 3;
 int chap_max_transmits = 10;
 int chap_rechallenge_time = 0;
+int chap_client_timeout_time = 60;
 
 /*
  * Command-line options.
  */
 static option_t chap_option_list[] = {
-	{ "chap-restart", o_int, &chap_timeout_time,
-	  "Set timeout for CHAP", OPT_PRIO },
+	{ "chap-restart", o_int, &chap_server_timeout_time,
+	  "Set timeout for CHAP (as server)", OPT_PRIO },
 	{ "chap-max-challenge", o_int, &chap_max_transmits,
 	  "Set max #xmits for challenge", OPT_PRIO },
 	{ "chap-interval", o_int, &chap_rechallenge_time,
 	  "Set interval for rechallenge", OPT_PRIO },
+	{ "chap-timeout", o_int, &chap_client_timeout_time,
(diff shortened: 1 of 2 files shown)


paulusmack:
> The client can't send a `CHAP Response` if it never received a `CHAP Challenge`. The server may not resend this or the retries may also be lost, so the client must timeout in this scenario.

Right, I agree with giving up if no challenge is received in 60 seconds.

> The server may not accept a duplicate `CHAP Response` a full 60 seconds later, especially when it has already sent a `CHAP Success`. This would add an assumption about how the other side of the link will behave. It looks like `ppp` would do this, but RFC 1334 does not require it. Only PAP explicitly specifies that retries will occur indefinitely.

Actually RFC1334 says this:

```
         Implementation Note: Because the Success might be lost, the
         authenticator MUST allow repeated Response packets after
         completing the Authentication phase.  To prevent discovery of
         alternative Names and Secrets, any Response packets received
         having the current Challenge Identifier MUST return the same
         reply Code returned when the Authentication phase completed
         (the message portion MAY be different).  Any Response packets
         received during any other phase MUST be silently discarded.
```

I will probably merge your commit, but I still think the response should be retransmitted (with a smaller timeout, say 3s) if no success/failure packet is received.

nomis:
In the UK it can take sometimes average 7s or take up to 30s to get a response, so I'd be reluctant to resend earlier than 60s because it could cause interoperability problems.

Neustradamus:
@SimonTate: What do you think about this PR?

SimonTate:
@Neustradamus , I agree that this is a good change and 60s sounds like a reasonable default timeout to me. At the end of the day, this is configurable for anyone wanting to speed it up for their environment.

+1 from me.

paulusmack:
I'll merge this since it is an improvement even without the retransmission of the CHAP response.

PR #162, opened 2020-08-07
Fix setting prefix for IPv6 link-local addresses

pali:
PPP IPV6CP protocol exchanges only IPv6 Interface Identifiers which are
used for generating IPv6 link-local addresses. As on PPP link cannot be
more hosts and IPv6 link-local addresses are designed for addressing a
single link, set route prefix for local and remote IPv6 link-local
addresses to /128 which means single IPv6 address.

This ensures that Linux kernel would not try to route other IPv6 link-local
addresses over PPP link.

Routable IPv6 prefix on PPP link is later configured either by ICMPv6
Router Advertisement packets or manually/statically. Routable local IPv6
address is then assigned by SLAAC, DHCPv6 or manual/static configuration.

So this change has no effect for routable IPv6 address or routable IPv6
prefix.

Note that pppd for IPv4 already sets netmask to 255.255.255.255, one single
IPv4 address.

Fixes: https://github.com/paulusmack/ppp/issues/121

### DIFF
--- a/pppd/sys-linux.c
+++ b/pppd/sys-linux.c
@@ -2757,7 +2757,7 @@ int sif6addr (int unit, eui64_t our_eui64, eui64_t his_eui64)
     memset(&ifr6, 0, sizeof(ifr6));
     IN6_LLADDR_FROM_EUI64(ifr6.ifr6_addr, our_eui64);
     ifr6.ifr6_ifindex = ifr.ifr_ifindex;
-    ifr6.ifr6_prefixlen = 10;
+    ifr6.ifr6_prefixlen = 128;
 
     if (ioctl(sock6_fd, SIOCSIFADDR, &ifr6) < 0) {
 	error("sif6addr: ioctl(SIOCSIFADDR): %m (line %d)", __LINE__);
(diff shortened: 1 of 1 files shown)


Neustradamus:
@paulusmack: Thanks a lot!

leonid-s-usov:
Hello!

Got here accidentally via an internet search, but having read the description I have an issue with it

> This ensures that Linux kernel would not try to route other IPv6 link-local addresses over PPP link.

There is no such thing as global routing of link-local addresses. For every such case, an explicit interface is required in addition to a link-local address for the packet to be sent. 

So, although it does make sense to mask the destination as /128, it not for the mentioned routability reason - rather because this is the nature of link-local addresses.

pali:
> For every such case, an explicit interface is required in addition to a link-local address for the packet to be sent.

IIRC there is exception when explicit inteface is not required: If you have in system only one interface with IPv6 address.

leonid-s-usov:
> > For every such case, an explicit interface is required in addition to a link-local address for the packet to be sent.
> 
> IIRC there is exception when explicit inteface is not required: If you have in system only one interface with IPv6 address.

It doesn't change the point. It's rather a "syntactic sugar" that implies the interface if it isn't explicitly mentioned

PR #165, opened 2020-09-01
pppd: Do not call update_link_stats() for every LcpSendEchoRequest() call

pali:
Function update_link_stats() is doing more than reading number of received
bytes, e.g. it changes state of link_stats_valid.

This change replace update_link_stats() by get_ppp_stats() in
LcpSendEchoRequest() function to avoid any side effects.

### DIFF
--- a/pppd/lcp.c
+++ b/pppd/lcp.c
@@ -2339,12 +2339,10 @@ LcpSendEchoRequest (f)
      */
     if (lcp_echo_adaptive) {
 	static unsigned int last_pkts_in = 0;
+	struct pppd_stats cur_stats;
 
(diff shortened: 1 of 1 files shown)


paulusmack:
Is this fixing an actual observed bug, or does it just seem like a good idea?

pali:
I have not observed any bug related this option yet (does not mean that it does not exist), so for now it is like just a good idea.

pali:
Basically this patch is calling "proper" method for retrieving number of incoming packets. Previous code without this patch is doing more things which it should not.

Neustradamus:
@paulusmack: Have you done tests, better?

PR #171, opened 2020-10-01
Enable IPv6 by default

pali:
Now in year 2020 we should not have disabled IPv6 support by default.

So let both IPv4 and IPv6 enabled by default.


paulusmack:
Thanks Pali.

Neustradamus:
@pali: Thanks :)

PR #174, opened 2020-11-09
Fedora patches

yarda:
Downstream patches we currently use in Fedora. They are slightly reworked to apply to master. It's related to #111. 

We also apply https://www.nikhef.nl/~janjust/ppp/ppp-2.4.8-eaptls-mppe-1.300.patch. I don't know why it's not upstreamed. It works quite OK in Fedora for a while.

### DIFF
--- a/chat/Makefile.linux
+++ b/chat/Makefile.linux
@@ -13,7 +13,7 @@ CDEFS=	$(CDEF1) $(CDEF2) $(CDEF3) $(CDEF4)
 COPTS=	-O2 -g -pipe
 CFLAGS=	$(COPTS) $(CDEFS)
 
-INSTALL= install
+INSTALL= install -p
 
 all:	chat
 
--- a/chat/Makefile.linux
+++ b/chat/Makefile.linux
@@ -13,6 +13,7 @@ CDEFS=	$(CDEF1) $(CDEF2) $(CDEF3) $(CDEF4)
 COPTS=	-O2 -g -pipe
 CFLAGS=	$(COPTS) $(CDEFS)
 
+STRIP_FLAG=-s
 INSTALL= install -p
 
 all:	chat
@@ -25,7 +26,7 @@ chat.o:	chat.c
 
 install: chat
 	mkdir -p $(BINDIR) $(MANDIR)
-	$(INSTALL) -s -c chat $(BINDIR)
+	$(INSTALL) $(STRIP_FLAG) -c chat $(BINDIR)
 	$(INSTALL) -c -m 644 chat.8 $(MANDIR)
 
 clean:
--- a/chat/chat.8
+++ b/chat/chat.8
@@ -200,7 +200,7 @@ The \fBSAY\fR directive allows the script to send strings to the user
 at the terminal via standard error.  If \fBchat\fR is being run by
 pppd, and pppd is running as a daemon (detached from its controlling
 terminal), standard error will normally be redirected to the file
-/etc/ppp/connect\-errors.
+/var/log/ppp/connect\-errors.
 .LP
 \fBSAY\fR strings must be enclosed in single or double quotes. If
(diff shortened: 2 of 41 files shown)


Neustradamus:
@yarda: Thanks for your PR but it is not better to have one by patch?

Do not forget: https://github.com/paulusmack/ppp/blob/master/Submitting-patches.md

You must do "Signoff"...

yarda:
NP, but it will take some time.

paulusmack:
Most of the commits need a signoff and a better description.

I don't want to add a hard dependency on udev, because some systems (e.g. embedded systems built using buildroot) do not have it.

paulusmack:
I could cherry-pick the commits that are simple and obvious and have a sign-off...

Neustradamus:
@yarda: @sthibaul has done some PRs linked to Debian (and used by a lot of others), can you look what is used by Fedora?
- 183 -> 200: https://github.com/paulusmack/ppp/pulls

It will help you to clean this mixed PR.

Thanks in advance.

Neustradamus:
@yarda: 2.4.9 is out :)
Please look what must be added into Fedora ^^

yarda:
> @yarda: 2.4.9 is out :)
> Please look what must be added into Fedora ^^

I built 2.4.9 for Fedora Rawhide.

I will PR the remaining patches one by one, so feel free to close this PR.

There is currently one known regression - we had downstream patch which was obtaining the first interface name from the udev, because in 2.4.8 there was hardcoded "eth0". I dropped this patch and used the upstream code, but it seems that the pppoe-discovery still hardcodes "eth0", so this should be also fixed, at the moment I have no patch for it (the udev approach is probably overkill).

yarda:
> There is currently one known regression - we had downstream patch which was obtaining the first interface name from the udev, because in 2.4.8 there was hardcoded "eth0". I dropped this patch and used the upstream code, but it seems that the pppoe-discovery still hardcodes "eth0", so this should be also fixed, at the moment I have no patch for it (the udev approach is probably overkill).

I opened #219.

pali:
And it is the issue? `pppoe-discovery` is just debugging tool. Anybody who is debugging pppoe can supply argument...

yarda:
> And it is the issue? `pppoe-discovery` is just debugging tool. Anybody who is debugging pppoe can supply argument...

Probably not, just nice to have.

I opened PRs for all other patches we had in Fedora. Most of them were added by previous Fedora maintainer. Choose what you think makes sense for upstream.

pali:
I'm not the maintainer, but maybe would not it better to remove default eth0 and requires user to always specify interface name? Now when udev automatically renames eth0 to eth_some_hash it would have small impact on existing users as they already need to specify interface name. And choosing arbitrary (= first one from that list of hashes) does not sound like a good idea.

pali:
@yarda: Hint about enumerating network interfaces: You do not have to use udev at all. There is standard C/POSIX function `if_nameindex()` which returns all available interfaces. And then via e.g. Linux specific ioctl `SIOCGIFHWADDR` you can check if interface is ethernet or not.

yarda:
> @yarda: Hint about enumerating network interfaces: You do not have to use udev at all. There is standard C/POSIX function `if_nameindex()` which returns all available interfaces. And then via e.g. Linux specific ioctl `SIOCGIFHWADDR` you can check if interface is ethernet or not.

Thanks for the tip. It needs somebody to write the patch, it shouldn't be hard. I was just providing what we had in Fedora.

pali:
I have already done it. Moreover it is merged and also avilable in the newly released version.

yarda:
> I have already done it. Moreover it is merged and also avilable in the newly released version.

Thanks. I am closing this PR, because it's obsoleted now. I opened individual PRs for all the remaining patches and #219 for the network interface discovery. Feel free to also close it.

PR #177, opened 2020-11-24
pppd: Add support for EAP-MSCHAPv2 in server mode

tisj:
This implements EAP-MSCHAPv2 as specified by
draft-kamath-pppext-eap-mschapv2-02 [1] in server mode (as the client
mode is already proposed by Thomas [2]).

The flow is camparable to MD5-Challenge, with some additional steps for
confirmation:

* Client sends an EAP request
* Server answers with MD5-Challenge
* Client Naks and requests MSCHAPv2
* Server answers with MSCHAPv2-Challenge
* Client answers with MSCHAPv2-Response
* Server answers with MSCHAPv2-{Success/Failure}
* Client possibly confirms MSCHAPv2-{Success/Failure}

I reused as much as possible from pppd/chap-new.c and pppd/chap_ms.c,
but most of the implementation is protected by static functions.

Therefore eap_chapms2_verify_response is an exact copy of
chapms2_verify_response, likewise for eap_chap_verify which is an exact
copy of chap_verify_response. This is not optimal and subject for
improvement.

By using a chap_digest_type struct/object with validator
(verify_response), validation is compatible with other (external)
plugins that have a specific CHAP implementation exposed in the
chap_verify_hook (like radius, windbind, ...).

Changes in eap.h are identical to [2], except for the additional
eapMSCHAPv2Chall server state.

[1] https://tools.ietf.org/html/draft-kamath-pppext-eap-mschapv2-02
[2] https://github.com/enaess/ppp-mschap-v2.git

Signed-off-by: Tijs Van Buggenhout <tvbuggen@gmail.com>

### DIFF
--- a/pppd/eap.c
+++ b/pppd/eap.c
@@ -84,6 +84,11 @@
 #ifdef USE_EAPTLS
 #include "eap-tls.h"
 #endif /* USE_EAPTLS */
+#ifdef CHAPMS
+#include "magic.h"
+#include "chap_ms.h"
+#include "chap-new.h"
+#endif /* CHAPMS */
 
 eap_state eap_states[NUM_PPP];		/* EAP state; one for each unit */
 #ifdef USE_SRP
@@ -706,6 +711,9 @@ int status;
 		}
 		break;
 
+#ifdef CHAPMS
+	case eapMSCHAPv2Chall:
+#endif
 	case eapMD5Chall:
 		if (status != 0) {
 			esp->es_server.ea_state = eapBadAuth;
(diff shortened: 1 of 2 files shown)


pppd/eap.c
```diff
@@ -2304,6 +2521,103 @@ int len;
...
+
+			if (vallen != MS_CHAP2_RESPONSE_LEN) {
+				error("EAP: Invalid MSCHAPv2-Response "
+						"length %d", vallen);
+				eap_figure_next_state(esp, 1);
+				break;
+			}
+
+			/* Not so likely to happen. */
+			if (vallen >= len + sizeof (rhostname)) {
```
paulusmack, line 2554: This is a repeat of the buffer overflow bug that resulted in a CVE, that was fixed in commit 8d7970b8f3db.

tisj: Thanks for the review @paulusmack 

I updated the commit with the following patch:

```diff
@@ -2541,7 +2527,7 @@
                        GETCHAR(vallen, inp);
                        len -= 4;

-                       if (vallen != MS_CHAP2_RESPONSE_LEN) {
+                       if (vallen != MS_CHAP2_RESPONSE_LEN || vallen > len) {
                                error("EAP: Invalid MSCHAPv2-Response "
                                                "length %d", vallen);
                                eap_figure_next_state(esp, 1);
@@ -2549,7 +2535,7 @@
                        }

                        /* Not so likely to happen. */
-                       if (vallen >= len + sizeof (rhostname)) {
+                       if (len - vallen >= sizeof (rhostname)) {
                                dbglog("EAP: trimming really long peer name down");
                                BCOPY(inp + vallen, rhostname, sizeof (rhostname) - 1);
                                rhostname[sizeof (rhostname) - 1] = '\0';
```

I also converted function definitions to ANSI C style.

enaess:
Just to clarify, I was able to re-use the code in chap-new.h by exposing another function to obtain the chap_digest_type structure for MSCHAPv2 and use that. In doing so, we reduce some of the clutter in eap.c which really ought to have a hammer taken to it...

tisj:
> Just to clarify, I was able to re-use the code in chap-new.h by exposing another function to obtain the chap_digest_type structure for MSCHAPv2 and use that. In doing so, we reduce some of the clutter in eap.c which really ought to have a hammer taken to it...

@enaess Indeed it would avoid code duplication in eap.c (from chap_ms and chap-new). However the find function breaks "protection" of the chap_digest_types. I'm not sure whether this is acceptable or if some other mechanism could be invented to avoid this.

enaess:
> > Just to clarify, I was able to re-use the code in chap-new.h by exposing another function to obtain the chap_digest_type structure for MSCHAPv2 and use that. In doing so, we reduce some of the clutter in eap.c which really ought to have a hammer taken to it...
> 
> @enaess Indeed it would avoid code duplication in eap.c (from chap_ms and chap-new). However the find function breaks "protection" of the chap_digest_types. I'm not sure whether this is acceptable or if some other mechanism could be invented to avoid this.

What do you mean by "break protection"? 

Paul merged my change, and I think the author of the original code didn't expose the find function because it would be unused. 

I didn't mean for my previous comment as a criticism of your change. Rather now as this function is exposed, could the digest be used over providing a new implementation? 

The callbacks in the chap-new.c file decouples the packet packing, e.g. eap over chap, but provide the internal math to generate challenges, verify and validate the success and failure responses. 

One thing I noticed is that the callback functions processes both the value length and value attribute of the request/response. I had to at least account for this when using these functions. Now, I haven't looked too much on the server side as I don't have a way of testing this, but do you think this could be done? It could be a bit cleaner ...

enaess:
@tisj -- Could you have a whirl testing the following change?

https://github.com/paulusmack/ppp/commit/16f2541c0a92d64ac7716f3f09341850d8696b6d

tisj:
> 
> 
> @tisj -- Could you have a whirl testing the following change?
> 
> [16f2541](https://github.com/paulusmack/ppp/commit/16f2541c0a92d64ac7716f3f09341850d8696b6d)

@enaess : See [tisj:eap-mschapv2-refactor:e981ea0517](https://github.com/tisj/ppp/commit/e981ea0517f67f1129b38a68e94b09e17c4a5820)

Going to test this further tomorrow in real life (windows 10 client, etc).

enaess:
Works for me

enaess:
@tisj any update?

tisj:
> Works for me

I second that ;-)

> @tisj any update?

Creating pull request now.

PR #178, opened 2020-11-26
Close discovery socket after session completed

xqjcool1:
After the session is complete, the socket is left unmanaged. When the interface receives PADIs from other device, the packets is putting in the socket's Recv-Q, which eat system memory.

[root@test ~]# ss -f link
Netid  Recv-Q Send-Q    Local Address:Port    Peer Address:Port                
p_raw  10269952 0          ppp_disc:eth1           *


Neustradamus:
Replaced by https://github.com/paulusmack/ppp/pull/179.

PR #188, opened 2020-12-30
Add option to strip MS domain name

sthibaul:
Some Windows 9x/ME clients might be erroneously transmitting the MS domain
along the login name. This allows to strip them on the server side.


paulusmack:
This really should have added the new option to pppd.8, but I merged it anyway.

PR #189, opened 2020-12-30
support building pppdump with the system zlib

sthibaul:


paulusmack:
@sthibaul I reverted this one since it caused compile failures (on Fedora at least), which reminded me that the zlib.c here is not the same as upstream; it has extra functions that I added a long time ago.

PR #193, opened 2020-12-30
Fix buffer overflow in rc_mksid()

sthibaul:
rc_mksid converts the PID of pppd to hex to generate a pseudo-unique string.

If the process id is bigger than 65535 (FFFF), its hex representation will be
longer than 4 characters, resulting in a buffer overflow.

The bug can be exploited to cause a remote DoS.


sthibaul:
Ah, the buffer was extended already

Neustradamus:
@sthibaul: Thanks!
Have you found the linked commit?

sthibaul:
That was 858976b1fc3107f1261aae337831959b511b83c2

Neustradamus:
@sthibaul: Thanks for your precision :)

PR #199, opened 2020-12-30
Handle IPv6 RADIUS attributes

sthibaul:


sthibaul:
Ah, sorry, that's https://github.com/paulusmack/ppp/pull/176

PR #200, opened 2020-12-30
Add replacedefaultroute option

sthibaul:
Adds an option to pppd to control whether to replace existing default routes
when using the 'defaultroute' option.

If defaultroute and replacedefaultroute are both set, pppd replaces an existing
default route with the new default route. The old default route is restored when
the connection is taken down.

### DIFF
--- a/pppd/ipcp.c
+++ b/pppd/ipcp.c
@@ -196,6 +196,15 @@ static option_t ipcp_option_list[] = {
       "disable defaultroute option", OPT_ALIAS | OPT_A2CLR,
       &ipcp_wantoptions[0].default_route },
 
+#ifdef __linux__
+    { "replacedefaultroute", o_bool,
+				&ipcp_wantoptions[0].replace_default_route,
+      "Replace default route", OPT_PRIV | 1
+    },
+    { "noreplacedefaultroute", o_bool,
+				&ipcp_wantoptions[0].replace_default_route,
+      "Do not replace default route", 0 },
+#endif
     { "proxyarp", o_bool, &ipcp_wantoptions[0].proxy_arp,
       "Add proxy ARP entry", OPT_ENABLE|1, &ipcp_allowoptions[0].proxy_arp },
     { "noproxyarp", o_bool, &ipcp_allowoptions[0].proxy_arp,
@@ -269,7 +278,7 @@ struct protent ipcp_protent = {
     ip_active_pkt
 };
 
-static void ipcp_clear_addrs (int, u_int32_t, u_int32_t);
+static void ipcp_clear_addrs (int, u_int32_t, u_int32_t, bool);
 static void ipcp_script (char *, int);	/* Run an up/down script */
 static void ipcp_script_done (void *);
 
@@ -1716,7 +1725,8 @@ ip_demand_conf(int u)
     if (!sifnpmode(u, PPP_IP, NPMODE_QUEUE))
 	return 0;
     if (wo->default_route)
-	if (sifdefaultroute(u, wo->ouraddr, wo->hisaddr))
+	if (sifdefaultroute(u, wo->ouraddr, wo->hisaddr,
+					    wo->replace_default_route))
 	    default_route_set[u] = 1;
     if (wo->proxy_arp)
 	if (sifproxyarp(u, wo->hisaddr))
@@ -1804,7 +1814,8 @@ ipcp_up(fsm *f)
      */
     if (demand) {
 	if (go->ouraddr != wo->ouraddr || ho->hisaddr != wo->hisaddr) {
-	    ipcp_clear_addrs(f->unit, wo->ouraddr, wo->hisaddr);
+	    ipcp_clear_addrs(f->unit, wo->ouraddr, wo->hisaddr,
+				      wo->replace_default_route);
 	    if (go->ouraddr != wo->ouraddr) {
 		warn("Local IP address changed to %I", go->ouraddr);
 		script_setenv("OLDIPLOCAL", ip_ntoa(wo->ouraddr), 0);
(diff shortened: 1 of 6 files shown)


pppd/ipcp.c
```diff
@@ -196,6 +196,16 @@ static option_t ipcp_option_list[] = {
       "disable defaultroute option", OPT_ALIAS | OPT_A2CLR,
       &ipcp_wantoptions[0].default_route },
 
+#ifdef __linux__
+    { "replacedefaultroute", o_bool,
+				&ipcp_wantoptions[0].replace_default_route,
+      "Replace default route", OPT_PRIV | 1
+    },
+    { "noreplacedefaultroute", o_bool,
+				&ipcp_allowoptions[0].replace_default_route,
```
paulusmack, line 205: This is set up so that the options subsystem will clear ipcp_allowoptions[0].replace_default_route if the noreplacedefaultroute option is given; however that won't have any effect. The similar code for the defaultroute option uses OPT_ENABLE to get nodefaultroute to veto defaultroute, but you don't have OPT_ENABLE on the replacedefaultroute option. In any case, replacedefaultroute being privileged means you probably don't want it to be vetoed by a previous noreplacedefaultroute. The change to pppd.8 below says that it would be, though in fact it won't, you probably don't want it to be, and clearing ipcp_allowoptions here is pointless.

pppd/ipcp.c
```diff
@@ -1716,7 +1726,12 @@ ip_demand_conf(int u)
     if (!sifnpmode(u, PPP_IP, NPMODE_QUEUE))
 	return 0;
     if (wo->default_route)
+#ifndef __linux__
```
paulusmack, line 1729: This is ugly, having an #ifndef inside the function. Why not just add the extra parameter to the code in sys-solaris.c as well as sys-linux.c, and make the solaris version return an error if it is set?

pppd/pppd.8
```diff
@@ -792,7 +797,12 @@ disable both forms of hardware flow control.
 .TP
 .B nodefaultroute
 Disable the \fIdefaultroute\fR option.  The system administrator who
-wishes to prevent users from creating default routes with pppd
+wishes to prevent users from adding a default route with pppd
+can do so by placing this option in the /etc/ppp/options file.
+.TP
+.B noreplacedefaultroute
+Disable the \fIreplacedefaultroute\fR option. The system administrator who
+wishes to prevent users from replacing a default route with pppd
```
paulusmack, line 805: Users can't set replacedefaultroute anyway since it's privileged, so this sentence is misleading.

pppd/pppd.8
```diff
@@ -133,6 +133,11 @@ the gateway, when IPv6CP negotiation is successfully completed.
 This entry is removed when the PPP connection is broken.  This option
 is privileged if the \fInodefaultroute6\fR option has been specified.
 .TP
+.B replacedefaultroute
+This option is a flag to the defaultroute option. If defaultroute is
+set and this flag is also set, pppd replaces an existing default route
+with the new default route.
+.TP
```
paulusmack, line 140: Should say that this option is privileged.

pppd/ipcp.c
```diff
@@ -1829,7 +1845,12 @@ ipcp_up(fsm *f)
 
 	    /* assign a default route through the interface if required */
 	    if (ipcp_wantoptions[f->unit].default_route) 
+#ifndef __linux__
```
paulusmack, line 1848: Similarly this ifndef and the ones below could go away...

pppd/sys-linux.c
```diff
@@ -1638,22 +1643,53 @@ int have_route_to(u_int32_t addr)
...
-	    error("not replacing existing default route through %s with metric %d",
-		  rt.rt_dev, dfl_route_metric);
-	return 0;
+int sifdefaultroute (int unit, u_int32_t ouraddr, u_int32_t gateway, bool replace)
+{
+    struct rtentry rt, tmp_rt;
+    struct rtentry *del_rt = NULL;
+
+    if (default_rt_repl_rest) {
+       /* We have already reclaced the original defaultroute, if we
```
paulusmack, line 1662: s/reclaced/replaced/

paulusmack (changes requested):

pppd/ipcp.c
sthibaul: > replacedefaultroute being privileged means you probably don't want it to be vetoed by a previous noreplacedefaultroute

Mmm, I don't see why?
While I understand why replacedefaultroute should be privileged to avoid e.g. disrupting other work on the system, a user may want to call some ppp network, but avoid affecting the default route even if the usual case is to replace the default route.

pppd/sys-linux.c
sthibaul: fixed so.

pppd/ipcp.c
sthibaul: yep, done so.

pppd/pppd.8
sthibaul: Right, fixed so.

pppd/ipcp.c
sthibaul: Indeed, fixed so.

pppd/ipcp.c
paulusmack: > > replacedefaultroute being privileged means you probably don't want it to be vetoed by a previous noreplacedefaultroute
> 
> Mmm, I don't see why?
> While I understand why replacedefaultroute should be privileged to avoid e.g. disrupting other work on the system, a user may want to call some ppp network, but avoid affecting the default route even if the usual case is to replace the default route.

What I meant was that you don't want replacedefaultroute to be ineffective because of an earlier noreplacedefaultroute. That is, replacedefaultroute should always set the replace_default_route flag. This is unlike the defaultroute option, which does not work if there has been an earlier nodefaultroute option seen.

sthibaul: Ah ok you mean the ordering. I have now fixed it so.

pppd/pppd.8
sthibaul: I have reworked this as well.

PR #202, opened 2020-12-31
pppd/Makefile.linux: guard libutil usage with musl non-define

kanavin:
libutil is only available in glibc, and otherwise needs to be
disabled.

Signed-off-by: Alexander Kanavin <alex.kanavin@gmail.com>

### DIFF
--- a/pppd/Makefile.linux
+++ b/pppd/Makefile.linux
@@ -79,7 +79,9 @@ PLUGIN=y
 #USE_SRP=y
 
 # Use libutil
+ifndef USE_MUSL
 USE_LIBUTIL=y
+endif
 
 # Enable EAP-TLS authentication (requires MPPE support, libssl and libcrypto)
 USE_EAPTLS=y


Neustradamus:
@kanavin: Thanks for your PR but why have you closed?

kanavin:
@Neustradamus after I submitted this, I found that an extra variable is unnecessary: 'make USE_LIBUTIL=' works just as well. But maybe it's better to reopen this so there is an explicit setting for musl users that can be used elsewhere if needed?

Neustradamus:
@paulusmack will give this point of view ^^

paulusmack:
So on the one hand we can do "make USE_LIBUTIL=" vs. on the other hand we could do "make USE_MUSL=y". It doesn't seem compelling.

We already have a compilation test in Makefile.linux for crypt.h/libcrypt. We could do a similar test to see if logwtmp is declared in <utmp.h>.

paulusmack:
This is fixed by commit 248c11d375c5.

PR #203, opened 2020-12-31
Document chapms-strip-domain option on pppd(8)

sthibaul:


pppd/pppd.8
```diff
@@ -298,6 +298,9 @@ to \fIn\fR seconds (default 3).
 .B chap-timeout \fIn
 Set timeout for CHAP authentication by peer to \fIn\fR seconds (default 60).
 .TP
+.B chapms-strip-domain
```
pali, line 301:
```suggestion
.B chapms\-strip\-domain
```
dash character should be escaped in the TROFF language

sthibaul: right, fixed

pppd/pppd.8
```diff
@@ -298,6 +298,9 @@ to \fIn\fR seconds (default 3).
 .B chap-timeout \fIn
 Set timeout for CHAP authentication by peer to \fIn\fR seconds (default 60).
 .TP
+.B chapms\-strip\-domain
+Strip the domain prefix before the Username.
```
paulusmack, line 302: I think this needs to be a bit more verbose, since it's unclear which username you are talking about or when this option would be effective.  The way it is written currently sounds like this could affect the username set with the 'user' option and which this pppd sends to the peer. Instead, as I understand it, this is for when pppd is authenticating the peer (i.e. being a server) using MS-CHAP and the peer is providing the username with the domain prepended, and we need to strip off that domain prefix in order to match a username stored in the secrets file.

sthibaul: Ok, extended so

PR #205, opened 2020-12-31
Rename rp-pppoe.so plugin to pppoe.so

pali:
Original out-of-tree rp-pppoe plugin for pppd software is still available
at website https://dianne.skoll.ca/projects/rp-pppoe/ and receives new
releases. The last update is from the June 2020.

Currently it is ambiguous if user is using original out-of-tree rp-pppoe
plugin or in-tree pppd's rp-pppoe plugin. These two plugins are different,
come from different sources but share same name.

Some users want to use original rp-pppoe plugin and not pppd's in-tree
rp-pppoe plugin. Also some distribution want to package both plugins,
pppd's in-tree and original rp-pppoe.

So for this reason and also because all other PPP over <something> plugins
have just pppo prefix, rename in-tree rp-pppoe.so plugin to just pppoe.so.

This will allow to distinguish and make it clear what is the original
rp-pppoe plugin and what is pppd's in-tree pppoe plugin.

When installing pppd create a compatibility symlink from pppoe.so to
rp-pppoe.so so nothing would be broken. This compatibility symlink may be
removed by Linux distribution which do not want to have compatibility with
the old name as before and rather use rp-pppoe.so name for original
rp-pppoe software.


paulusmack:
This mostly looks great; I just have a couple of comments:

- In the first commit, why do you make a new fatal() rather than just calling the existing one from utils.c? In general all the global symbols from pppd are available to plugins.

- In the second commit, please put OPT_ALIAS in the flags for all of the legacy compatibility options. At the moment it doesn't do anything, but it could be useful information for print_options() for example.

pali:
> - In the first commit, why do you make a new fatal() rather than just calling the existing one from utils.c? In general all the global symbols from pppd are available to plugins.

This change is only in pppoe-discovery.c file which is compiled only to pppoe-discovery binary. It is not used by pppd binary nor by any plugin. pppoe-discovery is independent binary, it is not part of pppd and therefore does not use pppd's utils.c file.

pali:
> - In the second commit, please put OPT_ALIAS in the flags for all of the legacy compatibility options. At the moment it doesn't do anything, but it could be useful information for print_options() for example.

Done!

paulusmack:
> > * In the first commit, why do you make a new fatal() rather than just calling the existing one from utils.c? In general all the global symbols from pppd are available to plugins.
> 
> This change is only in pppoe-discovery.c file which is compiled only to pppoe-discovery binary. It is not used by pppd binary nor by any plugin. pppoe-discovery is independent binary, it is not part of pppd and therefore does not use pppd's utils.c file.

Ah yes, of course.

PR #209, opened 2021-01-01
pppoe-discovery: Use discovery code from pppoe plugin

pali:
This pull request in firt commits cleanup pppoe plugin code so it can be
used by pppoe-discovery binary and in final two commits is internal copy
of discovery code from pppoe-discovery.c completely removed. So with this
pull request pppoe-discovery binary is build from same source code as
pppd's pppoe.so plugin.


paulusmack:
On the whole, I would rather leave this PR until after the release. I cherry-picked the first commit, which is a bug-fix.

pali:
Ok!

Neustradamus:
@pali: I understand that it will be not merged before the new version, but please rebase :)

pali:
done, rebased

pali:
@paulusmack: New version was released, anything else is needed for this pull request?

Neustradamus:
@paulusmack: Can you look the @pali PR (you have said after 2.4.9 release)?

pali:
@paulusmack: Can you look at this pull request now?

pali:
@paulusmack: ping; Anything else is needed for this pull request now?

paulusmack:
There are two reasons why I haven't pulled this yet: first, it's a large change to review, and secondly, it makes us diverge further from the "upstream" rp-pppoe code. Maybe the second reason doesn't matter any more, but it is a hurdle I would have to get over.

pali:
Ok! If there is something which needs to be improved or changed, please let me know. rp-pppoe and this pppd's pppoe plugins are already too different.

Neustradamus:
@paulusmack: Have you seen the last @pali comment?

"rp-pppoe and this pppd's pppoe plugins are already too different."

Neustradamus:
@paulusmack: I think that it is possible to merge this PR because rp-pppoe has already changed.

Neustradamus:
Recall : @paulusmack: I think that it is possible to merge this PR because rp-pppoe has already changed.

After it, @pali will work on the https://github.com/ppp-project/ppp/issues/219 ticket.

paulusmack:
I grabbed rp-pppoe-3.15.tar.gz and compared it with what we have. It looks like there are a couple of minor things that we could copy, but that doesn't need to impede this patchset.

The one thing in this patchset that makes me nervous is the change from %.*v to %.*s in pppoe_printpkt(), because that means that a buggy, hacked or malicious AC has a way to get pppd to send control characters or non-ASCII characters to syslog, and there might be vulnerabilities in syslog or log-parsing software which could then be exploited. I understand the desire not to use %v in the stand-alone pppoe-discovery program, but I want it used in the pppd plugin.

pali:
The issue with which I was dealing was that `pppoe-discovery` program behaved differently than pppd's `pppoe` plugin. And it was because these they have slightly different code (something was fixed in plugin but not in discovery program). So for debugging purposes it is important to have same code.

Anyway, if want from me to improve this patch set, I can do it, just I need to know what. For sure I can fix `%.*s` to not use it at all.

So, please let me know.

Neustradamus:
@paulusmack: Have you seen the last pali answer?

paulusmack:
I merged this and added another commit to force text with non-printing characters in it to be printed as hex, as the easiest and simplest way to stop it from sending non-printing characters to syslog. If people complain about the hex output then we can do something more elaborate.

Thanks pali.

pali:
Ok!

Anyway, as I said, if any more improvements is needed, let me know and I will try to do it!

Neustradamus:
Thanks @paulusmack and @pali!

Issue #138, opened 2020-02-24
Add support for EAP-MSCHAPv2


enaess:
There is an implementation available here. I also created a branch for this fix here: https://github.com/enaess/ppp-eap-mschapv2

https://w3logistics.com/blog/archives/438-EAP-MSCHAPv2-for-pppd-2-4-7.html

Neustradamus:
@paulusmack: What do you think?

Can you look for it?

Thanks in advance.

Neustradamus:
The PR is here: https://github.com/paulusmack/ppp/pull/139

enaess:
Okay, let me do this again against the latest tip of master branch.

Neustradamus:
@enaess: Thanks, a new PR is here:
- https://github.com/paulusmack/ppp/pull/175

@paulusmack: Can you look?

Neustradamus:
@enaess, @tisj: Please look @paulusmack comments :)
- https://github.com/paulusmack/ppp/pull/175
- https://github.com/paulusmack/ppp/pull/177

I really hope that your good job will be merged soon to have a new perfect version before Debian 11 freeze, after it, the next will be in 2023...

enaess:
@Neustradamus @paulusmack 

Added new pull request that contains all the changes to add EAP-MSCHAPv2. 

Please have a look at #211

PR #211, opened 2021-01-01
Add support for EAP-MSCHAPv2 #138

enaess:
This is the client side support for EAP-MSCHAPv2

Updated the original patch to also perform mutual authentication and use the chap-new API for CHAP/CHAPv2 digest algorithms as this properly cache the previous response in case of re-transmits. 

Also did length check on input per @paulusmack request in earlier review.


paulusmack:
I'm still not happy with the length checking, sorry. We now check that the size from the header (mssize) of a received packet matches the actual length, and that the size byte of the challenge field is MS_CHAP2_PEER_CHAL_LEN, but for a CHAP_CHALLENGE packet I don't see any check that len >= MS_CHAP2_PEER_CHAL_LEN. Consequently we could get len going negative and being passed to BCOPY(inp, rhostname, len), leading to a potential DoS or worse.

enaess:
@paulusmack I agree, that was sloppy. Added more length checks to address the issues you called out. Do you mind having another look? 

Ps. I ended up merging in changes from paulusmack/master -> enaess/ppp-master, then bringing those into the branch I am working. Now, that is called out as a separate commit, and I hope that isn't going to be a problem with the workflow on github...

paulusmack:
The DCO check is whinging because you put in a comma and didn't put angle brackets around the email address. Don't worry, I think I will do a squash-and-merge once the checks have run, and sort out the signoff manually.

PR #212, opened 2021-01-01
Fix setting IPv6 peer address

pali:
On Linux IPv6 peer address cannot be set via SIOCSIFDSTADDR ioctl like it
is for IPv4 peer address. Linux kernel does not support SIOCSIFDSTADDR for
AF_INET6 PPP interfaces.

The only way how to set IPv6 peer address on Linux is via kernel netlink
interface which is just a little bit complicated compared to one ioctl
call.

Linux kernel for a long time automatically adds IPv6 peer address from
interface into route table so it is not needed to explicitly set routing
for remote peer address. pppd already does not do it for kernel versions
newer than 2.1.16. So the same check is used also for IPv6 peer route
address.

Prior this patch ppp interface was configured as:

    $ ip -6 address show dev ppp0
    2: ppp0: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1500 state UNKNOWN qlen 3
        inet6 fe80::2/128 scope link
           valid_lft forever preferred_lft forever

    $ ip -6 route show dev ppp0
    fe80::1 metric 1 pref medium
    fe80::2 proto kernel metric 256 pref medium

And after applying this patch as:

    $ ip -6 address show dev ppp0
    2: ppp0: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1500 state UNKNOWN qlen 3
        inet6 fe80::2 peer fe80::1/128 scope link
           valid_lft forever preferred_lft forever

    $ ip -6 route show dev ppp0
    fe80::1 proto kernel metric 256 pref medium
    fe80::2 proto kernel metric 256 pref medium

As can be seen IPv6 peer address is now correctly set on the interface and
also kernel correctly fill route table for IPv6 peer address.

Please note that old ifconfig utility cannot show nor change IPv6 peer
address. Peer address is supported only for IPv4 addresses as opposite of
the local addresses where both IPv4 and IPv6 are supported. It is because
old ifconfig utility is also using ioctl interface which cannot handle it.
Therefore for any testing it is really required ip utility or other utility
with netlink interface (and not ioctl interface).

### DIFF
--- a/pppd/sys-linux.c
+++ b/pppd/sys-linux.c
@@ -121,6 +121,12 @@
 #include <linux/ppp_defs.h>
 #include <linux/if_ppp.h>
 
+#ifdef INET6
+#include <linux/netlink.h>
+#include <linux/rtnetlink.h>
+#include <linux/if_addr.h>
+#endif
+
 #include "pppd.h"
 #include "fsm.h"
 #include "ipcp.h"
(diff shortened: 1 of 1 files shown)


joakim-tjernlund:
I recall asking Linux networking about the missing SIOCSIFDSTADDR for IPv6 and think I got the answer that it
is missing because noone has added it, yet. I remember looking at adding it but quickly discovered that it was non trivial, at least for me.

pali:
Really? I thought that all these ioctls are deprecated and supported only for compatibility with 1.x/2.0 kernels and new (kernel) code should use only modern netlink interface which is supported since 2.2 kernel. I have looked at kernel code when I was writing this patch and SIOCSIFDSTADDR is supported in IPv6 world only for SIT (IPv6-in-IPv4) point-to-point interfaces. So my guess was that adding support for other point-to-point (PPP) interfaces could be easy. But I have not tried it.

pali:
Now I see the reason why it is not supported. Peer address is bounded with local address and you have to specify to which local address you want to set peer address. You cannot add to interface only peer addresses. Interface can have: no address, local address only or local address with peer address.

So via ioctl either supply both local and peer and then as a pair to interface, or add a peer address to existing local address without peer. This is general for peer addresses.

In IPv4 this issue of specifying local address for peer does not exist as all IPv4 ioctls support only one address per interface, more precisly they operate only with the primary IPv4 addresses. All these ioctls do not see secondary IPv4 addresses.

On the other hand in IPv6 they operate over all addresses, so existing SIOCSIFDSTADDR cannot be implemented for IPv6 with current API at all.

PR #214, opened 2021-01-02
Enabled Buildroot CI test builds on pull request.

RICCIARDI-Adrien:


enaess:
For anyone wondering how to run an instance of the CI test builds with buildroot, the way to do this is:
```bash
git clone --depth=1 --branch=2020.11.1 https://git.busybox.net/buildroot

cd buildroot/package/pppd
# Allow package to build when musl libc is selected
sed -i '/depends on !BR2_TOOLCHAIN_USES_MUSL/d' Config.in
# Upstream version always needs OpenSSL
sed -i '/select BR2_PACKAGE_OPENSSL/c\\select BR2_PACKAGE_OPENSSL' Config.in
# Do not check for package hash, so there is no need to compute it
rm pppd.hash
# Buildroot patch is already applied upstream
rm -f 0001-pppd-Fix-bounds-check.patch
# Get package sources from head of master branch
sed -i "/PPPD_VERSION =/c\\PPPD_VERSION = bfffd94ad7455fb8cdf876b4b696c06756e9d1ee" pppd.mk
sed -i '/PPPD_SITE =/c\\PPPD_SITE = https://github.com/enaess/ppp' pppd.mk
sed -i '9iPPPD_SITE_METHOD = git' pppd.mk
# Plugin rp-pppoe has been renamed in "pppoe"
sed -i 's/rp-pppoe/pppoe/g' pppd.mk

cd ../..
echo "BR2_PACKAGE_PPPD=y" >> configs/qemu_x86_64_defconfig
echo "BR2_PACKAGE_PPPD_FILTER=y" >> configs/qemu_x86_64_defconfig
echo "BR2_PACKAGE_PPPD_RADIUS=y" >> configs/qemu_x86_64_defconfig
echo "BR2_PACKAGE_PPPD_OVERWRITE_RESOLV_CONF=y" >> configs/qemu_x86_64_defconfig

echo "BR2_TOOLCHAIN_BUILDROOT_MUSL=y" >> configs/qemu_x86_64_defconfig

make qemu_x86_64_defconfig
make
```

enaess:
@RICCIARDI-Adrien is there any way to make the buildroot CI work any faster? One permutation of the configuration takes 45 minutes. That's a bit excessive when the build on Ubuntu normally takes 2 minutes. The Solaris CI takes 15 minutes, but somehow it waits for a long time on an available runner / machine

RICCIARDI-Adrien:
Currently, the Buildroot configurations are real word builds and maintained upstream, this way it is really faster to maintain out Buildroot CI. To improve somehow the build speed, we would need to maintain as many configurations as there are architectures in the CI, and maybe test less.

enaess:
Yeah, I see that now....

But in reality, this is a CI of buildroot + pppd, no? If I understand this correctly, we end up building entire buildroot up until the point there is a embedded distribution with gcc, kernel, libc, et al. 

With the current setup for GitHub CI, I am not sure if there exist a better way ...

RICCIARDI-Adrien:
Correct. This is the simplest way I found to cross-compile PPP for multiple architectures and using multiple libc, but this is quite slow...
We might find a faster solution later (or maybe GitHub Actions will increase the amount of virtual cores per Linux container !).

PR #222, opened 2021-01-05
Do not accidentaly leak FDs

yarda:
It's usually better to use CLOEXEC.

### DIFF
--- a/pppd/auth.c
+++ b/pppd/auth.c
@@ -475,7 +475,7 @@ setupapfile(char **argv)
         free(fname);
 	return 0;
     }
-    ufile = fopen(fname, "r");
+    ufile = fopen(fname, "re");
     if (seteuid(euid) == -1)
 	fatal("unable to regain privileges: %m");
     if (ufile == NULL) {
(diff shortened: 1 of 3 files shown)


paulusmack:
Do all the libcs support the "e" modifier on the fopen mode argument? That seems to be a glibc-specific extension. I'm concerned that all these fopen calls might start failing for people using a different libc.

Neustradamus:
@yarda: Have you looked the @paulusmack comment?

Neustradamus:
@yarda: Have you looked the @paulusmack comment?

yarda:
It seems it's GNU extension and that it is supported since glibc 2.27, IMHO since year 2007. It's not problem for Fedora, because we have glibc. Could we add some autodetection? E.g. use 'e' if supported. It could use C macros. Does it make sense? If yes, I could change the PR.

Neustradamus:
@paulusmack: Have you seen the last @yarda comment?

Neustradamus:
@paulusmack: Have you seen the last @yarda comment?

enaess:
@yarda If people are worried about the Linux extension "e" for the fopen modifier, could the following be used (and also supported on SunOS): 
```
if (fcntl(fileno(file), F_SETFD, FD_CLOEXEC)) == -1)
     warn("Could not set close on execute flag on file, %s", filename);
```

Neustradamus:
@yarda: Have you looked the @enaess comment?

PR #223, opened 2021-01-05
Do not hardcode libdir

yarda:
Some distros use /usr/lib64 for 64 bit libs instead of /usr/lib
(e.g. Fedora), so do not hardcode it.


yarda:
Maybe there is better way than hardcoding gcc.

paulusmack:
You broke the solaris build...

paulusmack:
> Maybe there is better way than hardcoding gcc.

Right, should be $(CC).

yarda:
> You broke the solaris build...

What was the error on the solaris? I wasn't able to get the log from the CI.

I will change it to CC, but I think it works only with gcc. It shouldn't be problem because without gcc it should expand just to:
/usr/lib//....

yarda:
Could you provide some logs from the solaris? All what I was able to get from it is:
```
The process '/usr/bin/ssh' failed with exit code 1
```
This is quite unhelpful.

Neustradamus:
@yarda: At the bottom of this PR: "Some checks were not successful"
- https://github.com/paulusmack/ppp/pull/223/checks?check_run_id=1683399705

yarda:
@Neustradamus thanks, I can see the log now.

yarda:
I don't know why some tests were cancelled, could you relaunch them? At least it seems the problem with solaris is fixed now. Or I could do some dummy force push.

Neustradamus:
@yarda: Are you sure that your PR is really to "LIBDIR"?

cc @glensc:
- https://github.com/pld-linux/ppp/blob/master/ppp-lib64.patch

glensc:
@Neustradamus 

🚧 Never link to branches, share a permalink 🚧
- see https://news.ycombinator.com/item?id=8046710
  and https://help.github.com/articles/getting-permanent-links-to-files/
  and https://docs.gitlab.com/ee/user/shortcuts.html#project-files
- hit `y`
- and perhaps for https://github.com/josephfrazier/octopermalinker (GitHub)

NOTE: edit your post and fix it with a permalink!

Neustradamus:
@glensc: Thanks for your comment.

Can you look for this PR?

And create a PR for the patch that I have requested you?

Thanks in advance.

glensc:
@Neustradamus NOTE: edit your post and fix it with a permalink!

Neustradamus:
@glensc: I dislike but: https://github.com/pld-linux/ppp/blob/be4cb0a92e97a417e3a26a95e15032079c289836/ppp-lib64.patch

glensc:
@Neustradamus edit button is top right of each comment:

image

glensc (approved):

yarda:
> @yarda: Are you sure that your PR is really to "LIBDIR"?
> 
> cc @glensc:
> 
>     * https://github.com/pld-linux/ppp/blob/be4cb0a92e97a417e3a26a95e15032079c289836/ppp-lib64.patch

@Neustradamus I am not sure whether I understand your question. The problem is that on 64 bits some distros store arch specific 64 bit libraries under /usr/lib64 other under /usr/lib. AFAIK this is distro specific. The proposed patch tries to auto-detect it by using gcc. If there is no gcc it should fall-back to /usr/lib. Maybe there is more elegant way how to do it.

paulusmack:
This change relies on $(CC) having a -print-multi-os-directory option. PR #253 illustrates perfectly why I don't want to do that.

glensc:
Perhaps better (portable) is to define `LIB` variable that distributions can override:

```makefile
LIB = lib
LIBDIR = $(DESTDIR)/$(LIB)
```

```spec
make LIB=%{_lib} ... install
````


Oh, and `DESTDIR` should be renamed to `PREFIX` and `INSTROOT` replaced to be `DESTDIR`. Likely in two iterations, so downstream distributions can take time to adapt...

**Comment by paulusmack** [comment]
> Perhaps better (portable) is to define `LIB` variable that distributions can override:
> 
> ```makefile
> LIB = lib
> LIBDIR = $(DESTDIR)/$(LIB)
> ```
> 
> ```python
> make LIB=%{_lib} ... install
> ```

I like that suggestion better.

> Oh, and `DESTDIR` should be renamed to `PREFIX` and `INSTROOT` replaced to be `DESTDIR`. Likely in two iterations, so downstream distributions can take time to adapt...

That would be fine with me, assuming no downstream maintainers object.

**Comment by yarda** [comment]
> ```python
> make LIB=%{_lib} ... install
> ```

Me (as downstream maintainer) has no problem with that.

**Comment by Neustradamus** [comment]
@yarda: After previous comments, have you planned an update of this PR?

**Comment by yarda** [comment]
@Neustradamus I updated the PR.

**Comment by Neustradamus** [comment]
@yarda: Thanks! :)

@glensc, @paulusmack: What do you think?

**Comment by enaess** [comment]
Just reading this now, but would the fix I have for adding pkgconfig support also address this? It exposes a --libdir=DIR to ./configure. If you don't specify it, the same defaults are used as before the change.

See pull request #270

**Comment by yarda** [comment]
> Just reading this now, but would the fix I have for adding pkgconfig support also address this? It exposes a --libdir=DIR to ./configure. If you don't specify it, the same defaults are used as before the change.
> 
> See pull request #270

Yep, your PR seems to address it and for me it seems better than this hack. I could then use:
```
 ./configure --libdir=%{_libdir} 
```
So closing this.

PR #224, opened 2021-01-05
Use CLOEXEC

yarda:

### DIFF
--- a/pppd/tty.c
+++ b/pppd/tty.c
@@ -559,7 +559,7 @@ int connect_tty(void)
 				status = EXIT_OPEN_FAILED;
 				goto errret;
 			}
-			real_ttyfd = open(devnam, O_NONBLOCK | O_RDWR, 0);
+			real_ttyfd = open(devnam, O_NONBLOCK | O_RDWR | O_CLOEXEC, 0);
 			err = errno;
 			if (prio < OPRIO_ROOT && seteuid(0) == -1)
 				fatal("Unable to regain privileges");
@@ -713,7 +713,7 @@ int connect_tty(void)
 	if (connector == NULL && modem && devnam[0] != 0) {
 		int i;
 		for (;;) {
-			if ((i = open(devnam, O_RDWR)) >= 0)
+			if ((i = open(devnam, O_RDWR | O_CLOEXEC)) >= 0)
 				break;
 			if (errno != EINTR) {
 				error("Failed to reopen %s: %m", devnam);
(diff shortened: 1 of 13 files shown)


pppd/tty.c
```diff
@@ -559,7 +559,7 @@ int connect_tty(void)
 				status = EXIT_OPEN_FAILED;
 				goto errret;
 			}
-			real_ttyfd = open(devnam, O_NONBLOCK | O_RDWR, 0);
+			real_ttyfd = open(devnam, O_NONBLOCK | O_RDWR | O_CLOEXEC, 0);
```
paulusmack, line 562: This will prevent the tty file descriptor from being accessed by the connector script.

paulusmack:
I don't think we can just add O_CLOEXEC to every single open() call. They are going to have to be considered one by one.

Neustradamus:
@yarda: Have you looked the @paulusmack comments?

Neustradamus:
@yarda: Have you looked the @paulusmack comments?

yarda:
I am not author of the original patch, but for me it seems the original author was trying to resolve some FD leak problem and used "big hammer" approach. But there are definitely places where the O_CLOEXEC have sense. It's not too much places, could somebody go through them?

Neustradamus:
@paulusmack: Have you seen the last @yarda comment?

Neustradamus:
@paulusmack: Have you seen the last @yarda comment?

enaess:
@yarda - your change raises a few questions:

1.  What problem does this actually resolve, other than trying to address code with a "best-practice" approach? 
2.  See Paul's comments, there legit reasons to why a FD is open and should remain open after exec(); most notably is how pppd actually work. You can specify an option: pty "/usr/bin/sstpc --log-level 4 ... sstp-server.your-org.com". In this case the controlling terminal (tty or pty) is the STDIN (0) of the spawned program. 
3.  Other instances of open() or fopen(), the file is closed shortly afterwards, the risk of them being open while execv() is being invoked is pretty slim considered that pppd is pretty much single threaded.

pppd does a few operations where exec() is involved, including ip-up, ip-down, etc. The intention is to have hooks (or scripts) to run (with root privileges) to configure the system, e.g. add routes, send a message to systemd or network-manager, etc.

One approach is to enumerate the number of execv() calls in the code, and for each of these, run a executable that does the following: a) open /proc/self/fd and inspect the files. Typically the next available fd is 3 (as stdin: 0, stdout: 1, stderr: 2 are already in use). b) the script could just be the "ls" command with /proc/self/fd but you will have to redirect the output to a file less you have access to the console in that case.

Based on the said investigations, we should fix the individual cases where pppd "leaks" fds to any child process where it should not.

Neustradamus:
@yarda: Please read all comments :)

Neustradamus:
@paulusmack and @yarda: Please read all comments ^^

pali:
FDs 0, 1 and 2 should not be really marked as close-on-exec. Otherwise spawned scripts can be confused or not work if they do not have valid FDs 0, 1 and 2. It is not really sane nor best practice to mark 0, 1 and 2 as O_CLOEXEC.

PR #226, opened 2021-01-05
Use /var for variable data

yarda:


vincele:
Why not using `RUNDIR` instead of hardcoding `/var/run` in a lot of places ?
Some distros use `/run` instead of `/var/run`...

yarda:
> Why not using `RUNDIR` instead of hardcoding `/var/run` in a lot of places ?
> Some distros use `/run` instead of `/var/run`...

I tried to drop hardcodes where possible. The scripts will need rewrite to use preprocessor.

Neustradamus:
@vincele: What do you think?

paulusmack:
I can understand a distro wanting to change these paths, and I think the new locations are better.

However, I also don't want to break people's existing installations that might have dependencies on things being in /etc/ppp, for example, ip-up scripts that use /etc/ppp/resolv.conf. So I'm not sure what the best way to go here is.

vincele:
@Neustradamus this looks better

pppd/pathnames.h
```diff
@@ -33,9 +33,9 @@
 #define _PATH_AUTHUP	 _ROOT_PATH "/etc/ppp/auth-up"
 #define _PATH_AUTHDOWN	 _ROOT_PATH "/etc/ppp/auth-down"
 #define _PATH_TTYOPT	 _ROOT_PATH "/etc/ppp/options."
-#define _PATH_CONNERRS	 _ROOT_PATH "/etc/ppp/connect-errors"
+#define _PATH_CONNERRS	 _ROOT_PATH "/var/log/ppp/connect-errors"
```
Neustradamus, line 36: Why not to add:
`#define _PATH_VARLOG 	"/var/log/ppp/"`
after
`#define _PATH_VARRUN 	"/var/run/ppp/"`

Neustradamus:
@yarda: Please look my last comment :)

yarda:
@Neustradamus I updated the PR.

Neustradamus:
@yarda: Thanks! :)

@vincele, @paulusmack: What do you think?

vincele:
I glanced around and saw a lot more hardcoded pathes, so this PR is probably not the right place to tackle the hardcoding problem in general.

My initial remark was assuming pathes were variabilized everywhere, and this PR just missed that.
Sorry, I can't be of more help.

paulusmack:

I think I don't want to break backward compatibility, and this patch does that. By backward compatibility I mean that if someone has a working setup with ppp-2.4.9, say, and they install ppp-2.4.10, their setup should not stop working.

Neustradamus:
It will be possible to add:
- "alias" feature
- symlinks
- if exist, move files/folders...
- ...

enaess:
Could it be possible to specify this with an option to ./configure? Typically, autoconf would allow one to override the setting using
`
--localstatedir=DIR       modifiable single machine data [prefix/var]
`

yarda:
Or the override could be done in the make, e.g.:
make RUNDIR=/var/run/ppp LOGDIR=/var/log/ppp

And by default set RUNDIR and LOGDIR to the /etc/ppp for backward compatibility?

Regarding the pathname.c, I could add -D compile option. But I don't know what to do with the man page and scripts. Maybe the configure script support and some sed magic is needed.

Neustradamus:
@paulusmack: Have you seen the last comment of @yarda?

paulusmack:
Adding a --localstatedir parameter to configure would be fine. We would then need configure to adjust pppd.8 using sed or similar. I don't think we necessarily need to adjust everything in scripts/ since those are mostly intended as examples.

enaess:
I am working on a change that would fix this

enaess:
@yarda - do you have what you need from the configure / make system to resolve your problem? Can this issue be closed now?

Neustradamus:
@yarda: Have you seen the @enaess comment?

enaess:
I think this issue can be closed now, the autoconf stuff supports the following:

`--with-logfile-dir=DIR  Specify the log directory for pppd`

Neustradamus:
@yarda: Have you seen the @enaess comment?

cc @paulusmack.

PR #227, opened 2021-01-05
Put locks to /var/lock/ppp

yarda:

### DIFF
--- a/pppd/utils.c
+++ b/pppd/utils.c
@@ -758,7 +758,7 @@ complete_read(int fd, void *buf, size_t count)
 /* Procedures for locking the serial device using a lock file. */
 #ifndef LOCK_DIR
 #ifdef __linux__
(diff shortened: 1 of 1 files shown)


paulusmack:
Since this is about locking the serial port, the original intention was very much to interact with any other program that might want to use the serial port (e.g. getty, kermit, uucp, etc.), not just other pppds, so using a pppd-specific location defeats the purpose.

For me to take this I would have to be convinced that no current Linux distro has any defined convention for coordinating access to serial ports between different programs, and the only possible use of the 'lock' option is to lock against other pppds.

Neustradamus:
@yarda: Have you looked the @paulusmack comment?

Neustradamus:
@yarda: Have you looked the @paulusmack comment?

yarda:
@paulusmack thanks for info. It makes sense. It's not my patch, I am dropping it downstream.

PR #255, opened 2021-02-16
Minor cleanup to make use of the eui64_magic_nz() macro

njh:
`eui64_magic_nz()` is functionality equivalent to `while (eui64_iszero(wo->ourid))  eui64_magic(wo->ourid);` and is used everywhere else in `ipv6cp.c`.

### DIFF
--- a/pppd/ipv6cp.c
+++ b/pppd/ipv6cp.c
@@ -1129,9 +1129,8 @@ ipv6_check_options(void)
 	    if (!eui64_iszero(wo->ourid))
 		wo->opt_local = 1;
 	}
-	
-	while (eui64_iszero(wo->ourid))
-	    eui64_magic(wo->ourid);
+
+	eui64_magic_nz(wo->ourid);
     }
 
     if (!wo->opt_remote) {


pali (changes requested):
> `eui64_magic_nz()` is functionality equivalent to `while (eui64_iszero(wo->ourid)) eui64_magic(wo->ourid);`

Nope, it is not! Please look at implementation correctly:
https://github.com/paulusmack/ppp/blob/98ec18f098e5ef68e3a8cc6954fcaf5a7fb8b7be/pppd/eui64.h#L87-L89

With this your change you are completely breaking `ipv6cp-use-ipaddr` option for local IPv6 address as you are unconditionally overwriting already set IPv6 address from previous steps.

njh:
Whoops 😞  Thanks for reviewing @pali 

Do you think it would be impossibly hard work to create an automated test-suite for pppd?

pali:
> Do you think it would be impossibly hard work to create an automated test-suite for pppd?

Should be possible. But I think it would be very boring as it would be required to write test for every pppd option and as you can see in manual page, pppd supports tons of options... And different networks, protocols and people are using different combinations.

PR #257, opened 2021-02-22
pppd: defaultroute6 is not frequent option and should not be used

pali:
In pppd.8 manpage move defaultroute6 option from FREQUENTLY USED OPTIONS
into OPTIONS and add a warning that this option should not be needed or
used on IPv6 networks. Option defaultroute6 is needed only for broken IPv6
networks. Also add nodefaultroute6 into sample options file.

### DIFF
--- a/pppd/pppd.8
+++ b/pppd/pppd.8
@@ -127,12 +127,6 @@ is no other default route with the same metric.  With the default
 value of -1, the route is only added if there is no default route at
 all.
 .TP
-.B defaultroute6
-Add a default IPv6 route to the system routing tables, using the peer as
-the gateway, when IPv6CP negotiation is successfully completed.
-This entry is removed when the PPP connection is broken.  This option
-is privileged if the \fInodefaultroute6\fR option has been specified.
-.TP
 .B replacedefaultroute
 This option is a flag to the defaultroute option. If defaultroute is
 set and this flag is also set, pppd replaces an existing default route
@@ -353,6 +347,17 @@ escaped for both the transmit and the receive direction.
 Disable MRU [Maximum Receive Unit] negotiation.  With this option,
 pppd will use the default MRU value of 1500 bytes for both the
 transmit and receive direction.
+.TP
+.B defaultroute6
+Add a default IPv6 route to the system routing tables, using the peer as
+the gateway, when IPv6CP negotiation is successfully completed.
+This entry is removed when the PPP connection is broken.  This option
+is privileged if the \fInodefaultroute6\fR option has been specified.
+\fBWARNING: Do not enable this option by default\fR.  IPv6 routing tables
+are managed by kernel (as apposite to IPv4) and IPv6 default route is
+configured by kernel automatically too based on ICMPv6 Router Advertisement
+packets.  This option may conflict with kernel IPv6 route setup and should
+be used only for broken IPv6 networks.
 .TP
 .B deflate \fInr,nt
 Request that the peer compress packets that it sends, using the
--- a/sample/options
+++ b/sample/options
@@ -32,6 +32,12 @@ noipdefault
 # enable this on a server that already has a permanent default route
 #nodefaultroute
 
+# Default IPv6 route is automatically configured by kernel based on
+# received ICMPv6 Router Advertisement packets.
+# pppd should not touch default IPv6 route to prevent breaking IPv6 setup.
+# Enabling defaultroute6 is needed only for broken IPv6 setup.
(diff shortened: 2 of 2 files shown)


doverride:
Excuse my ignorance. I use pppd in openwrt and i noticed that with the latest release pppd started to write the routing table. I stumbled upon this. If nodefaultroute6 should not be needed, should it then not be a default. If true then why is nodefaultroute6 explicitly set in sample/config indicating that it is not a default? Should defaults not be commented out in configs instead?

pali:
@doverride: Seems that in your post you are confused with **no**defaultroute6 and defaultroute6. I think it is documented in manual page cleanly. If **no**defaultroute6 is specified (e.g. in config file) then defaultroute6 option is privileged operation. So ordinary non-root users cannot use defaultroute6 option. By default both **no**defaultroute6 and defaultroute6 options are disabled, meaning that defaultroute6 is *not* privileged operation and that by default pppd does not insert default ipv6 device route. And in sample option file is specified **no**defaultroute6 as it is safe default to prevent ordinary non-root users to use defaultroute6 option. Is it clear now?

doverride:
pali <notifications@github.com> writes:

> @doverride: Seems that in your post you are confused with nodefaultroute6 and defaultroute6. I think
> it is documented in manual page cleanly. If nodefaultroute6 is specified (e.g. in config file) then
> defaultroute6 option is privileged operation. So ordinary non-root users cannot use defaultroute6
> option. By default both nodefaultroute6 and defaultroute6 options are disabled, meaning that
> defaultroute6 is not privileged operation and that by default pppd does not insert default ipv6 device
> route. And in sample option file is specified nodefaultroute6 as it is safe default to prevent
> ordinary non-root users to use defaultroute6 option. Is it clear now?

Yes, Thanks.

Still wondering why all of a sudden pppd write the routing table with
latest release but that is unrelated and not important

>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.
>

-- 
gpg --locate-keys dominick.grift@defensec.nl
Key fingerprint = FCD2 3660 5D6B 9D27 7FC6  E0FF DA7E 521F 10F6 4098
https://sks-keyservers.net/pks/lookup?op=get&search=0xDA7E521F10F64098
Dominick Grift

pali:
> Still wondering why all of a sudden pppd write the routing table with latest release

Maybe some another OpenWRT patch which can be harmful in some situation? Or does it happens also with upstream pppd without any custom patches?

doverride:
pali <notifications@github.com> writes:

>  Still wondering why all of a sudden pppd write the routing table with latest release
>
> Maybe some another OpenWRT patch which can be harmful in some situation? Or does it happens also with
> upstream pppd without any custom patches?

Maybe yes, they carry quite a few and they also dropped a few that were
adopted upstream. Not sure whether this also happens with vanilla
upstream, I might look into that but it does not visibly break anything.

Thanks again

>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.
>

-- 
gpg --locate-keys dominick.grift@defensec.nl
Key fingerprint = FCD2 3660 5D6B 9D27 7FC6  E0FF DA7E 521F 10F6 4098
https://sks-keyservers.net/pks/lookup?op=get&search=0xDA7E521F10F64098
Dominick Grift

Neustradamus:
@hauke, please read the talk from @doverride here, it is linked to OpenWrt.

hauke:
@doverride  between which version did you see the changed behavior? 
did it change after this commit https://github.com/openwrt/openwrt/commit/030bc39c350d301d3cbca4725e845a752c359c5f or do you compare OpenWrt 19.07 with master or something similar?

doverride:
Hauke Mehrtens <notifications@github.com> writes:

> @doverride between which version did you see the changed behavior?
> did it change after this commit openwrt/openwrt@030bc39 or do you compare OpenWrt 19.07 with master or
> something similar?

I only do master, I noticed the change with the last commit on
master: "hauke ppp: Update to version 2.4.9 5 days ago"

>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.
>

-- 
gpg --locate-keys dominick.grift@defensec.nl
Key fingerprint = FCD2 3660 5D6B 9D27 7FC6  E0FF DA7E 521F 10F6 4098
https://sks-keyservers.net/pks/lookup?op=get&search=0xDA7E521F10F64098
Dominick Grift

Issue #258, opened 2021-02-23
Need access to EAP-TLS's MSK


enaess:
I've spent better part of 3 hours on a call with the Microsoft specifications team debugging the SSTP protocol to resolve an issue related to SSTP and using the MPPE keys derived to produce the correct crypto binding. Apparently, to resolve this I need the entire 32-bytes of the mppe_send_key instead of the concatenating 16-bytes from each send/receive key. 

Looking at the eap-tls code, this can be fixed by expanding the mppe-send/recv-key in both places to provide sufficient space to store the 32-bytes for each send/recv key produced by this function. Other places needs to be fixed to use only the required 16-bytes where necessary i.e. chap_ms.c.

@jjkeijser you have any comments on this. I can likely get a PR for you tomorrow.

```C
    dbglog("EAP-TLS PRF label = %s", prf_label);
    prf_size = strlen(prf_label);
    if (SSL_export_keying_material(ets->ssl, out, sizeof(out), prf_label, prf_size, 
                                   context, context_len, 0) != 1)
    {    
        warn( "EAP-TLS: Failed generating keying material" );
        return;
    }    

    /*   
     * We now have the master send and receive keys.
     * From these, generate the session send and receive keys.
     * (see RFC3079 / draft-ietf-pppext-mppe-keys-03.txt for details)
     */
    if (client)
    {    
        p = out; 
        BCOPY( p, mppe_send_key, sizeof(mppe_send_key));
        p += EAPTLS_MPPE_KEY_LEN;
        BCOPY( p, mppe_recv_key, sizeof(mppe_recv_key));
    }    
    else 
    {    
        p = out; 
        BCOPY( p, mppe_recv_key, sizeof(mppe_recv_key));
        p += EAPTLS_MPPE_KEY_LEN;
        BCOPY( p, mppe_send_key, sizeof(mppe_send_key));
    }    

    mppe_keys_set = 1; 
````

**Comment by jjkeijser** [comment]
whoa , I am very confused here...   
* the MS SSTP spec says they follow RFC3079 for MPPE key generation
* RFC3079 states that each MPPE key is 16 bytes longs. 
* note that MPPE key derivation works the same for MS-CHAP as EAP-TLS and that this code was added in ~ v2.4.4, prior to any EAP-TLS stuff

* furthermore the SSTP spec v20180912 says:
Higher-Layer Authentication Key (HLAK)
```
[...]

PR #259, opened 2021-02-23
Allowing storage space for 32-bytes MPPE keying material.

enaess:
Fix for issue #258 
This *crude* patchset expands the current size of the mppe_send|recv_key buffers to store up to 32 bytes of keying material. Note that the MPPE key passed up to the kernel is still 16 bytes (will not work with larger keys). 

I am not necessarily advocating for taking this fix, but more of a proof of concept. There are probably better ways to encapsulate the keys (e.g write an mppe.c file that allows for set/get/clear of current mppe keys, moving it away from the chap_ms.c file), etc.

### DIFF
--- a/pppd/ccp.c
+++ b/pppd/ccp.c
@@ -1427,8 +1427,8 @@ ccp_up(fsm *f)
 	notice("%s transmit compression enabled", method_name(ho, NULL));
 #ifdef MPPE
     if (go->mppe) {
-	BZERO(mppe_recv_key, MPPE_MAX_KEY_LEN);
-	BZERO(mppe_send_key, MPPE_MAX_KEY_LEN);
+	BZERO(mppe_recv_key, sizeof(mppe_recv_key));
+	BZERO(mppe_send_key, sizeof(mppe_send_key));
 	continue_networks(f->unit);		/* Bring up IP et al */
     }
 #endif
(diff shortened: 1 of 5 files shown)


Neustradamus:

paulusmack:
This PR is obviously just preparation for something else, and it's hard to evaluate it without seeing how the larger buffers are going to be used. I assume that in future that sometimes only 16 bytes will be used and sometimes (depending on the particular authentication method used?) up to 32 bytes will be used.

It seems to me that if you are going to make the buffers bigger but still only put 16 bytes in them sometimes, then you probably need a variable to tell the consumers of the material how many valid bytes are in there.

My main question would be whether this PR is really ready to go, or are you likely to want to do things slightly differently once you actually have producers and consumers using more than 16 bytes?

enaess:
Good point, but as far as "breaking" the ABI by expanding the size of the keys, you are okay with this fix? I can add another length variable to the fix before you merge it. 

I just made a bunch of fixes to network-manager-sstp that uses this expanded key and it would be very helpful to get this or a better fix merged.

@paulusmack ?

paulusmack:
If you are going to need a length variable then add it here.

I would also feel more comfortable with the change if you zeroed the whole of the buffers before copying in the key material.

enaess:
Closing this PR as #267 is the real fix for the issue.

Issue #250, opened 2021-02-13
EAP-TLS to expose hook to verify peername


enaess:
You have to specify "remotename" which may not be initially known to the end-user. For example, the server name you connect to may be azure-gw-xyzxyz.<somedomain>.com, but the name of the server which performs the EAP authentication may be a server internal to the network you are accessing. Omitting the "remotename" will cause PPP client to terminate the connection.

I'd like for ppp when using EAP-TLS to expose a hook/function that a plug-in can help validate the remote name (or not). Similarly to exposing the parameters for TLS authentication when using OpenVPN. 

See the attached screenshot. 

![Screenshot from 2021-02-12 16-32-28](https://user-images.githubusercontent.com/1057596/107835894-f2e6fb80-6d4f-11eb-8983-cf3ca2f330c6.png)

enaess:
@jjkeijser - Do yo have an opinion on this?

Neustradamus:
To follow this ticket.

jjkeijser:
I'm a bit confused here - this is already possible **without** a hook:
```
  remotename  </CN= part of server certificate>
```

which will connect to a server only if the server certificate name matches **exactly** what is specified in `remotename`.

Having said that, adding a hook/mechanism similar to what openvpn has is possible (e.g. wildcard matching) but I'd much rather implement something similar to what openvpn does (allow wildcards in `remotename` ) than add a hook to call some external code (**security risk!**)

enaess:
@jjkeijser 

Yes, the base-case feature of "exact match" works out of the box for "remotename". 

This however, is a problem for the end user when he connects for the first time. The "remotename" can be something widely different from the SSTP server s/he connects to, which isn't known beforehand. Currently, the only way to infer this on the client side is by looking in the logs (not very user friendly). In addition, there might be a pool of RADIUS servers, each server with a different certificate / subject name.

The OpenVPN configuration allows for the following: 
* Don't verify name (default)
* Verify whole subject exactly
* Verify name exactly
* Verify name by prefix
* Verify subject partially (Legacy Mode, strongly discouraged)

While the administrator can provide the user this information when he configures it, I've found that when authenticating to a Azure Cloud, the authentication provider is a permutation of the gateway instance name (automatically generated?). Which is a real problem for the end user, though it should probably be best resolved with a suffix match.

[...]

PR #261, opened 2021-03-05
EAP-TLS: Verify Subject or CommonName by suffix

enaess:
Closes Issue #250

This feature matches closely what OpenVPN and the network-manager-openvpn plugin does for
certificate verification. It allows the end user to configure the certificate to be
matched by it's common name (entire string), it's subject name, or the suffix of a subject name
the latter is especially useful if you are trying to match against a random server in a RADIUS
pool. Lastly, it also allow you to turn off the certificate matching all together.

tls-verify-method can have the following parameter values:
- none
- subject   - The entire subject, e.g. /CN=some.server.org
- name      - The entie common name, e.g. some.server.org
- suffix    - The latter part of a name, e.g. servers.org

Secondly, it also introduces a new parameter 'tls-verify-key-usage' which permits checking
of the 'server' or 'client' side purpose and X509 extended key attributes.
For example, in client mode, it will verify that received certificate has the 'server' side
attributes enabled.

### DIFF
--- a/README.eap-tls
+++ b/README.eap-tls
@@ -126,7 +126,7 @@ EAP-TLS authentication support for PPP
 
       ca <ca-file>
         Use the CA public certificate found in <ca-file> in PEM format
-      ca-path <directory>
+      capath <directory>
         Use the directory <directory> as the CA public certificate directory
       cert <cert-file>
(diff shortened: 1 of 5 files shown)


Neustradamus:

Issue #218, opened 2021-01-05
Logic error in chap_ms.c


paulusmack:
From David Binderman:

chap_ms.c:956:44: warning: logical ‘or’ of collectively exhaustive tests is always true [-Wlogical-op]

Source code is

   if (policy != MPPE_ENC_POL_ENC_ALLOWED ||
        policy != MPPE_ENC_POL_ENC_REQUIRED)
        return;

Maybe better code:

   if (policy != MPPE_ENC_POL_ENC_ALLOWED &&
        policy != MPPE_ENC_POL_ENC_REQUIRED)
        return;

Neustradamus:
@enaess, @tisj, @jjkeijser, @dankor, @nomis: What do you think?

nomis:
All this function does is disable 40-bit encryption if 128-bit encryption is selected, and disable 128-bit encryption if 40-bit encryption is selected. The only place this is used is from the RADIUS plugin.

The proposed change makes sense but the function doesn't really use the `policy` argument. It's referenced again in the next `if` statement but only if `ccp_wantoptions[0].mppe` is 0. If `ccp_wantoptions[0]` is 0 then the rest of the function can't do anything meaningful.

Arguably the `policy` value should be able to require MPPE too but it currently doesn't?

This is all a bit redundant given that MPPE uses RC4 which is sufficiently broken that no-one should still be using it.

tisj:
The policy is not enforced in this function (but elsewhere?), it is only meant to restrict the allowed encryption types.

The specification says that when encryption is required, any of the encryption types may be used, but at least one must be chosen. In that sense the function is somewhat meaningful.

It seems that the original writer may have wanted to express:

```
    if (!(policy == MPPE_ENC_POL_ENC_ALLOWED ||
            policy == MPPE_ENC_POL_ENC_REQUIRED))
```

Which is logic wise the same as the proposed fix. Whatever is more readable...

enaess:
I'd say, fix the warning message by the compiler and move on. @nomis is likely correct in some parts of this, but the use of mppe seems like it needs some more work, see issue #57. 

The RC4 / MPPE encryption is a compatibility problem more than it is a security problem. For SSTP, the tunnel is already encrypted, but Microsoft servers still insist on enabling the MPPE option less one modifies the NAP policies to turn it off. 

I'd also like to see issue #57 fixed, or perhaps be the default behavior such that the client doesn't insist on the MPPE being enabled but will allow it if server requires it.

[...]

PR #262, opened 2021-03-08
Fix for issue #218, a logical error in comparing valid encryption policies

enaess:
RFC2548 describes the proper values of the MS-MPPE-Encryption-Policy attribute.
and it can only hold 2 values: 1 (encryption allowed) and 2 (encryption required).

See
   https://tools.ietf.org/html/rfc2548, section 2.4.4

The correct comparison should be made with an && and not a ||


Neustradamus:

PR #128, opened 2019-10-21
pppd: add experimental support for PEAP protocol, an extension of EAP

rustylife:
current patch implements client functionality for PEAPv0/EAP-MSCHAPv2,
which is most common setup deployed by companies utilizing
Microsoft RRAS as their VPN solution

Signed-off-by: Rustam Kovhaev <rkovhaev@gmail.com>

PR #264, opened 2021-03-08
Allow use of PCKS12 certificates when using EAP-TLS

enaess:
This implements the ability to specify the option 'pkcs12' to allow users to provide a PKCS12 formatted file as user credentials. I've tested the current code using PEM, Encrypted Key with PEM, PKCS12. Still pending verification of PKCS11 tokens.

### DIFF
--- a/pppd/auth.c
+++ b/pppd/auth.c
@@ -255,6 +255,7 @@ char *cacert_file  = NULL;	/* CA certificate file (pem format) */
 char *ca_path      = NULL;	/* directory with CA certificates */
 char *cert_file    = NULL;	/* client certificate file (pem format) */
 char *privkey_file = NULL;	/* client private key file (pem format) */
+char *pkcs12_file  = NULL;	/* client private key envelope file (pkcs12 format) */
 char *crl_dir      = NULL;	/* directory containing CRL files */
 char *crl_file     = NULL;	/* Certificate Revocation List (CRL) file (pem format) */
 char *max_tls_version = NULL;	/* Maximum TLS protocol version (default=1.2) */
@@ -445,6 +446,7 @@ option_t auth_options[] = {
     { "key", o_string, &privkey_file, "EAP-TLS client private key in PEM format" },
     { "crl-dir", o_string, &crl_dir,  "Use CRLs in directory" },
     { "crl", o_string, &crl_file,     "Use specific CRL file" },
+    { "pkcs12", o_string, &pkcs12_file, "EAP-TLS client credentials in PKCS12 format" },
     { "max-tls-version", o_string, &max_tls_version,
       "Maximum TLS version (1.0/1.1/1.2 (default)/1.3)" },
     { "tls-verify-key-usage", o_bool, &tls_verify_key_usage,
@@ -2464,6 +2466,8 @@ have_eaptls_secret_client(char *client, char *server)
 
 	if ((cacert_file || ca_path) && cert_file && privkey_file)
 		return 1;
+	if (pkcs12_file)
+		return 1;
 
     filename = _PATH_EAPTLSCLIFILE;
     f = fopen(filename, "r");
@@ -2647,7 +2651,7 @@ scan_authfile_eaptls(FILE *f, char *client, char *server,
 int
 get_eaptls_secret(int unit, char *client, char *server,
 		  char *clicertfile, char *servcertfile, char *cacertfile,
-		  char *capath, char *pkfile, int am_server)
+		  char *capath, char *pkfile, char *pkcs12, int am_server)
 {
     FILE *fp;
     int ret;
@@ -2661,6 +2665,7 @@ get_eaptls_secret(int unit, char *client, char *server,
 	bzero(cacertfile, MAXWORDLEN);
 	bzero(capath, MAXWORDLEN);
 	bzero(pkfile, MAXWORDLEN);
+	bzero(pkcs12, MAXWORDLEN);
 
 	/* the ca+cert+privkey can also be specified as options */
 	if (!am_server && (cacert_file || ca_path) && cert_file && privkey_file )
@@ -2671,6 +2676,14 @@ get_eaptls_secret(int unit, char *client, char *server,
 		if (ca_path)
 			strlcpy( capath, ca_path, MAXWORDLEN );
 		strlcpy( pkfile, privkey_file, MAXWORDLEN );
+	}
+    else if (!am_server && pkcs12_file)
+	{
+		strlcpy( pkcs12, pkcs12_file, MAXWORDLEN );
+		if (cacert_file)
+			strlcpy( cacertfile, cacert_file, MAXWORDLEN );
+		if (ca_path)
+			strlcpy( capath, ca_path, MAXWORDLEN );
 	}
 	else
 	{
--- a/pppd/eap-tls.c
+++ b/pppd/eap-tls.c
@@ -42,6 +42,7 @@
 #include <openssl/err.h>
 #include <openssl/ui.h>
 #include <openssl/x509v3.h>
+#include <openssl/pkcs12.h>
 
 #include "pppd.h"
 #include "eap.h"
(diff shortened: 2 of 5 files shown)


Neustradamus:

pppd/auth.c
```diff
@@ -2657,7 +2661,7 @@ get_eaptls_secret(int unit, char *client, char *server,
 	bzero(pkfile, MAXWORDLEN);
 
 	/* the ca+cert+privkey can also be specified as options */
-	if (!am_server && (cacert_file || ca_path) && cert_file && privkey_file )
```
paulusmack, line 2660: It seems like this change doesn't change the actual logic, so why the extra pair of parentheses?

pppd/auth.c
```diff
@@ -2666,6 +2670,14 @@ get_eaptls_secret(int unit, char *client, char *server,
 			strlcpy( capath, ca_path, MAXWORDLEN );
 		strlcpy( pkfile, privkey_file, MAXWORDLEN );
 	}
+    else if (!am_server && pkcs12_file)
```
paulusmack, line 2673: The pkcs12 array gets initialized on this path, but not if this path isn't taken. Since the caller is passing in an uninitialized stack array, and you subsequently test pcks12[0], you need either to initialize pcks12 on all paths through this function, or else initialize it in all callers (the former is probably easier).

pppd/eap-tls.c
```diff
@@ -305,25 +312,30 @@ SSL_CTX *eaptls_init_ssl(int init_server, char *cacertfile, char *capath,
     long         tls_version = TLS1_VERSION; 
 #endif
 
+    info("Using PKCS12: %s", pkcs12);
```
paulusmack: Do we really want to print this if no pkcs12 file was specified? It would be confusing to the user to print "Using PCKS12" when we aren't.

paulusmack (changes requested):

enaess:
@paulusmack  - I fixed your suggested fixes in auth.c. The fix in eap-tls.c was already handled by a later commit.

pppd/eap-tls.c
enaess: This is already fixed in a later change

enaess (commented):
I have them fixed now.

Neustradamus:
@paulusmack: What do you think about changes done by @enaess after your first review?

enaess:
@Neustradamus @paulusmack - I did a git rebase -i and pushed that change. This include my changes w/review feedback into a single change. Let me know if that helps.

Neustradamus:

enaess:
@paulusmack is there any outstanding work for me to do on this, or is it all in your ballpark?

Neustradamus:
@enaess: Rebase is needed?

enaess:
@Neustradamus I did a git rebase, but the merge is not trivial. Will have to take a deeper look tomorrow.

enaess:
@Neustradamus - Merged the conflict. @jjkeijser - Could you please sync this git branch and give it a review / spin? I've tested the pkcs12 portion of it, but I don't have a certificate device yet.

Neustradamus:
@jjkeijser: Have you done a test of it for @enaess?

enaess:
@Neustradamus - I've tested this and it works fine with pkcs12 certificate, a cert + key (pem+rsa key), and cert + encrypted key (pem + encr-rsa-key).

Neustradamus:
@enaess: Thanks for your confirmation :)

enaess:
The change is a lot more palatable now, than the previous incarnation. The flow for pkcs11 (smart cards) can be validated and it is the same as before. Having that said, I was browsing for a pkcs11 platform I could test the change with. Though I believe @jjkeijser already have this setup.

jjkeijser:
this patch fails when using a pkcs#11 device; you can test this yourself by setting up a "SoftHSM" device (e.g. see https://github.com/opendnssec/SoftHSMv2)  ; the same setup works with v2.4.9 but with this patch I get:

```
Initialising engine
Using our private key '20210420' in engine
EAP-TLS: Private key fails security check
EAP-TLS SSL error stack:
error:140A80BE:SSL routines:SSL_CTX_check_private_key:no private key assigned
eaptls_init_ssl_client: fail
```

jjkeijser:
Further comment:  you've changed
```
        if (!SSL_CTX_use_certificate_chain_file(ctx, certfile))
        {
            error( "EAP-TLS: Cannot use public certificate %s", certfile );
            goto fail;
        }
    } 
```
to
```
    if (!SSL_CTX_use_certificate(ctx, cert))
    {    
        error("EAP-TLS: Cannot use public certificate");
        goto fail;
    }    
```
and *for pkcs12 files* you throw in an extra call to `SSL_CTX_add_extra_chain_cert` later on.  That breaks the case where a user certificate PEM file contains a chain of certificates.

enaess:
@jjkeijser thanks for your feedback. Looks like I missed a use-case and broke the pkcs11 support. I'll give the SoftHSM project a spin. Though, I ordered a PVI card and a card reader from Amazon yesterday.

enaess:
@paulusmack @jjkeijser @Neustradamus 

In incorporated @jjkeijser suggested changes and re-tested the change.

*Re-cap*
Moved the SSL_CTX_use_PrivateKey() to also handle pkcs11 loaded certificates and that fixes the case where pkcs11 keys via engine was loaded. The certificate chain is null except for the case pkcs12 loaded it, and I reverted the use of loading a single certificate to load the certificate chain file. 

Re-ran my tests, works fine with pkcs12, pkcs11 and certificate + key.

enaess:
@jjkeijser @paulusmack - Any thoughts or update on this?

enaess:
A merge of this is blocking work on #275

jjkeijser:
Apologies for the late response:  
I have looked at the patch, tested it against a hardware PKCS#11 device (so not softhsm2) and all works as it should.
Patch ACKed.

enaess:
@jjkeijser that's terrific! Thanks for the thumbs up :) - Now if we can get @paulusmack to merge it, then I can fix the pkcs11: URI path. Next would be adding ./configure support for using openssl crypto functions vs. built in ones. And then get the PEAP-patch working.

jjkeijser:
I am not so sure about using openssl crypto functions instead of the built in ones ...  keep in mind that most crypto stuff used by pppd is quite old (md2, md4, md5, sha1) and will quite likely be dropped from openssl in future builds - or at the very least some distro's will build openssl without support for it. In that case you will want to fall back on the builtin code, which is quite small anyways.    I'd say this is *not* worth the effort.

paulusmack:
> I am not so sure about using openssl crypto functions instead of the built in ones ... keep in mind that most crypto stuff used by pppd is quite old (md2, md4, md5, sha1) and will quite likely be dropped from openssl in future builds - or at the very least some distro's will build openssl without support for it. In that case you will want to fall back on the builtin code, which is quite small anyways. I'd say this is _not_ worth the effort.

I agree, these functions are not very much code and having our own versions insulates us against changes in the library APIs.  It also means that for embedded uses, one doesn't have to build in the whole of openssl.

enaess:
How about I look into pull request #128, and the related issue #265? I don't think renaming the functions in sha1 would be the right solution, rather; this code do require use of OpenSSL. If we are using OpenSSL, then there is no real purpose of including built in sha1/md5 unless they are missing in the openssl implementation (detection of the presence of said functions is really a functionality of the configure script). Not sure where I end up with this, but I'd like to see PEAP support in pppd.

paulusmack:
> How about I look into pull request #128, and the related issue #265? I don't think renaming the functions in sha1 would be the right solution, rather; this code do require use of OpenSSL. If we are using OpenSSL, then there is no real purpose of including built in sha1/md5 unless they are missing in the openssl implementation (detection of the presence of said functions is really a functionality of the configure script). Not sure where I end up with this, but I'd like to see PEAP support in pppd.

Yes, please look into that pull request and issue, that would be appreciated.

Regarding sha1 and md5, how often does the openssl library change its API? Hopefully not very often, but it doesn't seem like the answer is "never" either. Particularly for MD5 which is used for CHAP, which is a basic part of any PPP implementation, having our own copy of a small amount of code that won't get changed at the whim of another project is attractive. I am less concerned about sha1 in that respect.

enaess:
Yeah I get it.

Suddenly, you compile with a FIPS enabled version of OpenSSL and the things blows up or just isn't there. I don't think they change their API that often, but it's annoying when they do and you have to support both. 

Then you have distributions like Debian/Ubuntu which applies a patch to remove vendor supplied crypto. I could imagine some of the embedded stuff too would like to have a configuration to eliminate duplicate features ...

PR #267, opened 2021-03-21
Expose the MPPE keys generated

enaess:
The lengthy part of this fix is to refactor the handling of MPPE keys
by moving it into mppe.c and thus reducing the clutter in chap_ms.c.
It does so by renaming the mppe_set_keys/2 to the corresponding
mppe_set_chapv1/mppe_set_chapv2 versions and updates callers of these
functions.

Having done so, it conveniently allows the name "mppe_set_keys" to be
re-used for this new purpose which will copy the key material up to
it's size and then clear the input parameters (avoids leaving the MPPE
keys on the stack).

Additional functions exported via the MPPE code to allow plugins et al. to
access the MPPE keys, clear the keys, and check if they are set. All
plugins and CCP code has been updated to use this API.

This fixes GitHub Issue #258 

### DIFF
--- a/pppd/Makefile.linux
+++ b/pppd/Makefile.linux
@@ -106,6 +106,8 @@ CFLAGS   += -DMSLANMAN=1
 endif
 ifdef MPPE
 CFLAGS   += -DMPPE=1
+PPPDOBJS += mppe.o
+PPPDSRC  += mppe.c
 HEADERS  += mppe.h
 endif
 endif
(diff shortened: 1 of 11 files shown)


enaess:
@paulusmack what do you think?

enaess:
@paulusmack I updated the change comment and title of the PR to be more informative (and not to start with "Fix for issue. .."

Neustradamus:
Thanks @enaess and @paulusmack!

Issue #275, opened 2021-05-10
Support RFC7512 format for specifying PKCS11 keys with EAP-TLS 


enaess:
I am on Ubuntu 20.04, and I tried to use SoftHSM and OpenSC to use PKCS11 tokens with EAP-TLS. 

Relevant packages:
* OpenSC (apt-get install opensc)
* SoftHSM (apt-get install softhsm2)
* libp11 from OpenSC (apt-get install libengine-pkcs11-openssl1.1)

As of current, using pppd w/pkcs11 support uses the old notation of pkcs:<slot>:<id>. If your certificate is located in slot 0, and with an id of 04, you specify this using pkcs11:0:04 for certfile and keyfile parameters.

However, this notation is considered "deprecated" or "legacy" with the recent version of OpenSSL's engine provided by the OpenSC project (pkcs11.so). Going forward, pppd needs to pass the certfile / keyfile parameters directly to the engine (without stripping "pkcs11:" prefix, and then allow the user to specify the certificate / key using RFC7512 (The PKCS11 URI scheme).

Current documentation of examples leads the user to think the output of the p11tool which lists the PKCS11 URI be the value to pass as the cert/key parameters to pppd. This will currently not work.

Neustradamus:
@enaess: This ticket can be closed no?

PR #280, opened 2021-06-05
Allow users to use Engine and ID strings unmodified to the OpenSSL engine

enaess:
This change allows users to specify "cert" and "key" arguments where the prefix leading up to the ':' indicates the engine e.g. pkcs11, the entire string is passed down to the engine library which does the parsing. In the case of pkcs11.so (OpenSC project's libp11) engine for OpenSSL, it will now correctly parse the PKCS11 URI string (See RFC7512).

### DIFF
--- a/pppd/eap-tls.c
+++ b/pppd/eap-tls.c
@@ -288,9 +288,8 @@ SSL_CTX *eaptls_init_ssl(int init_server, char *cacertfile, char *capath,
 {
 #ifndef OPENSSL_NO_ENGINE
     char        *cert_engine_name = NULL;
-    char        *cert_identifier = NULL;
     char        *pkey_engine_name = NULL;
-    char        *pkey_identifier = NULL;
+    char        *idx;
 #endif
     SSL_CTX     *ctx;
     SSL         *ssl;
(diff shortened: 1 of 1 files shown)


enaess:
@jjkeijser - Could you please review this.

jjkeijser:
just reviewed it:  this is not **fully** rfc7512 compliant, as that says you **shall** start with "pkcs11:"  while you are just checking for ":" ; I'd recommend to switch this to
```
if ( strncmp( certfile, "pkcs11:", 7 ) == 0 )
```
and similarly for
```
if ( strncmp( privkeyfile, "pkcs11:", 7 ) == 0 )
```
Other than that the patch looks fine.

enaess:
@jjkeijser I did the consideration of just doing a strncmp("pkcs11:"..), but opted to keep it generic as the parsing is done by the pkcs11 library / module provided to OpenSSL. 

The generic mechanism now expects the <engine>:<identifier> as before. No documentation update is needed. If someone provides a different OpenSSL engine, they have a chance to look at the entire string. 

The title of the check-in and/or bug can be changed to remove the claim of support for the RFC. I can go either way ... Just let me know

Eivind

jjkeijser:
Hi,

I think the easiest solution is to drop the rfc7512 compliance, or water it down a bit, to 'more in line with rfc7512'.
It's just that I think we should avoid claiming compliance which is not true. 

On 11 Jun 2021, 18:20, at 18:20, "Eivind Næss" ***@***.***> wrote:
***@***.*** I did the consideration of just doing a

enaess:
I concur, and I'll have the updated fix some time this weekend.

enaess:
@jjkeijser PR and CL updated. This closes issue #275

enaess:
@paulusmack this PR is ready for merge.

PR #285, opened 2021-06-16
PPP enhanced ifname + RFC8044 dictionary compatibility

EasyNetDev:
Hi,

I did a small change in dict.c source code to accept ipv4addr RADIUS data type in parallel with ipaddr.
In RFC8044 ipaddr is not anymore present, but instead ipv4addr is present. 
All dictionaries are comming with ipv4addr instead ipaddr and PPP is generating an error if you are not correcting the dictionary file.

### DIFF
--- a/pppd/utils.c
+++ b/pppd/utils.c
@@ -137,6 +137,7 @@ vslprintf(char *buf, int buflen, char *fmt, va_list args)
     int c, i, n;
     int width, prec, fillch;
     int base, len, neg, quoted;
+    long lval = 0;
     unsigned long val = 0;
     char *str, *f, *buf0;
     unsigned char *p;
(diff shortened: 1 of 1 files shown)


paulusmack:
Why are there 3 commits, which first make a change, then undo the change, then do exactly the same change again?

EasyNetDev:
I reverted the first one. I forgot a parenthesis in the first commit.
I'm not used with git. That's why you see 3 commits: first the wrong one with missing parenthesis, second one I tried to revert to original source and the 3rd is the correct one and tested.

EasyNetDev:
Sorry, but I tried to create a new pull request and somehow is added in my previous pull request. Now I notice my new patch is in the same pull request with RFC8044

Neustradamus:
@EasyNetDev: I think that @paulusmack has requested you to do it: https://dev.to/andreisfedotov/how-to-join-several-commits-into-one-323j.

Note: Your username in your git settings is your old username...

PR #290, opened 2021-07-02
radattr: tighten permissions on radattr file to avoid information lea…

jkroonza:
…kage.

Depending on the invoking process's umask it's possible that the radattr
file (which in certain cases can contain crytographic keys) be stored
with permissions such that world-read access is possible, resulting in
sensitive information being leaked to local users.

Signed-off-by: Jaco Kroon <jaco@iewc.co.za>


paulusmack:
Looks fine, thanks.

PR #291, opened 2021-07-02
Add RFC8044 dictionary compatibility for IPv4 address

EasyNetDev:
This patch adds ipv4addr RADIUS data type compatible with RFC8044.
New dictionaries from RADIUS is using ipv4addr instead of old ipaddr data type. This patch is avoiding modification of RADIUS dictionaries to be compatible with PPP.

### DIFF
--- a/pppd/plugins/radius/dict.c
+++ b/pppd/plugins/radius/dict.c
@@ -150,7 +150,7 @@ int rc_read_dictionary (char *filename)
 			{
 				type = PW_TYPE_INTEGER;
 			}
-			else if (strcmp (typestr, "ipaddr") == 0)
+			else if (strcmp (typestr, "ipaddr") == 0 || strcmp (typestr, "ipv4addr") == 0)
 			{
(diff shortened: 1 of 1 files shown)


EasyNetDev:
@paulusmack could you please check if is OK like this?

jkroonza (commented):
This looks good to me, and I think sorts one of my comments on the PR regarding the interface name %fmt issues.

That said, I don't see this in any of the freeradius dictionary files, which all use ipaddr.  So this really just future-proofs things.

I don't have merge/approval rights, but this looks good to me.

jkroonza:
> Hi @jkroonza,
> 
> After I update my radius client in Debian Bullseye I got this file, for example:

-- snip --

> As you can see they started to move to ipv4addr instead ipaddr. In the dictionary.microsoft I can still find ipaddr.

ACK.

I'm using freeradius currently, I might be a few versions behind either way on the host I used to check.  As stated:  I agree this should ship.  Based on your experience, sooner rather than later.

PR #293, opened 2021-07-09
Update pppd/pppd.h (net/ppp_defs.h does not exist on musl libc)

scardracs:
Hi,

when I try to build packages like NetworkManager on Gentoo Linux with musl, it ends saying net/ppp_defs.h does not exist, because it is in linux/ppp_defs.h. This fix that problem.

EDIT

musl does not define any include/net/ppp_defs.h while glibc does with something like that

#ifndef _NET_PPP_DEFS_H  
#define _NET_PPP_DEFS_H 1  
  
#include <bits/types/time_t.h>  
#include <asm/types.h>  
#include <linux/ppp_defs.h>  
  
#endif /* net/ppp_defs.h */  

That's why on glibc it does not fail at build while musl does

EDIT 2

Following what I've said before you can check that include/bits/types/time_t.h does not exist on musl too (but seems unnecessary with it)

Signed-off-by: Marco Scardovi <marco@scardovi.com>


scardracs:
I've tested this change in linux-5.13.1

paulusmack:
I doubt this would compile on Solaris. I have compiled with musl previously without problems, so I wonder where it broke?

scardracs:
> I doubt this would compile on Solaris. I have compiled with musl previously without problems, so I wonder where it broke?

I don't really know as net/ppp_defs.h does not exist (I've checked since 4.4 and it was always under include/linux)

scardracs:
That's normal as musl does not handle u_int32_t as it isn't a POSIX standard value but only valid for glibc

enaess:
This would be an excellent candidate for the autotools change I've been working on. I am not a big fan of pulling in the include/* directory with a bunch of system specific includes. Instead, the configure script should detect the presence of system headers and use them if they exist, the include/* headers would be a fallback if everything else fails.

mrgtwentythree:
net/ppp_defs.h is where the file is installed for at least netbsd, and
probably other BSDs as well, so this change would be bad there.

perhaps you want to spell this as

#ifdef __linux__
#include <linux/ppp_defs.h>
#else 
#include <net/ppp_defs.h>
#endif 

which would likely take care of solaris build too.

scardracs:
> net/ppp_defs.h is where the file is installed for at least netbsd, and
> probably other BSDs as well, so this change would be bad there.
> 
> perhaps you want to spell this as
> 
> #ifdef **linux**
> #include <linux/ppp_defs.h>
> #else
> #include <net/ppp_defs.h>
> #endif
> 
> which would likely take care of solaris build too.

looks better now?

scardracs:
Updated OP with more infos

mrgtwentythree:
> > net/ppp_defs.h is where the file is installed for at least netbsd, and
> > probably other BSDs as well, so this change would be bad there.
> > perhaps you want to spell this as
> > #ifdef **linux**
> > #include <linux/ppp_defs.h>
> > #else
> > #include <net/ppp_defs.h>
> > #endif
> > which would likely take care of solaris build too.
> 
> looks better now?

yes.  thanks.

enaess:
@scardracs it would be nice if you could give #296 a spin. It may not solve your particular problem (yet), but would serve as another data point.

scardracs:
> @scardracs it would be nice if you could give #296 a spin. It may not solve your particular problem (yet), but would serve as another data point.

I serious think it could be a really awesome change on ppp 😉

scardracs:
Ping: run again actions, they should work now with latest commit

scardracs:
🤦🏼

Neustradamus:
@scardracs: Have you seen the result?

scardracs:
> @scardracs: Have you seen the result?

I saw it, unfortunately

scardracs:
@Neustradamus IDK why it does something like that, seems uclibc need to calls something else

pppd/pppd.h
```diff
@@ -57,7 +57,18 @@
...
 #include <sys/types.h>		/* for u_int32_t, if defined */
 #include <sys/time.h>		/* for struct timeval */
+
+/*
+ * linux kernel does not have net/ppp_defs.h by default.
+ * this fix is for musl as it does not create net/ppp_defs.h
+ * like glibc.
+*/
+
+#ifdef __linux__
```
enaess, line 67: How about:
```
#ifdef HAVE_LINUX_PPP_DEFS_H
#include <linux/ppp_defs.h>
#endif

#ifdef HAVE_NET_PPP_DEFS_H
#include <net/ppp_defs.h>
#endif
```

Then ideally, the logic of setting the defines would be in pppd/Makefile.linux, this would be something similar to how it currently detects utmp.h.

```
# Use libutil; test if logwtmp is declared in <utmp.h> to detect
UTMPHEADER = "\#include <utmp.h>"
ifeq ($(shell echo $(UTMPHEADER) | $(CC) -E - 2>/dev/null | grep -q logwtmp && echo yes),yes)
USE_LIBUTIL=y
endif
```

Now swap that out for linux/ppp_defs.h, or net/ppp_defs.h

It would be easy if the autotools change list would be merge. Your update to the pppd.h would still remain, but Makefile.linux section would be replace with an AC_CHECK_HEADER.

scardracs: The ifdef you suggest can't work as glibc has both linux/ and net/ so it would check for both. Personally I don't know about uclibc

enaess: But there should be a combination that would work for all, e.g. if have <net/ppp_defs.h> use that, or use <linux/ppp_defs.h>, vice versa

enaess: Also, to @paulusmack point, why is it that gentoo/muslc does not provide this file, and buildroot does? Did you configure the muslc library correctly, or are you missing a patch?

scardracs: The net/ppp_defs.h is not provided by kernel but libc itself. There's a patch like this one http://lists.busybox.net/pipermail/buildroot/2016-March/156823.html for it

enaess: What are you trying to build, pppd, or some third part plugin? The pppd can (and should) use the provided ppp_defs.h if the system has it missing

But if you are looking at building a ppp plugin, then you may not need this. The pppd/*.h headers is a bit of a cluster f** and needs some TLC.

scardracs: PPP itself build without problems but things like NetworkManager does not build because missing net/ppp_defs.h. when I tried changing #include <net/ppp_defs.h> to linux/ works without problems

enaess: I see, which part of network manager are you trying to compile? 

Ideally, I think the Network Manager project should include the right ppp_defs.h and use meson, or autoconf to setup the correct include prior to including ppp/pppd.h for the system it is being compiled on. 

As I am starting to think of this, the pppd project should drop the include from pppd.h and require others to fix that up. The pppd project could wrap it under #ifder/#endif stuff, but it is contingent upon configuring their system correctly with defines dictated by pppd.

That sort of dependency i.e. network-manager needs to know what define to enable to compile the source with seems like a no go, i.e. wrap it with HAVE_NET_PPP_DEFS_H.  

I've been advocating for this header include mess cleanup for a while...

scardracs:
I suggest to close this PR until I fix it properly

Neustradamus:
@scardracs: I think it is better to update this PR directly no?

scardracs:
> @scardracs: I think it is better to update this PR directly no?

I don't have Gentoo (linux in general) ATM due to private life needs, that's why

pppd/pppd.h
scardracs: I have an idea: Instead of using #ifdef __linux___ we could use #!defined __GLIBC__ or something similar, that's because uclibc follows glibc way while musl doesn't. Can it be a solution?

scardracs:
I'm running the actions here https://github.com/scardracs/ppp/actions/runs/1062568572

scardracs:
Please delete the workflows I've triggered 🙈

pppd/pppd.h
Neustradamus: @enaess: What do you think?

scardracs: I've tried with no luck

enaess:
I still think this header file should be removed from pppd.h and inserted in appropriate .c files. Any project that depends on this file should properly detect and include the file. We may need to add some backup defines, e.g. ifndef EAPT_MSCHAP .. define ... endif

Having that said, I want to fixup the header files and better define what header files are included as a part of the make install step vs what is private header files. The other half of this is to use the system headers, then use the included linux/ppp_defs.h as a backup ..

mjeveritt:
Inclined to agree with @enaess on this one, tbh. It seems relying on headers using ppp as a proxy is really not very smart, especially since its a custom build system already.

GorFy1 (approved):

scardracs:
@GorFy1 these changes are broken so I don't suggest to accept them

Issue #265, opened 2021-03-14
interposed functions in sha1.h


rustylife:
ppp code has 3 functions globally defined in sha1.h that interpose openssl library functions
i noticed this while trying to debug HMAC() function failure with EVP_sha1() for #128 
openssl function HMAC() calls HMAC_Init_ex(), which then calls SHA1_Init() and SHA1_Update() functions from ppp code instead of  proper function from openssl library
there are multiple ways to go about fixing this, but first i would like to get your opinion and feedback on how you would like to get it fixed

Neustradamus:
@paulusmack: What do you think?

cc: @enaess, @jjkeijser, @tisj, @pali?

enaess:
Makefile.linux defines OPENSSL if USE_SRP is configured (and is not enabled on Linux). USE_EAPTLS also depends openssl but only specifies -lssl ... Seems like some options needs better definitions. If we link with openssl, then there is no need to link with the internal sha1.h/c. 

IMO the makefile needs some cleaning up.

jjkeijser:
if you look closer you will see that the Makefile contains
  NEEDCRYPTOLIB = y
and later on that says

ifdef NEEDCRYPTOLIB
  LIBS     += -lcrypto
endif

this was done to avoid linking in '-lcrypto' multiple times. The same could be applied to the 'ifdef USE_SRP' part. However, I did not touch that part of the code and hence I also did not touch the Makefile part.

As for the sha1 functions:  I had a patch once to replace all md5+sha1 code with the OpenSSL versions but that patch was excluded from my EAP-TLs patch. You could also consider replacing the (ancient) md5 code but what would we gain?  the md4/md5/sha1 functions comprise 31kb, so it's not *that* much code to swap out.

tisj:
Bear in mind that embedded platforms like to keep dependencies minimal, and openssl is a big library. Operating systems like OpenWRT mostly use ppp for internet connectivity, and therefore don't need fancy stuff like TLS or SRP. Moreover they offer a choice in what SSL library to use for your system, such as mbedtls, ustream-ssl, wolfssl, .. and openssl.

If compatibility with openssl is an issue, I would propose to either rename (prefix) the internal crypto stuff (eg. PPP_SHA_Init) as a simple resolution, or as a bigger/modular approach make generic abstractions to support different crypto implementations, so that you can choose which one to use at build time (internal, openssl, other ssl libraries...)

jjkeijser:
[...]

PR #296, opened 2021-07-11
Use autoconf/automake to configure and make ppp

enaess:
This change brings in autoconf/automake scripts to configure the ppp project. Current change doesn't eliminate the previous build system, but the new script autogen.sh will overwrite configure, and generate the basic Makefile.in and Makefile files.

Features can now be enabled by command line:

  * Microsoft Extensions,
    - MSCHAP
    - MPPE
    - MS LAN Manager support
  * IPXCP protocol
  * CBCP protocol
  * EAP-TLS support
  * EAP-SRP support
  * Max session lifetime by byte count
  * Plugins
  * Multilink
  * IPv6 support

Control linkage with
  * OpenSSL (-lssl -lcrypto), using --with-openssl=[dir]
  * OpenSSL engine suppport --without-openssl-engine
  * systemd (-lsystemd)
  * libatm (-latm), --with-atm=[dir]
  * libsrp (-lsrp), --with-srp=[dir]
  * libpam (-lpam), --with-pam=[dir]

Also, the configure script is made sensitive to features of OpenSSL. Like the presence or absence of DES, SHA, MD4 and MD5 crypto support. In the cases where either of these are missing, the support will be directly compiled into pppd and plugins.

In addition, package maintainers can now control the installation paths with standard --prefix=, or --localstatedir=, or --sysconfdir= to configure. On top of that, they can now control the following directories:
  * runtime directory w/--with-runtime-dir
  * logfile directory w/--with-logfile-dir
  * plugin directory w/--with-plugin-dir

In the case where automake isn't the right solution, namely: SunOS kernel module build, the original Makefile infrastructure is preserved and reused.

Care was taken to only cosmetically touchup the source files in this change. This means:
  * Insert HAVE_CONFIG_H and include config.h in all .c files.
  * Change HAS_SHADOW to HAVE_SHADOW_H
  * Introduce HAVE_CRYPT_H into the source code where appropriate
  * Added ifdef MPPE where appropriate
  * USE_SRP required a few changes as it didn't compile
  * Touchup some compile warning in pppstats directory on SunOS

Introduced a new pppdconf.h file that exports the appropriate defines to a module that wants to provide a module that pppd can dynamically load. This will define/undef features like MPPE, CHAPMS such that the project doesn't have to guess what features pppd is compiled with.


enaess:
While I haven't figured out the buildroot setup, I suspect the project needs a few changes to properly incorporate the correct configuration using autotools / automake instead of it's own setup. I've included a .yaml file for doing a quick "make && make dist && make distcheck" 

Also, to create a new release you'd do a "make distcheck" to verify integrity of the build system and then grab the ppp-<version>.tar.gz file. 

I note that future work here should be to 
- Incorporate existing system's /usr/include/linux headers and build against them instead of the provided header files. There may need to be some kind of compatibility layer to provide part of the definitions until the upstream linux kernel can provide the correct exported structures, etc.
- Cleanup of the exported headers for pppd project. Some are missing include guards, others could be omitted and also be made as an include in pppd.h such that all the client source file needs to include is <pppd/pppd.h>. This would also automatically include the pppdconf.h to define the features, enabled, etc.

enaess:
Also fixed a compiler issue when using (CC=clang ./configure) clang to build pppd

enaess:
Should fix #265 
Replace #226

enaess:
@paulusmack @yarda @Neustradamus

enaess:
Before you all give me a large lecture on not to replace the existing build, please take a moment and take this change for a spin. You should be able to create the new distribution of pppd-2.4.10 by typing in "make dist" or "make distcheck" if you also wanted to verify the archive by compiling it using the VPATH option.

I've also spent a significant amount of time on the SunOS build to make sure it build on that platform as well. The autoconf/automake packages should already be installed on Solaris 11.4, but I did add a safety net on the solaris.yaml file to ensure the build would succeed with these packages installed.

enaess:
The other option would be slowly add the features into configure, which I started out with w.r.t. openssl and proper detection of features enabled in it. This would be a cumbersome process and would involve a bunch of shell scripting. Borrowing that from autoconf seems to speed things up, and make the code more "maintainable".

enaess:
I plan to include the PEAP changes here next now as #265 isn't going to be a problem.

enaess:
@RICCIARDI-Adrien Would you happen to know how to fix the buildroot CI build to use the new autoconf/automake changes here?

RICCIARDI-Adrien:
@enaess OK, I will look at this. You have to know that the CI may break with all other existing pull requests as the build steps will change, is that OK?

enaess:
Could you create a PR to my branch so when / if this gets merged, it would fix the build root stuff too?

yarda:
> @paulusmack @yarda @Neustradamus

+1 for the autotools, but I am not upstream and I didn't review the PR.

RICCIARDI-Adrien:
Hi @enaess, when executing `autoreconf.sh` followed by `configure.sh`, I get the following error : `checking for pcap_create in -lpcap... configure: error: --enable-filter is enabled by default, but libpcap is isn't found`. Is `libpcap` needed  by default for PPP ?

configure.ac
```diff
@@ -0,0 +1,420 @@
...
+
+#
+# Multilink require Trivial Database Support
+AM_CONDITIONAL(WITH_TDB, test "x${enable_multilink}" != "xno")
+AM_COND_IF([WITH_TDB],
+    AC_DEFINE([USE_TDB], 1, ["Include TDB support"]))
+
+#
+# Disable support for activating pppd on packet activity (filter support)
+AC_ARG_ENABLE([filter],
```
RICCIARDI-Adrien, line 175: Is the libpcap library detection section adapted to cross-compilation ? I'm really not used to autotools, sorry if my question is stupid !

enaess:
That was the default configuration for Linux. You able to install that dependency first, or disable it by using --disable-filter?

configure.ac
enaess: You are probably right. I'll change it to use --with-pcap and you'd want to specify the folder. 

It's just that I am not going to be able to update the change before I have a computer in front of me, which would be next Friday

RICCIARDI-Adrien: No worry, --disable-filter is working for now. I'm able to build the whole PPP, but not to install it. I will try a quick fix for installation, if that does not work I might disable this step, so your pull request should build. Later steps would be enabling filters support, then maybe installation to Buildroot target directory (but this is not really useful in this CI).

enaess: I can't remember but --prefix="path" is one, but I think there is a magic variable "DESTDIR". I know that is different than the previous INSTROOT="" 

Use Google, autotools are pretty standard. I don't remember from the top of my head if DESTDIR is the one. 

You'd do something similar to 

make install DESTDIR={PATH}

it would install it into DESTDIR+prefix

enaess: Did you have any success with this?

RICCIARDI-Adrien: Installation seems to work. However, I get problems with cross-compilation : the generated objects on MIPS64 build are x86_64 ones, I'm working on it.

RICCIARDI-Adrien: I'm quite close to something working (with filters support), but I'm hitting a PPP musl issue when linking pppd :
```
/home/ar/Documents/ppp/buildroot/output/host/lib/gcc/mips64-buildroot-linux-musl/9.3.0/../../../../mips64-buildroot-linux-musl/bin/ld: pppd-session.o: in function `session_start':
session.c:(.text+0xd4): undefined reference to `logwtmp'
/home/ar/Documents/ppp/buildroot/output/host/lib/gcc/mips64-buildroot-linux-musl/9.3.0/../../../../mips64-buildroot-linux-musl/bin/ld: session.c:(.text+0xe4): undefined reference to `logwtmp'
/home/ar/Documents/ppp/buildroot/output/host/lib/gcc/mips64-buildroot-linux-musl/9.3.0/../../../../mips64-buildroot-linux-musl/bin/ld: pppd-session.o: in function `session_end':
session.c:(.text+0x26c): undefined reference to `logwtmp'
/home/ar/Documents/ppp/buildroot/output/host/lib/gcc/mips64-buildroot-linux-musl/9.3.0/../../../../mips64-buildroot-linux-musl/bin/ld: session.c:(.text+0x278): undefined reference to `logwtmp'
collect2: error: ld returned 1 exit status
```
Previously there was a custom hack to detect whether `logwtmp` was present, is it possible to do such a thing with autotools ? https://github.com/ppp-project/ppp/blob/301230f8fde94b5b80e9dfba51816ede340e6b64/pppd/Makefile.linux#L86

enaess: Absolutely!

That's the entire point of Autotools. With the particular libc variant, do you know if logwtmp is present, or it needs to link with a different library? 

If this function is not present, there should be a function in sys-linux.c that is keyed off whether utmp.h exist on that system. It maybe exist but it may not provide logwtmp

enaess: Say, 

When I get to make these changes, what's the best way for me to test them? I don't see much in the .yaml file for script, etc...

RICCIARDI-Adrien: Musl does not provide `logwtmp` as it is deprecated, but some other functions from `libutil` are here, that's why this is a bit tricky. The best way to fix that would be to rewrite the needed PPP parts to use the correct functions, but this is out of the scope of this pull request.
You are right, a custom `logwtmp` is provided in `sys-linux.c`. I will try this evening if not defining `HAVE_LOGWTMP` is enough to make musl libc work.
I can give you the commands to run if you want to test. Maybe it's time to think to a way to make the Buildroot CI available to developers, through a script or a container.
Current working build for glibc and uClibc (with libpcap) is here : https://github.com/RICCIARDI-Adrien/ppp/tree/update_ci_for_autotools

paulusmack:
First, thank you @enaess for spending time on this.

My reaction is that doing this is probably necessary because we are getting a proliferation of options and variations -- inclusion/exclusion of bits of functionality, use (or not) of various libraries, differences between libraries (e.g. glibc vs. musl), etc. 

However, my reservations about autotools still stand, in particular the fact that it is really hard to debug problems (the scripts that autotools generates are very hard to follow). I would feel much happier about merging this if I could know that you @enaess (and others) intend to be around to fix any problems that arise.

I would also note that providing a lot of options is good from the point of view of people who want the particular combination they use to work easily, but also bad from the point of view of us trying to test the code, since we get an exponentially increasing number of combinations to test.

Neustradamus:
@paulusmack: Since a moment, you can see that some people contribute a lot in the PPP project.
For example, I think there will be not a problem from @enaess to solve if there are bugs in the future.

Neustradamus:
@enaess: You have done a little mistake: "2020-07-11" instead of "2021-07-11".

configure.ac
enaess: @RICCIARDI-Adrien - Could you give it another spin. I did improve the detection of -lutil / logwtmp, if it doesn't find it, it will use the built-in version instead. Also, fixed the --with-pcap instead of using --enable/disable-filter.

enaess:
@Neustradamus @paulusmack - 

Thank you for not shooting it down right away. I do carry the GitHub app on my phone, so I do get the messages rather quickly. Doubt this will be "BugFree (TM)", but since I have a particular interest in this project since sstp-client uses it, I'd be happy to help out. 

The pppd project has long had issues with permutations of a configuration not working, or needed some TLC. The EAP-SRP support was already broken (didn't compile) when I started adding the autotools support for configuring it. I suspect plenty of other bugs in e.g. libatm that was embedded here for the pppol2tp project. Customary, linking to the library if it exist would be preferable so you could get upstream changes (security fixes, etc). Others here (radius client)? included it's own version of config.h that was setup for a particular system. Man ... Anyone could go crazy trying to maintain this ...

At least for compiling and making sure it still link can be solved with using an appropriate CI build. Most open source projects rely on its users to report bugs, and I don't think this is any different unless you have $$ for a QA team..

This change would be most useful to package maintainers out there. You started this pre-2000 and what the project got going for it is its wide spread inclusion into almost every Linux distribution (and also a gazillion number of routers deployed out there). I remember adding pppd and pppoe support to WatchGuard's small office routers 10+ years ago. Then you have OpenWRT, Ubuntu, Redhat, Fedora, etc. that applies a bunch of patches to make this compile in their particular way. I have not done a full survey of patch to build system vs. feature, but it's a significant part of the entire patch set.

I'd strongly advocate for this change for that reason. There are plenty of capable people out there with some shell programming experience. The autotools / m4 with funny quotation, is just a variant. Anyone with background in Computer Science should be able to pick it up. It's just a bunch of conditional if, and maybe a for-loop or two.

configure.ac
RICCIARDI-Adrien: Build started with your latest changes, fingers crossed !

RICCIARDI-Adrien: Congrats @enaess, Buildroot CI seems to work like a charm with all libc and libpcap support. However, something went wrong on Solaris with libpcap support (libpcap-related autotools tests seem to be triggered on Solaris whereas the feature is not supported) : https://github.com/RICCIARDI-Adrien/ppp/runs/3088795322?check_suite_focus=true#step:3:2002

RICCIARDI-Adrien: You can fix the Solaris issue on this PR branch as Solaris CI has not changed.

enaess:
@paulusmack @carlsonj  - Looks like set_filter() is #ifdef'ed out on Solaris. It does properly compile if I enable it and add -lpcap on my Solaris 11.4 VM. What's the history here? Does the feature not work with the kernel or is there some other incompatibility?

enaess:
The other question I have here is whether e.g. the Solaris kernel module could be left in the source tree (github), but be left out of the pppd release? I mean, could pppd compile and on its own use the kernel module(s) from Solaris instead of compiling and installing a new module? If so, then the kernel module could be omitted from the pppd distribution archive (while still be present in the github source tree for the curious readers).

configure.ac
enaess: I did fix it, it will compile ... But the buildroot CI isn't updated on my end yet. How can I incorporate your changes here?

RICCIARDI-Adrien: I've rebased with your changes, build is starting. If it succeeds, I can create a PR for your branch `enaess:ppp-autotools`, so by merging my PR the commit should come to this PR, is that OK for you ?
Another question : are the `Makefile.linux` or `Makefile.sol2` (present for instance in `pppd` directory) still needed ? When CI is working and merged in your PR branch, maybe you can remove all traces of the legacy build system, so end maintainers can update their PPP packages in an easier way ?

enaess: Thank you and am looking forward to the PR. Yes, I will remove the legacy stuff in a separate change

SoapGentoo:
@paulusmack this would go a long way for us in Gentoo too, given how we provide users with the option of musl and glibc. Autotools scales better and provides for better hooks and it allows for better cross-compilation support. [Our current build script](https://gitweb.gentoo.org/repo/gentoo.git/tree/net-dialup/ppp/ppp-2.4.9-r4.ebuild#n35) is a complete mess, and this PR is going in the right direction. I and other people in Gentoo Base system team like @thesamesam are willing to help with any Autotools issues.

Makefile.am
```diff
@@ -0,0 +1,77 @@
...
+	(cd solaris ; make -f Makefile)
+
+clean-generic:
+	(cd solaris ; make -f Makefile clean)
+
+install-am:
+	(cd solaris ; make -f Makefile install)
+endif
+
+sampledir = $(sysconfdir)/@PACKAGE@
```
SoapGentoo, line 25:
```suggestion
sampledir = $(sysconfdir)/$(PACKAGE)
```

chat/Makefile.am
```diff
@@ -0,0 +1,9 @@
+
+sbin_PROGRAMS = chat
+man8_MANS = chat.8
+
+chat_SOURCES = chat.c
+chat_CFLAGS = -DTERMIOS -DSIGTYPE=void -UNO_SLEEP -DFNDELAY=O_NDELAY
```
SoapGentoo, line 6:
```suggestion
chat_CPPFLAGS = -DTERMIOS -DSIGTYPE=void -UNO_SLEEP -DFNDELAY=O_NDELAY
```
These are all macro directives

chat/Makefile.am
```diff
@@ -0,0 +1,9 @@
+
+sbin_PROGRAMS = chat
+man8_MANS = chat.8
```
SoapGentoo, line 3:
```suggestion
dist_man8_MANS = chat.8
```
and remove the `EXTRA_DIST`

contrib/pppgetpass/Makefile.am
```diff
@@ -0,0 +1,18 @@
...
+
+pppgetpass_vt_SOURCES = pppgetpass.vt.c
+pppgetpass_vt_CFLAGS = -Wno-unused-result
+
+if WITH_GTK
+noinst_PROGRAMS += pppgetpass.gtk
+
+pppgetpass_gtk_SOURCES = pppgetpass.gtk.c
+pppgetpass_gtk_CFLAGS  = -Wno-deprecated-declarations -Wno-discarded-qualifiers
+pppgetpass_gtk_CFLAGS += @GLIB_CFLAGS@ @GTK_CFLAGS@
```
SoapGentoo, line 12:
```suggestion
pppgetpass_gtk_CPPFLAGS += $(GLIB_CFLAGS) $(GTK_CFLAGS)
```

contrib/pppgetpass/Makefile.am
```diff
@@ -0,0 +1,18 @@
...
+pppgetpass_gtk_LDADD   = @GLIB_LIBS@ @GTK_LIBS@
```
SoapGentoo, line 13:
```suggestion
pppgetpass_gtk_LDADD   = $(GLIB_LIBS) $(GTK_LIBS)
```

contrib/pppgetpass/Makefile.am
```diff
@@ -0,0 +1,18 @@
+noinst_PROGRAMS = pppgetpass.vt
+man8_MANS = pppgetpass.8
```
SoapGentoo, line 2: `dist_` pattern

m4/ax_check_openssl.m4
```diff
@@ -0,0 +1,119 @@
...
+                AC_MSG_RESULT([no])
+                $2
+            ])
+        CPPFLAGS="$save_CPPFLAGS"
+        LDFLAGS="$save_LDFLAGS"
+        LIBS="$save_LIBS"
+
+        AC_SUBST([OPENSSL_INCLUDES])
+        AC_SUBST([OPENSSL_LIBS])
+        AC_SUBST([OPENSSL_LDFLAGS])
```
SoapGentoo, line 117: I would strongly consider using `pkg-config` over this macro

pppd/Makefile.am
```diff
@@ -0,0 +1,165 @@
+sbin_PROGRAMS = pppd
+man8_MANS = pppd.8 srp-entry.8
+if WITH_SRP
+sbin_PROGRAMS += srp-entry
+endif
+
+pkgconfigdir   = $(libdir)/pkgconfig
```
SoapGentoo, line 7: `pkg.m4` already defines this

pppd/Makefile.am
```diff
@@ -0,0 +1,165 @@
...
+    lcp.c \
+    magic.c \
+    main.c \
+    options.c \
+    session.c \
+    tty.c \
+    upap.c \
+    utils.c
+
+pppd_CFLAGS = -I${top_srcdir}/include -DPPPD_RUNTIME_DIR='"@PPPD_RUNTIME_DIR@"' -DPPPD_LOGFILE_DIR='"@PPPD_LOGFILE_DIR@"'
```
SoapGentoo, line 60:
```suggestion
pppd_CPPFLAGS = -I${top_srcdir}/include -DPPPD_RUNTIME_DIR='"@PPPD_RUNTIME_DIR@"' -DPPPD_LOGFILE_DIR='"@PPPD_LOGFILE_DIR@"'
```

pppd/Makefile.am
```diff
@@ -0,0 +1,165 @@
...
+pppd_LDFLAGS =
+pppd_LIBS =
+
+if LINUX
+pppd_SOURCES += sys-linux.c
+pppd_LIBS += @CRYPT_LIBS@ @UTIL_LIBS@
```
SoapGentoo, line 66:
```suggestion
pppd_LIBS += $(CRYPT_LIBS) $(UTIL_LIBS)
```

pppd/Makefile.am
```diff
@@ -0,0 +1,165 @@
...
+if WITH_IPXCP
+pppd_SOURCES += ipxcp.c
+endif
+
+if WITH_MPPE
+pppd_SOURCES += mppe.c
+endif
+
+if WITH_PCAP
+pppd_CFLAGS += @PCAP_CFLAGS@
```
SoapGentoo, line 96: parentheses

SoapGentoo (changes requested):

chat/Makefile.am
enaess: Fixed

Makefile.am
```diff
@@ -0,0 +1,77 @@
+ACLOCAL_AMFLAGS="-Im4"
+
+SUBDIRS = chat contrib pppd pppstats pppdump
```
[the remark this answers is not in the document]
enaess: That's a lot of work .. While it is a big pain, could we make it work like this for ppp? It's a small package and compilation takes about 2 minutes on the Ubuntu CI system (make && make check && make distcheck)

Makefile.am
```diff
@@ -0,0 +1,77 @@
...
+#   This is to work around the kernel module for PPP on Sun Solaris
+if SUNOS
+all-am:
+	(cd solaris ; make -f Makefile)
+
+clean-generic:
+	(cd solaris ; make -f Makefile clean)
+
+install-am:
+	(cd solaris ; make -f Makefile install)
```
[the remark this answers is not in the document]
enaess: Fixed

Makefile.am
enaess: Fixed

chat/Makefile.am
enaess: Fixed, and revisited the other places this was used.

contrib/pppgetpass/Makefile.am
enaess: Fixed

Makefile.am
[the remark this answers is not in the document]
SoapGentoo: For the first cut we can leave it, but even a 20s -> 10s build time improvement can mean a lot to me when rebuilding 1000 packages on my system

m4/ax_check_openssl.m4
enaess: This could be possibly future work? Systems like Solaris doesn't have pkgconfig, so there are some additional logic that would be needed to override this. Also, libpcap, supplies a pkgconfig file.

pppd/Makefile.am
enaess: I tried to remove this, and autoreconf barfed at me.

pppd/Makefile.am
enaess: Fixed

pppd/Makefile.am
enaess: Fixed

configure.ac
enaess: @RICCIARDI-Adrien I did merge your change, thank you so much for all your help here.

pppd/Makefile.am
enaess: Fixed

contrib/pppgetpass/Makefile.am
enaess: Fixed

contrib/pppgetpass/Makefile.am
enaess: Fixed

enaess:
@paulusmack Looks like the buildroot script was fixed by @RICCIARDI-Adrien, @SoapGentoo suggested a bunch of reviews, vouched for the change, and even volunteered to help out maintaining autotools for pppd. @yarda maintains pppd for Fedora (IIRC) and thumbed up the change.

The users of this change is mostly maintainers of distributions, and by the looks of it, you'd probably will hear quickly if something isn't working correctly (e.g. library is incorrectly linked, missing cflags, etc). These are users of the build system and also much likely to report bugs on this than any other common non-geek user of a dial up modem.

I've also with care compared the dist .tar.gz archive against 2.4.9, and it compares. I'd like to get back and add PEAP and then possibly fixup system header files (vs. included headerfiles) much similar to the conversation of ppp_defs.h missing on some systems. That would be an excellent candidate for this new build system.

@paulusmack and @carlsonj  - what do you say?

carlsonj:
On 7/16/21 8:14 PM, Eivind Næss wrote:
> The other question I have here is whether e.g. the Solaris kernel module 
> could be left in the source tree (github), but be left out of the pppd 
> release? I mean, could pppd compile and on its own use the kernel 
> module(s) from Solaris instead of compiling and installing a new module? 
> If so, then the kernel module could be omitted from the pppd 
> distribution archive (while still be present in the github source tree 
> for the curious readers).

The interfaces provided by the pppd kernel modules are (in Solaris 
terms) architecturally classified as "Private."  In practical terms, it 
means that no documentation is provided, nor is any guarantee of any 
sort of stability over time preserved -- including across patches.  The 
Solaris maintainers would, I expect, try not to break independent 
distributions intentionally, but it almost certainly wouldn't be a high 
priority to fix them if they did become broken after depending on 
undocumented interfaces.  (Mere existence of a header file in the 
Solaris world does *not* mean that the interfaces are intended to be 
used or relied upon.)

So, it's not technically correct for anything but the 
Solaris-distributed code to depend on those modules, and they were 
renamed in order to avoid any possible conflict (if one were to install 
both this pppd from source and the binaries from the Solaris 
distribution).  You could probably make it work, though.

If these kernel module sources need to be removed from the release, then 
it would probably be a good idea to provide a pointer where users can go 
get them.  (That is assuming that the goal is not just removal of the 
Solaris support.  If that's the goal, then removing it all rather than 
breaking it might be a better solution.)

-- 
James Carlson         42.703N 71.076W         ***@***.***>

carlsonj:
On 7/17/21 4:45 PM, Eivind Næss wrote:
> +        AC_SUBST([OPENSSL_INCLUDES])
> +        AC_SUBST([OPENSSL_LIBS])
> +        AC_SUBST([OPENSSL_LDFLAGS])
> 
> This could be possibly future work? Systems like Solaris doesn't have 
> pkgconfig, so there are some additional logic that would be needed to 
> override this. Also, libpcap, supplies a pkgconfig file.

Modern Solaris (>= 10) does include pkg-config.  The only things 
normally included in the database are open source packages included with 
the system, but that might be sufficient here.

Older Solaris (9 and before) doesn't include it.  So, whether you can 
rely on it depends on what versions you're targeting.  (The right answer 
might be to test for its existence first if you need it.)

https://docs.oracle.com/cd/E37838_01/html/E66175/gplhi.html

-- 
James Carlson         42.703N 71.076W         ***@***.***>

carlsonj:
On 7/16/21 4:56 PM, Eivind Næss wrote:
> @paulusmack <https://github.com/paulusmack> @carlsonj 
> <https://github.com/carlsonj> - Looks like set_filter() is #ifdef'ed out 
> on Solaris. It does properly compile if I enable it and add -lpcap on my 
> Solaris 11.4 VM. What's the history here? Does the feature not work with 
> the kernel or is there some other incompatibility?

Wow, that's a blast from the past.  That code was commented out over 21 
years ago by Adi Masputra, early in the Vector PPP project and around 
the time that I started working with him at Sun.  The fact that it 
compiles on modern Solaris is an interesting testament to the stability 
of that platform.  ;-}

The feature defines two BPF filters used for identifying kinds of 
traffic in an on-demand link scenario.  One filter identifies traffic 
that is allowed to pass on the link.  The second filter identifies 
traffic that is sufficient to reset the idle timer and thus keep the 
link active if it's already up.

So, for instance, you could define (say) email on TCP port 25 to be 
important enough to bring up the link and keep it up, but (say) ICMP 
Echo to be allowed only if the link is already up and not to count as 
"activity."  The interface itself is a bit on the sketchy side, and the 
use-cases for it are a bit contrived.  If there are on-demand links in 
use, I'd expect that people use ipfilter instead to protect the 
interface from unwanted traffic.

My recollection is that it's based on a similar feature from the old 
proprietary add-on "Sun PPP" from the long-gone ISDN group, which the 
"Vector" project replaced with regular pppd.  I'd be really surprised if 
anyone ever really used it outside of our own internal testing.

I looked in the last set of Open Solaris sources that I cached, and I 
can find no trace of these ioctls in the kernel besides the definitions 
in the header files.  I could be mistaken about this, but I don't 
believe there was ever a release to customers that included these 
features.  (In general, Sun *never* removed features.  So, if we'd ever 
have shipped it, we would have had to keep it.  And it's not there now.) 
  The set_filter() function in the Sun-modified pppd source is 
conditional on this symbol:

#if defined(FILTER_PACKETS)

... and nothing ever defines it.

-- 
James Carlson         42.703N 71.076W         ***@***.***>

enaess:
> On 7/17/21 4:45 PM, Eivind Næss wrote: + AC_SUBST([OPENSSL_INCLUDES]) + AC_SUBST([OPENSSL_LIBS]) + AC_SUBST([OPENSSL_LDFLAGS]) This could be possibly future work? Systems like Solaris doesn't have pkgconfig, so there are some additional logic that would be needed to override this. Also, libpcap, supplies a pkgconfig file.
> Modern Solaris (>= 10) does include pkg-config. The only things normally included in the database are open source packages included with the system, but that might be sufficient here. Older Solaris (9 and before) doesn't include it. So, whether you can rely on it depends on what versions you're targeting. (The right answer might be to test for its existence first if you need it.) https://docs.oracle.com/cd/E37838_01/html/E66175/gplhi.html
> […](#)
> -- James Carlson 42.703N 71.076W ***@***.***>

Well, okay ... maybe just the version of autoconf doesn't provide pkg.m4 macros that will detect it. I've added that to the m4 directory so it kind of works for configure's part. We could update this patch to search for it for respective packages that is known to provide it (e.g. openssl and pcap).

enaess:
Gah, Github's commentary / reply function suck ...

@carlsonj - Regarding the ppp filter stuff, I did change the ifdef from 0 to if PPP_FILTER and it is controlled by configure -> config.h, this would be most consistent with what features are present on Linux. However, as you said ... this largely wouldn't work, and likely the function would just error out with a negative return code on any of the ioctls() in there. I could leave it as is, and change configure.ac to error out stating that the feature isn't supported on SunOS and leave a big comment in there for curious readers. Right now, the feature is tied into the presence of libpcap on the system (in which by default SunOS already have installed). 

The goal of this change would not be to eliminate any features that was already present in the shipping 2.4.9 version of pppd. It currently ships with the kernel module for SunOS, and future versions will too. It's just that the build using autotools will be used to build pppd (even on SunOS), and there is the "extra" part of the distribution that people on SunOS would be interested in. Right now, it's wired into the Automake stuff with a separate $(MAKE) -f .. in the Solaris folder. I'd say it should suffice for now. 

And apparently to your response, this is strictly required to be able to have pppd working on SunOS. The alternative would be to 1) Build it as a part of the make process, 2) keep it as a separate build step for SunOS folks, 3) eliminate it from the distribution archive while still keep it in Git repository (this is if current kernel modules would work).

Right now, the choice I made was to keep it building alongside with Automake and it seems to be working fine.

@carlsonj - Do you have any suggestions to this change as you review it, or inclinations?

carlsonj:
On 7/18/21 3:07 PM, Eivind Næss wrote:
> Gah, Github's commentary / reply function suck ...
> 
> @carlsonj <https://github.com/carlsonj> - Regarding the ppp filter 
> stuff, I did change the ifdef from 0 to if PPP_FILTER and it is 
> controlled by configure -> config.h, this would be most consistent with 
> what features are present on Linux. However, as you said ... this 
> largely wouldn't work, and likely the function would just error out with 
> a negative return code on any of the ioctls() in there. I could leave it 
> as is, and change configure.ac to error out stating that the feature 
> isn't supported on SunOS and leave a big comment in there for curious 
> readers. Right now, the feature is tied into the presence of libpcap on 
> the system (in which by default SunOS already have installed).

If I were making the change, I'd just delete the set_filter() code. 
Anything that's been under "#if 0" for over 20 years is something that 
(A) nobody needs, (B) is incredibly stale, and (C) doesn't actually 
work.  If anyone ever needed it again, that's what version control 
systems are for.  The current source shouldn't be a museum.

> The goal of this change would not be to eliminate any features that was 
> already present in the shipping 2.4.9 version of pppd. It currently 
> ships with the kernel module for SunOS, and future versions will too. 
> It's just that the build using autotools will be used to build pppd 
> (even on SunOS), and there is the "extra" part of the distribution that 
> people on SunOS would be interested in. Right now, it's wired into the 
> Automake stuff with a separate $(MAKE) -f .. in the Solaris folder. I'd 
> say it should suffice for now.

Ah, ok.  I always sort of appreciated the fact that pppd had its own 
lightweight configuration logic rather than casting in its lot with the 
GNU autoconf ... stuff.

I don't think anyone on Solaris would mind in the least if a separate 
step were required to do the build and/or installation, as long as it 
was documented somewhere.

> And apparently to your response, this is strictly required to be able to 
> have pppd working on SunOS. The alternative would be to 1) Build it as a 
> part of the make process, 2) keep it as a separate build step for SunOS 
> folks, 3) eliminate it from the distribution archive while still keep it 
> in Git repository (this is if current kernel modules would work).
> 
> Right now, the choice I made was to keep it building alongside with 
> Automake and it seems to be working fine.
> 
> @carlsonj <https://github.com/carlsonj> - Do you have any suggestions to 
> this change as you review it, or inclinations?

Keeping it building as part of the newfangled automake stuff would be 
the best solution, I think.

-- 
James Carlson         42.703N 71.076W         ***@***.***>

enaess:
@carlsonj 

If I understand it correctly, the SunOS kernel module provided by SunOS supports this ioctl
'''
/usr/include/net/pppio.h:#define        PPPIO_ACTIVEFILT _PPPIO(149)    /* set filter for "link active" pkts */
'''
But the one included in pppd's kernel module does not (the ./include directory does provide the define), but it's not handled in the kernel module. Is my understanding correct?

carlsonj:
On 7/19/21 1:30 PM, Eivind Næss wrote:
> @carlsonj <https://github.com/carlsonj>
> 
> If I understand it correctly, the SunOS kernel module provided by SunOS 
> supports this ioctl

It does not support it.  Header files are not documentation on Solaris. 
  If you actually issue the ioctl on any Solaris system, you'll find you 
just get back an error instead.

> '''
> /usr/include/net/pppio.h:#define PPPIO_ACTIVEFILT _PPPIO(149) /* set 
> filter for "link active" pkts */
> '''
> But the one included in pppd's kernel module does not (the ./include 
> directory does provide the define), but it's not handled in the kernel 
> module. Is my understanding correct?

No.  There are no kernel modules that support this ioctl.  It's a 
complete red herring.

As I mentioned earlier, on Solaris the header files are NOT considered 
any sort of documentation nor any promise of architectural stability. 
The fact that the ioctl value (just an integer) is declared there means 
nothing about actual kernel support.  If it's not in the man pages on 
Solaris, then it either does not exist or has a stability level that 
does not ordinarily permit projects outside of the Solaris source tree 
to depend on them.

In this case it just does not exist.

Here's the last Open Solaris snapshot I have from when I left Sun in 2009:

% find . -type f | xargs fgrep PPPIO_ACTIVEFILT
./usr/src/uts/common/net/pppio.h:#define	PPPIO_ACTIVEFILT _PPPIO(149)	/* 
set filter for "link active" pkts */
./usr/src/cmd/cmd-inet/usr.bin/pppd/sys-solaris.c:		if (strioctl(pppfd, 
PPPIO_ACTIVEFILT, active,
%

None of the snapshots I have access to include it, and the fact that the 
entire project team was reassigned at the end of the Vector PPP project 
means that it's exceedingly unlikely that anyone reached back and added 
or removed something complex like this.

For contrast, here's one that *is* supported:

% find . -type f | xargs fgrep PPPIO_XFCS
./usr/src/uts/common/io/ppp/sppptun/sppptun.c:	case PPPIO_XFCS:
./usr/src/uts/common/io/ppp/spppasyn/spppasyn.c:	case PPPIO_XFCS:
./usr/src/uts/common/io/ppp/spppasyn/spppasyn.c:		case PPPIO_XFCS:
./usr/src/uts/common/net/pppio.h:#define	PPPIO_XFCS	_PPPIO(152)	/* set 
transmit FCS type */
./usr/src/uts/common/net/pppio.h: * Values for PPPIO_XFCS/PPPIO_RFCS. 
(Note that we don't support
./usr/src/cmd/cmd-inet/usr.bin/pppd/sys-solaris.c:	if (strioctl(pppfd, 
PPPIO_XFCS, &fcs, sizeof (fcs), 0) < 0) {
%

Note, in particular, the "case" statements that handle the ioctl in the 
sppptun STREAMS driver and spppasyn STREAMS module.

You can find similar information on-line, if you wish, in the more 
modern IllumOS code, which was based on Open Solaris.  Here's a pointer 
to the source code for the PPP kernel modules:

https://github.com/illumos/illumos-gate/tree/master/usr/src/uts/common/io/ppp

Note the fact that NO kernel modules ("uts") actually use the value. 
It's not supported.  It doesn't work if you try it.  Adi removed the 
set_filter thing over two decades ago because we weren't going to use 
it, and he probably should have just deleted the code rather than used 
"#if 0" in order to avoid just this sort of confusion.  If I run into 
him again someday, I'll mention it.  :-/

(I was chair of the Sun Platform Software Architecture Committee for 
several years, so if you really want a run-down on how all the 
architectural bits work, I'd be happy to provide it.  But for a really 
simple answer: no man page == no support.)

(I was also part of the Solaris Vector PPP project, so, although some 
years have passed, I do recall some useful things about it, including 
some things that may surpass what can be learned by just typing "make.")

-- 
James Carlson         42.703N 71.076W         ***@***.***>

enaess:
Thank you so much for that long and elaborate response. My experience with SunOS is very minimal so I appreciate the explanation. 

I'll update the change with your suggestions later today.

enaess:
As am saying this, my VM running solaris crashed. Will likely fix it tomorrow morning.

SoapGentoo:
@enaess please rebase branches not not merge from master, it keeps the history easier to read

enaess:
@carlsonj - I've incorporated your feedback. @paulusmack this change is ready to be merged. For everyone who was involved in reviewing, and helping out with CI scripts, etc, thank you very much!

Neustradamus:
@enaess: Good job!

My comments:

This file with only the date is needed?
- https://github.com/ppp-project/ppp/blob/75f1f5fde9fa6a64d0c35af6dfac5086a0739334/pppd/patchlevel.h

It is not possible to have a better place for VERSION?
- https://github.com/ppp-project/ppp/blob/75f1f5fde9fa6a64d0c35af6dfac5086a0739334/configure.ac

And master branch before to create a release must have a VERSION-dev

enaess:
@Neustradamus 

I couldn't care less for including patchlevel.h. This is already provided by pppdconf.h which is supposed to be defining the compile-time-flags pppd was configured with such that if sstp-client compiles against MPPE, the mppe,h with proper definitions would compile as pppd was configured. I don't know the previous history of patchlevel and the DATE macro, but if it was up to me; I'd remove it. However, I'd tried really hard to keep things in its original format to avoid breaking any third party apps.

I don't understand your comment about VERSION and VERSION-dev. The VERSION is set by configure.ac in the very top of the file. It's already set in config.h and also set in pppdconf.h. The pppdconf.h is modeled similarly to openssl's opensslconf.h 

Another follow up step after this change is merged is to cleanup the public headers. This change should handle the Make / Configure and not change much of the existing code.

Neustradamus:
@enaess: It is bad to have a file with the date only.

It is not possible to have a file with the VERSION easy to update?

Currently when this PR will be merged, you have added the 2.4.10, it is not the time to have a 2.4.10 version.
It must be 2.4.10-dev.

enaess:
@Neustradamus - I am not sure what you mean by this. Do you insist on having patchlevel.h include the VERSION define? 

Traditionally, in Autotools world, this is defined in a central place of configure.ac with the AC_INIT() macro. This will propagate to config.h (and also pppdconf.h as config.h isn't distributed in /usr/include/ppp directory). 

There are several ways to manage the VERSION-dev. One way the Linux kernel has done this is to use odd-version numbers like 2.4.11 as the unstable, and 2.4.12 as the stable. ppp-project hasn't actually used this scheme before so that would be new. However, it also haven't used the -dev suffix either. So I am not sure. 

Setting version via AC_INIT() instead of patchlevel.h is new to the ppp-project. I'd say the convention is more aligned with the process related to autotools. We could of course, keep it in patchlevel,h, but that would require us to create a pppd/patchlevel.h.in that uses `#define VERSION @VERSION@` style macro. 

However, as I've said previously. These headers have implicit dependencies on each other and some are missing include guards. Others may not need to be included in the /usr/include/ppp directory at all., So, there needs to be some cleanup before the next release of ppp is created.

enaess:
@yarda @SoapGentoo - Would you two be able to do a dry run and test e.g. ppp over ssh (poor man's VPN) on Fedora and Gentoo?

You'll have to specify the correct logfile, runtime, and plugin directories per your distribution. I've also seen that some may need to override the sysconfdir and localstatedir variables too as the default is keyed off the prefix.

What am hoping for here is if either one of you two would spot any bugs in working with the new build system, I could update the PR

Though, this shouldn't be needed to merge the request...

Neustradamus:
@enaess: The question is:
- To keep or remove "patchlevel.h".
- To create a VERSION file with the version value or use the current file but with VERSION-dev which it will be change to 4.2.10 at commit and tag release.

PPP 2.4.9 is unstable? No.

Please look: https://semver.org

enaess:
@Neustradamus

I am by no means married to patchlevel.h and can go either way. It's been released in the past so it could be prudent to keep it even if it was empty 

I can update the version in configure.ac to say 2.4.9, no problem. We can change it when 2.4.10 is being released.

enaess:
@Neustradamus 

Also, VERSION is defined in two places with this change. 
- pppd/config.h
- pppd/pppdconf.h

The config.h is a private header for pppd package created by configure and meant to be included as the first include in each c file such that one can go

```
#ifdef HAVE_CRYPT_H
#include crypt.h
#endif
```

The pppdconf.h is also generated by configure, but is meant as the public header for pppd (much in similar with opensslconf.h). This file too defines VERSION

Ideally, it should define PPP_VERSION so it won't overlap with other packages defines, but oh well ...

Neustradamus:
@enaess: No, I have requested:
`2.4.10-dev`

enaess:
@Neustradamus - I changed the version in configure.ac to 2.4.9. @paulusmack will have to change it to 2.4.10 when release + tag happens. 

@paulusmack - what's the chance of getting this change merged?

enaess:
@Neustradamus 2.4.10-dev - done

Neustradamus:
@enaess: Thanks!

When the 2.4.10 will be official, the commit and tag of the release must have:
- Update of the version
- CHANGELOG

After the official release build, a new commit to change 2.4.11-dev.

Now "pppd/patchlevel.h" can be removed, no?
Date is not important here.

Note: I prefer a file like "/VERSION" or "pppd/VERSION" with only the version to easy update.

enaess:
@Neustradamus 

I believe the release process you outlined would be correct. I'll leave the patchlevel.h subject to input from @paulusmack, but I believe it can be removed (it's included via pppd.h, and that is a change we own). Hopefully, nobody directly includes this file.  I don't fully know the history of the DATE define, but it certainly could be automatically added via configure.ac and generation of pppdconf.h. I do hope it can be removed though ...

The ChangeLog update is a cumbersome, but I suppose we could find a script to automatically dump the git history between 2.4.9 and 2.4.10 and create the correctly formatted addition to this file. 

Autotools never complained of VERSION or generated it, not sure if that's a part of the GNU standard, is it? It does generate and install a bunch of new files (e.g. INSTALL), and require AUTHORS and a few others.

Makefile.am
```diff
@@ -0,0 +1,77 @@
...
+if SUNOS
+all-am:
+	(cd solaris ; $(MAKE) -f Makefile)
+
+clean-generic:
+	(cd solaris ; $(MAKE) -f Makefile clean)
+
+install-am:
+	(cd solaris ; $(MAKE) -f Makefile install)
+endif
```
SoapGentoo, line 23: btw, any reason not to just do
```
if SUNOS
SUBDIRS += solaris
endif
```
?

enaess: I've tried this, and it somehow require a Automake file in that directory.

SoapGentoo: oh sure, but can't you replace the `Makefile` in there too?

enaess: The Makefile in that directory is non-automake based and given that it builds a kernel module for SunOS, it's incompatible with automake..

You are welcome to sync out my fork and try to change it yourself.

enaess: This is really just for the SunOS folks that is interested in building the kernel module (and the previous build system already had that wired in), It could be decoupled from the autoconf stuff and documented in a read me file somewhere on how to build that separately. The ideal thing is to just support it as it used to and compile it inline with the automake stuff.

enaess: Now, ideally .. if you sync out my fork and give this a spin. I hope the initial hurdle of setting up the configure process and setting the right paths help shake out a few bugs. You'd be also able to make additional suggestions where you see necessary.

yarda:
> @yarda @SoapGentoo - Would you two be able to do a dry run and test e.g. ppp over ssh (poor man's VPN) on Fedora and Gentoo?
> 
> You'll have to specify the correct logfile, runtime, and plugin directories per your distribution. I've also seen that some may need to override the sysconfdir and localstatedir variables too as the default is keyed off the prefix.
> 
> What am hoping for here is if either one of you two would spot any bugs in working with the new build system, I could update the PR
> 
> Though, this shouldn't be needed to merge the request...

We have some very simple downstream smoke tests utilizing beakerlib. It will run on the next release / release candidate when packaged downstream.

enaess:
@paulusmack 

What's the chance of having this patch-set merged before the code diverges even further (w.r.t. other PRs)? It's received a fair share of review / feedback, and help from the community. They seem to for the most part to love the change.

Most distributions would probably have a similar ./configure command line to this dependent on their distribution settings:

```
./configure --prefix=/usr --localstatedir=/var --sysconfdir=/etc --disable-static --with-plugin-dir=/usr/lib/ppp/2.4.10 --with-runtime-dir=/var/run/ppp --with-logfile-dir=/var/log/ppp --enable-ipxcp --enable-multilink 
```

I did compile the sstp-client project and did some testing with that and it ran without any problems. Also, doing a "make install" compares to that of the 2.4.9 archive (with an extra extra man page for ppp-discovery, and a few header files that may have been forgotten). Libtool adds extra .la files as a part of the distribution. Package maintainers usually omit these files if not necessary. Other than that, it is as identical as possible.

Noted that DES encryption has been absent from -lcrypt for a few releases. Thus compiling pppd without openssl's DES support isn't a viable combination. Future work should possibly add a des.c file that provides those functions in case some of the embedded guys desire to run it without a crypto library (or provide a version of libcrypt).

Neustradamus:
Thanks @enaess (and all other contributors), it is now merged by @paulusmack!

Neustradamus:
@paulusmack: patchlevel.h can be removed no?

configure.ac
```diff
@@ -0,0 +1,417 @@
...
+
+#
+# Disable support for IPX control protocol
+AC_ARG_ENABLE([ipxcp],
+    AS_HELP_STRING([--enable-ipxcp], [Enable IPX Control Protocol support]))
+AM_CONDITIONAL(WITH_IPXCP, test "x${enable_ipxcp}" = "xyes")
+AM_COND_IF(WITH_IPXCP,
+    AC_DEFINE([IPX_CHANGE], 1, ["Have IPX Control Protocol"]))
+AS_IF([test "x${build_sunos}" = "xyes" && test "x${enable_ipxcp}" = "xyes"],
+    [AC_MSG_ERROR([IPXCP is not supported on SunOS, disable using --disable-ipxcp])])
```
pali, line 125: Why you named IPX protocol support `--enable-ipxcp` with define `IPX_CHANGE`...

configure.ac
```diff
@@ -0,0 +1,417 @@
...
+AS_IF([test "x${enable_mslanman}" = "xyes" && test "x${enable_microsoft_extensions}" != "xno"],
+    AC_DEFINE([MSLANMAN], 1, ["Have Microsoft LAN Manager support"]))
+
+#
+# Disable IPv6 support
+AC_ARG_ENABLE([ipv6-support],
+    AS_HELP_STRING([--disable-ipv6-support], [Disable IPv6 support]))
+AM_CONDITIONAL(WITH_INET6, test "x${enable_ipv6_support}" != "xno")
+AM_COND_IF([WITH_INET6],
+    AC_DEFINE(INET6, 1, ["Have IPv6 support"]))
```
pali, line 160: ... and IPv6 protocol support named `--enable-ipv6-support` with define `INET6`?

It looks like a big inconsistency.

In both cases you are enabling / disabling PPP control protocol (IPXCP and IPV6CP) and not support for particular protocol.

Technically speaking you can use IPX or IPv6 protocols over PPP also without corresponding control protocol in pppd. Just you need to handle stuff by your own way...

pali (commented):
Just one question:

configure.ac
enaess: @pali I was trying to be consistent with how these was defined in the previous configure scripts. The INET6 define does in fact disable IPv6 support entirely (and not just ipv6cp); The IPXCP is actually the IPXCP protocol. Grep for INET6 in the code ...

pali: > I was trying to be consistent with how these was defined in the previous configure scripts. 

Ok. (I was in impression that conversion to autotools means also cleanup of this stuff)

> The INET6 define does in fact disable IPv6 support entirely (and not just ipv6cp);

Apparently not. The only thing which pppd do with IPv6 is its Control Protocol. So _IPv6 entirely_ means IPV6CP protocol... Just code related to IPV6CP is not only in ipv6cp.c file but also in others...

enaess: @pali You know that area of the code better than me. Am I understanding you correctly when you say that INET6 define actually enable/disables the IPV6CP? If so, then I do agree with you that the configure option for consistency would be better as --enable/disable-ipv6cp to match --enable-ipxcp.

On a slightly different note, all the #ifdef INET6 does have a risk of bleeding over to other projects as it is used in pppd's public headers as well. I'd rather change that to something along the lines of WITH_INET6, or WITH_PPP_IPV6CP. The latter would make it crystal clear that the feature belongs to pppd and not overlap with a third party's INET6 if they would define it for whatever reason. The same goes with other defines, e.g. CHAPMS, etc. 

While I was trying to avoid changing the defines because it is very intrusive to the source code. #296 was meant to mostly refactor the Makefiles / Build system while leaving the source code only with cosmetic fixes to it. The cleaner the fix is, the more likely it would be accepted and merged.

I'd be totally fine with changing the configure script to take AC_ARG_ENABLE([ipv6cp], ....) to match that of the of ipxcp option. What do you think? I can craft another fix for this...

enaess: Also, before I head down that road, a thumbs up from @paulusmack would be helpful.

pali: I would say that there is something terribly wrong if there is a need to explicitly disable IPv6 at compile time in year 2021. Any kind of IPv6 support should not be disabled at compile time.

pppd for a long time can detect ppp link can do IPV6CP and disable its negotiation when other part does not support it. And in case other side is totally broken it is possible to disable IPV6CP at runtime by pppd option.

On the other hand, disabling IPX makes sense as there is probably nobody in year 2021 who uses IPX (and really nobody who has it over ppp).

enaess: @pali that is probably true. I think there might be a use-case for custom embedded systems that doesn't have e.g. IPv6 compiled into the kernel. Do you suggest removing the define altogether? If so, I'd want to see @paulusmack input ...

enaess:
@Neustradamus - Could you ping e.g. the Debian developer (Chris Boot) and check with him if current tip of pppd/master works for him? Ultimately if the next release will become successful is if these distribution maintainers doesn't get pissed off because we changed everything on them.

bootc:
@enaess / @Neustradamus - I've been watching these developments with interest, actually, and I find it encouraging that the autotools are being used for ppp. I haven't yet looked at the code or tried this out at all, to be honest, but I expect it to reduce the complexity of the packaging for sure. I likely won't have time to work on the Debian packaging in earnest for a little while, though.

enaess:
@bootc Cool! If you have any suggestions or outstanding patches to the make system, or if things that are currently difficult w.r.t. packaging. One thing to review is the use of " -module -avoid-version" for the LDFLAGS of the modules. Others, there are a --with-atm=DIR that allows one to link with the libatm1, etc.

PR #302, opened 2021-07-25
Rebase and Update the PEAP authentication change from Rustam Kovhaev

enaess:
This pull request rebases @rustylife previous PR with the intention of merging it with the existing functionality of CA verification, CRL verification, Certificate Checks, etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ autom4te.cache
 /configure.scan
 /depcomp
 /install-sh
+/test-driver
 /missing
 /INSTALL
 /pppd/stamp-h1
@@ -49,6 +50,10 @@ m4/lt~obsolete.m4
 # Generated Makefile
 Makefile
 Makefile.in
+
+# Test related stuff
+utest_*
+test-suite.log
 
 # Distfiles
 /ppp-*.tar.gz
--- a/configure.ac
+++ b/configure.ac
@@ -189,7 +189,15 @@ AC_ARG_ENABLE([eaptls],
     AS_HELP_STRING([--disable-eaptls], [Disable EAP-TLS authentication support]))
 AS_IF([test "x$enable_eaptls" != "xno"],
     AC_DEFINE([USE_EAPTLS], 1, ["Have EAP-TLS authentication support"]))
-AM_CONDITIONAL(WITH_EAPTLS, test "${enable_eaptls}" != "no")
+AM_CONDITIONAL(WITH_EAPTLS, test "x${enable_eaptls}" != "xno")
+
+#
+# Disable PEAP support
+AC_ARG_ENABLE([peap],
+    AS_HELP_STRING([--disable-peap], [Disable PEAP authentication support]))
+AS_IF([test "x${enable_peap}" != "xno"],
+    AC_DEFINE([USE_PEAP], 1, ["Have PEAP authentication support"]))
+AM_CONDITIONAL([WITH_PEAP], test "x${enable_peap}" != "xno")
 
 #
 # Disable OpenSSL engine support
@@ -232,14 +240,14 @@ AM_CONDITIONAL(WITH_OPENSSL, test "${with_openssl}" != "no")
 
 #
 # Check if OpenSSL has compiled in support for various ciphers
-AS_IF([test "${with_openssl}" != "no" ], [
+AS_IF([test "x${with_openssl}" != "xno" ], [
     AX_CHECK_OPENSSL_DEFINE([OPENSSL_NO_MD4], [md4])
     AX_CHECK_OPENSSL_DEFINE([OPENSSL_NO_MD5], [md5])
     AX_CHECK_OPENSSL_DEFINE([OPENSSL_NO_DES], [des])
     AX_CHECK_OPENSSL_DEFINE([OPENSSL_NO_SHA], [sha])
 ], [
-    AS_IF([test "x$enable_eaptls" != "xno"],
-        [AC_MSG_ERROR([OpenSSL not found, and if this is your intention then run configure --disable-eaptls])])
+    AS_IF([test "x${enable_eaptls}" != "xno" || test "x${enable_peap}" != "xno"],
+        [AC_MSG_ERROR([OpenSSL not found, and if this is your intention then run configure --disable-eaptls and --disable-peap])])
 ])
 
 AM_CONDITIONAL([OPENSSL_HAVE_MD4], test "x${ac_cv_openssl_md4}" != "xno")
@@ -414,4 +422,5 @@ Features enabled
     CBCP.................: ${enable_cbcp:-no}
     IPXCP................: ${enable_ipxcp:-no}
     EAP-TLS..............: ${enable_eaptls:-yes}
+    PEAP.................: ${enable_peap:-yes}
 "
--- a/pppd/Makefile.am
+++ b/pppd/Makefile.am
@@ -28,11 +28,12 @@ pppd_include_HEADERS = \
     md4.h \
     md5.h \
     mppe.h \
-    pppdconf.h \
     patchlevel.h \
     pathnames.h \
+    peap.h \
     pppcrypt.h \
     pppd.h \
+    pppdconf.h \
     session.h \
     sha1.h \
     spinlock.h \
@@ -125,6 +126,10 @@ endif
 
 if WITH_EAPTLS
 pppd_SOURCES += eap-tls.c
+endif
+
+if WITH_PEAP
+pppd_SOURCES += peap.c
 endif
 
 if !WITH_OPENSSL
--- a/pppd/Makefile.am
+++ b/pppd/Makefile.am
@@ -1,5 +1,14 @@
 sbin_PROGRAMS = pppd
 dist_man8_MANS = pppd.8
+check_PROGRAMS = \
+    utest_chap
+
+utest_chap_SOURCES = chap_ms.c pppcrypt.c utils.c
+utest_chap_CPPFLAGS = -DUNIT_TEST
+utest_chap_LDFLAGS =
+
+TESTS = $(check_PROGRAMS)
+
 if WITH_SRP
 sbin_PROGRAMS += srp-entry
 dist_man8_MANS += srp-entry.8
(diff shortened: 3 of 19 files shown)


enaess:
@rustylife could you please have a check on this Pull Request? I've gotten it compiling, but have a slight issue getting it to connect. Maybe my has an error in it, but I can't seem to trigger a PEAP-MSCHAPv2 to occur.

enaess:
@rustylife - also something I find annoying with this change is that there is no method to free the memory allocated by the allocate_buffers() function. While am sure you can make the argument that it won't matter because the memory is alive during the lifetime of the client session and it won't grow, it's just plain bad not to handle it.

rustylife:
i get SIGSEGV because tls_set_ca() tries to access null ptr:
0x000055fc3fc491bf in tls_set_ca (ctx=0x55fc408d5ac0, ca_dir=0x0, ca_file=0x0) at tls.c:364
364             ca_file = NULL;

peap peer/client config file does not really have to have any ca/capath options defined, it should just work with openssl defaults

enaess:
@rustylife - Could you for now use the CA used to sign the server certificate to bypass the problem. I'll look into it, but generally it should follow the EAP-TLS code w.r.t. CA certificates for consistency sake.

enaess:
@rustylife - I fixed up that CA file/dir crash you were having. Please give it another spin?

Neustradamus:
@jjkeijser: What do you think?

Neustradamus:
@paulusmack, @jjkeijser: What do you think?

enaess:
There is three things with this change
- LCP need to consider PEAPv0/MSCHAPv2 when checking if EAP/PEAP is a viable authentication mechanism
- peap.c allocate buffers that needs to somehow be freed
- I yet have to complete a successful authentication using PEAPv0/MSCHAPv2

rustylife:
>I fixed up that CA file/dir crash you were having. Please give it another spin?

thanks it did the job

tls handshake goes through just fine when i comment all tls_set_*() functions and add SSL_CTX_set_verify(psm->ctx, SSL_VERIFY_NONE, NULL) instead

but if i leave those newly added tls_set_*() functions in place, openssl lib cannot go through with handshake:
PEAP: S bit is set, starting PEAP phase 1
sent [EAP Response id=0x1 PEAP...]
rcvd [EAP Request id=0x2 PEAP...]
PEAP TLS: L bit is set
PEAP TLS: continue handshake
certificate verify depth: 0
Certificate verication disabled or no peer name was specified
SSL_do_handshake(): error:00000001:lib(0):func(0):reason(1)

enaess:
@rustylife - somehow the remote name was never set, just did a rebase/push again. Please give it another spin. If a user doesn't care, the option "tls-verify-method" should be set to "none" and "remotename" can be omitted.

enaess:
I managed to actually get to the MSCHAPv2 authentication as a part of the inner eap method, but I have some problem with my testbed. will have to look at this after work.

enaess:
@rustylife The change is starting to come together. Are you able to take my branch / PR for a spin? Note that I had to additionally cherry pick the md4 change I committed to master a few days ago.

@Neustradamus who do you think should review this? @paulusmack @carlsonj and maybe @jjkeijser ?

Will need to update readme / man page, and do some more testing with fallback / nak of these auth mechanisms and LCP. Maybe also consider making a unit test given the CMAC stuff ...

rustylife:
@enaess thanks for the heads up, I pulled your peap branch and cherry-picked that md4 change that fixed broken MS-CHAP.
I created the following peer config:
```
pty "pptp win19 --nolaunchpppd"
name "Administrator"
#remotename "Administrator"
ipparam test
require-mppe-128
usepeerdns
noauth
nodefaultroute
tls-verify-method none
capath /etc/ssl/certs
```
And I could not get past TLS handshake.
I had to move all tls_set_*() functions to else block and had to add the following instead:
```
if (!strcmp(TLS_VERIFY_NONE, tls_verify_method))
	SSL_CTX_set_verify(psm->ctx, SSL_VERIFY_NONE, NULL);
else {
```
and I managed to connect to PEAP VPN server just fine.

enaess:
@rustylife this check is in place for EAT-TLS as well. Users are supposed to match the server certificate 

For example:

remotename sstp-test.com
tls-verify-method suffix

Will try to match your server certificate against *sstp-test.com

enaess:
To be sure, it's the certificate name of the authenticating server. Server certificate for SSTP is verified by sstpc

rustylife:
I tried specifying remotename, same as certificate's CN, and host name, it still would not accept the self signed certificate, and I don't want to import that Windows certificate to my Linux machine.
I get the "sekurity" part :smile:, it is just there should be a way to simply disable all checks for self signed certificates for lab testing, otherwise what is the point in having "tls-verify-method none" option that does not work?
What am I missing here?

enaess:
Got it, I think we are confusing the remotename/tls-verify-method options which operates on properties of the certificate vs. actually verifying the certificate chain. 

When my configuration looks like this, 
```
# remotename, # or if set, but specifies 
tls-verify-method none
```
then this is the logs I get:

```
Aug  1 09:31:50 optiplex pppd[152193]: rcvd [EAP Request id=0x3 PEAP...]
Aug  1 09:31:50 optiplex pppd[152193]: PEAP TLS: all bits are off
Aug  1 09:31:50 optiplex pppd[152193]: PEAP TLS: continue handshake
Aug  1 09:31:50 optiplex pppd[152193]: certificate verify depth: 1
Aug  1 09:31:50 optiplex pppd[152193]: certificate verify depth: 0
Aug  1 09:31:50 optiplex pppd[152193]: Certificate verication disabled or no peer name was specified
Aug  1 09:31:50 optiplex pppd[152193]: sent [EAP Response id=0x3 PEAP...]
Aug  1 09:31:50 optiplex sstpc[152195]: SEND SSTP DATA PKT(186) 
Aug  1 09:31:50 optiplex sstpc[152195]:   PPP EAP ID: 3 RESPONSE PEAP
```

The original code from eap-tls.c which I moved into tls.c is consistent with what was working with eap-tls.c

``` 
 454     if (strlen(cacertfile) == 0) cacertfile = NULL;
 455     if (strlen(capath) == 0)     capath = NULL;
 456 
 457     if (!SSL_CTX_load_verify_locations(ctx, cacertfile, capath))
 458     {
 459         error("EAP-TLS: Cannot load verify locations");
 460         if (cacertfile) dbglog("CA certificate file = [%s]", cacertfile);
 461         if (capath) dbglog("CA certificate path = [%s]", capath);
 462         goto fail;
 463     }
 464 
```

In case both cacertfile or capath is NULL, SSL_CTX_load_verify_locations will return 0.

OpenSSL man page:
``` 
0 - The operation failed because CAfile and CApath are NULL or the processing at one of the locations specified failed. Check the error stack to find out the reason.
```

You are correct that most Windows domains is using a self-signed certificate for internal services and the CA certificate is typically installed on each machine once you join the domain. This would typically be available when you got to http://ca.your.domain.org and request a certificate via Microsoft Certificate Authority services.

The way this is coded up in EAP-TLS is to either get the ca certificate or capath argument from the options, or by denoting it per client when calling get_eaptls_secret(); the CA is specified for each user in the eaptls-secrets file.

One could argue that CA certificate won't be necessary, I'd still want users of pppd/peap to always specify one if that's their intention. In fact, I'll probably add the CA certificate as an option to the network-manager-sstp project for the authentication step similar to what I did for eap-tls.

enaess:
If you specify the "ca" option and point it to a .pem file that contains the CA for your test-bed, this should work fine.

enaess:
Somehow I foo-bar'ed the branch trying to do a "git rebase" of my changes with master of ppp project. I created another branch and cherry-picked the changes I needed that was already sync'ed with ppp/master. Closing this PR in favor of #307

jjkeijser:
> I tried specifying remotename, same as certificate's CN, and host name, it still would not accept the self signed certificate, and I don't want to import that Windows certificate to my Linux machine.
> I get the "sekurity" part smile, it is just there should be a way to simply disable all checks for self signed certificates for lab testing, otherwise what is the point in having "tls-verify-method none" option that does not work?
> What am I missing here?

what you miss here is that "tls-verify-method none" does **NOT** allow for the use of self-signed certificates. Self-signed certificates (except for CA certs) are never meant for production use. All "tls-verify-method none" should do is that the NAME specified in the certificate is not checked against the server or client name, that's all.

Self-signed certificates are prone to man-in-the-middle attacks, can NOT be used to properly establish a secure TLS connection and require a separate means to check the validity of the other ends self-signed certificate, e.g. a pre-shared fingerprint.

jjkeijser:
I've been reviewing the PEAP branch a bit ...
It looks impressive and quite nice to add yet another EAP mechanism.
But are there any RFC to check this against?  I'm struggling a bit with the peap_prfplus function in peap.c:   it uses SHA1 - is that still correct? I understand that on the outside it will use a TLS connection (with TLSv1.2+ thus SHA2+) but is it **specified** somewhere that the PEAP phase 2 ms-chap auth is done using SHA1 ?  

Another thing about this entire branch which worries me is that it is impossible to test it without server-side support - if one would always need a Windows server to test it with, then I will never be able to test or debug it.

enaess:
@jjkeijser Thank you for your time to look over this. 

I emailed the author of the original patch, Rustam, and the documents are available here

https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-peap/5308642b-90c9-4cc4-beec-fb367325c0f9

Quite nicely documented.

I've been meaning to stick in a link to it in the source code

enaess:
I tested this with Microsoft server and using SSTP and it works. The MPPE key derivation is correct too.

Not sure what the best way to setup pppd in server mode for this. What do you use?

jjkeijser:
the problem is, EAP-PEAP or EAP-MSCHAPv2 server-side support is not written yet, it seems -  so I don't know how to test it. It's on my list of things to try to add server-side support and it **might** be quick, but then again, it might now and the time I have to spend on `pppd` is very very limited

enaess:
To be correct, someone added EAP-MSCHAPv2 server side support. Unless it was the EAP inside of the PEAP you meant

enaess:
By the way, the latest version of this PR is #307

jjkeijser:
> To be correct, someone added EAP-MSCHAPv2 server side support. Unless it was the EAP inside of the PEAP you meant 

which branch is the server side support on? how do I check it out or download a tarball?  too many PRs and branches :(

enaess:
Sorry, Git and I are slowly becoming friends. #307 is the latest PR. If you want you can send me a PR on that branch in my fork. This does not support P(EAP)-MSCHAPv2. 

I meant the eap.c code has MSCHAPv2 server side support.

If we been smart, each of these auth protocols could have a generic interface and be reused in both EAP.c and PEAP.c, but oh well. Other alternative here would be to import wpa_supplicant or whatever the newer project is called (or at least borrow SE of their design)

enaess:
I am heading to bed. It's 10pm here in Seattle.

Neustradamus:
@enaess: +1 to have a common interface...

PR #303, opened 2021-07-28
Using OpenSSL for MD4 isn't compatibile how NTPasswordHash is using it.

enaess:
Sadly, introducing autotools broke chap authentication. This change fixes it such that the workaround for broken internal MD4 works.

Note, the NTPasswordHash function should probably just call MD4() function directly and md4.c should implement that with the workaround. That way, this silliness isn't exposed and copied to other places (if MD4 would be used again, if ever?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ autom4te.cache
 /configure.scan
 /depcomp
 /install-sh
+/test-driver
 /missing
 /INSTALL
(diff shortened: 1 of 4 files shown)


enaess:
@paulusmack I'd advocate for taking this patch immediately to restore feature functionality. Sorry for having broken it in the first place :'(

PR #305, opened 2021-07-31
pppd: Remove usage of incorrect constant MAXIFNAMELEN

pali:
MAXIFNAMELEN is currently hardcoded to 32, but maximal size of interface
name on Linux is just 15 + nul-term byte. This limit is already provided by
IFNAMSIZ macro defined in net/if.h header file.

So replace MAXIFNAMELEN usage by IFNAMSIZ to not silently truncate
interface name.

___

pppd: Replace IF_NAMESIZE by IFNAMSIZ for struct ifreq

Macros IF_NAMESIZE and IFNAMSIZ should be defined to the same value, but
struct ifreq uses IFNAMSIZ. So use "correct" macro.

### DIFF
--- a/pppd/main.c
+++ b/pppd/main.c
@@ -124,7 +124,7 @@
 
 
 /* interface vars */
-char ifname[MAXIFNAMELEN];	/* Interface name */
+char ifname[IFNAMSIZ];		/* Interface name */
 int ifunit;			/* Interface unit number */
 
 struct channel *the_channel;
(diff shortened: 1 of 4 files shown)


enaess:
@paulusmack @Neustradamus @pali - The changes looks good to me.

enaess (approved):

pali:
@enaess Thanks for review!

enaess:
Just as a side note, 

I am not a particularly a fan of using declarations of IFNAMSIZ in public header files. It would be better if we could leave the data structures e.g. a char *, or char[] with an open ended value (but declared in a private header for internal use). The problem I am trying to solve is reducing the number of additional headers is needed to include pppd.h to zero. The better way would be to declare a function with appropriate in/out values to get this information instead of relying on internal data structures ... that might be a larger change however

PR #307, opened 2021-08-03
PEAP/MSCHAPv2 support in PPPD

enaess:
Using PEAP/MSCHAPv2 is now supported in PPPD with this change.

  * Added options to perform CA/CRL checking and certificate validation consistent with what is already been done for EAP-TLS
  * Certificate validation is performed similar to EAP-TLS. In addition to checking the certificate chain, users can now set "remotename" and "tls-verify-method" to validate the certificate name
  * Validation of certificate purpose and extended key usage is controlled the option "tls-verify-key-usage".
  * Support for MPPE key generation via PEAP
  * Man page is updated where appropriate for the new options.
  * Added unit-tests for the PEAP code in case of crypto or parameters would change in the future.
  * Added the PEAP feature to configure scripts. Users can now control the feature by specifying --enable-peap/--disable-peap.

To acheive feature parity with the EAP-TLS change, the EAP-TLS common code was refactored into tls.c/.h such that it could be re-used in both instances.

Support for PEAP was added in part by Rustam Kovhaev and Eivind Næss

### DIFF
--- a/pppd/Makefile.am
+++ b/pppd/Makefile.am
@@ -1,13 +1,14 @@
 sbin_PROGRAMS = pppd
 dist_man8_MANS = pppd.8
-check_PROGRAMS = \
-    utest_chap
+check_PROGRAMS =
 
 utest_chap_SOURCES = chap_ms.c pppcrypt.c utils.c
 utest_chap_CPPFLAGS = -DUNIT_TEST
 utest_chap_LDFLAGS =
 
-TESTS = $(check_PROGRAMS)
+utest_peap_SOURCES = peap.c utils.c mppe.c
+utest_peap_CPPFLAGS = -DUNIT_TEST -I${top_srcdir}/include
+utest_peap_LDFLAGS =
 
 if WITH_SRP
 sbin_PROGRAMS += srp-entry
@@ -37,14 +38,16 @@ pppd_include_HEADERS = \
     md4.h \
     md5.h \
     mppe.h \
-    pppdconf.h \
     patchlevel.h \
     pathnames.h \
+    peap.h \
     pppcrypt.h \
     pppd.h \
+    pppdconf.h \
     session.h \
     sha1.h \
     spinlock.h \
+    tls.h \
     tdb.h \
     upap.h 
 
@@ -84,6 +87,7 @@ endif
 if WITH_CHAPMS
 pppd_SOURCES += chap_ms.c
 pppd_SOURCES += pppcrypt.c
+check_PROGRAMS += utest_chap
 else
 if WITH_SRP
 pppd_SOURCES += pppcrypt.c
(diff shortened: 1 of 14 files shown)


enaess:
@paulusmack @Neustradamus @rustylife - I believe the work needed to be done on this PR is complete, a.k.a. ready to merge! Commentary is in #302, but due to unfortunate git rebase mishap, I created another pull request.

pppd/Makefile.am
```diff
@@ -181,3 +192,6 @@ pppd_LDADD = $(pppd_LIBS)
 
 EXTRA_DIST = \
     ppp.pam
+
+TESTS = $(check_PROGRAMS)
```
amosshapira, line 196: I'm out of date with the syntax of these files, but if I follow it right then this will execute `check_PROGRAMS` (a-la Bash) and assign the output to `TESTS`? What would happen if `check_PROGRAMS` is empty?

pppd/peap.h
```diff
@@ -0,0 +1,76 @@
+/*
+ *      Copyright (c) 2011
+ *
+ * Authors:
+ *
+ *	Rustam Kovhaev <rkovhaev@gmail.com>
+ */
+
+#ifndef PPP_PEAP_H
+#define	PPP_PEAP_H
```
amosshapira, line 10: excessive indentation compared to lines 20 and onwards

pppd/peap.h
```diff
@@ -0,0 +1,76 @@
...
+
+#define	PEAP_PHASE_1			1
+#define	PEAP_PHASE_2			2
+
+#define	PEAP_HEADERLEN			6
+#define	PEAP_FRAGMENT_LENGTH_FIELD	4
+#define	PEAP_FLAGS_FIELD		1
+#define	PEAP_FLAGS_ACK			0
+
+#define PEAP_CAPABILITIES_TYPE		254
```
amosshapira, line 20: Different indentation along the entire list of `#define`s

pppd/Makefile.am
enaess: I just ran this and commented out the $(check_PROGRAMS) leaving TESTS empty, and sure enough it didn't run any tests as expected. 

Once can configure without MSCHAP and without PEAP using configure, and in that case; none of the unit-tests would run. More likely is that you disable PEAP and also disable the unit-test.

pppd/peap.h
enaess: Can be fixed.

enaess: Looks like this file already does tab stop at 8 and is consistent. I had to set vim to ts=8 and the file displayed correctly. I could be wrong, but I am not seeing the problem here.

Neustradamus:
@paulusmack, @rustylife, @tisj, @jjkeijser : What do you think?

pppd/peap.h
amosshapira: This is what it looks like for me on GitHub:
Screen Shot 2021-08-04 at 2 10 36 pm

enaess: Yes, the GitHub display seems wrong

enaess:
@pali maybe a second pair of eyes would be helpful here?

pali:
Uff, I have not been debugging MSCHAPv2 stuff for many years... So I really cannot review MSCHAPv2 integration code...

enaess:
@pali Ha! Well, it wasn't so much testing or eyeing problems in the code (ps. it works), as much as making sure I didn't clobber stuff I shouldn't have, etc.

jjkeijser:
> @jjkeijser: Have you comments?

I will review this next week - a  HOWTO guide on how to set up both a Linux client and server to use EAP-MSCHAP will be helpful, as otherwise I cannot test the code.

enaess:
I just added a comment to peap.c on where to find the Microsoft documentation for the PEAP protocol and did a rebase against the master branch. Seems like the two changes I wanted to commit originally is still intact ...

enaess:
For details on the PEAP protocol, look to Microsoft:
 -    https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-peap

enaess:
@jjkeijser - Currently, this is only provides the client side support. You'll have to setup a Microsoft RRAS server to test this (or at least for now).

paulusmack:
@enaess If you reckon this is ready to good, I'll take it, though I would also like to hear from @jjkeijser.

@jjkeijser : even if you can't test the code, I'd like to hear if you think it looks OK or not.

enaess:
This is ready to go from my side. @jjkeijser mentioned he wanted to look into adding the server side portion.

enaess:
A lot of @jjkeijser comments were added / lost in other incarnations of this PR (#302, others?)

paulusmack:
I'm happy enough to take the code, but I notice that pppd/peap.c says "Copyright (c) 2011" and lists Rustam Kovhaev as the author but does not have any licence statement giving permission to use the code. Since there is an explicit assertion of copyright, I think there needs to be an explicit mention of a licence. Is Rustam Kovhaev contactable?

enaess:
@paulusmack I have been in contact with him since the original patch was submitted. I kind of assumed 2011 was a typo, but I'll try to reach out him separately by email as well as tagging him here @rustylife. He is out of Vacouver, BC. Feels close enough, I could go knock on his door ;-)

rustylife:
hi @paulusmack @enaess the code can be freely used and redistributed
let's use the same license statement as in pppd/mppe.c file

enaess:
@rustylife what about the copyright, is 2011 correct or should that say 2021?

enaess:
K, I'll update the file(s).

enaess:
@paulusmack I have those files updated with license information.

Neustradamus:
@enaess: +1!

rustylife:
Thank you very much @paulusmack !

Neustradamus:
Thanks to @enaess for the PR (and @rustylife too), @amosshapira for review, and @paulusmack for merging.

PR #308, opened 2021-08-04
Fix compiler warnings

enaess:
This is really just house cleaning. Cleaning up compiler warnings for 
* GCC 
* CLANG

### DIFF
--- a/chat/chat.c
+++ b/chat/chat.c
@@ -201,7 +201,7 @@ SIGTYPE sighup (int signo);
 void unalarm (void);
 void init (void);
 void set_tty_parameters (void);
-void echo_stderr (int);
+int  echo_stderr (int);
 void break_sequence (void);
 void terminate (int status);
 void do_file (char *chat_file);
@@ -210,7 +210,7 @@ int  put_string (register char *s);
 int  write_char (int c);
 int  put_char (int c);
 int  get_char (void);
-void chat_send (register char *s);
+int  chat_send (register char *s);
 char *character (int c);
 void chat_expect (register char *s);
 char *clean (register char *s, int sending);
@@ -983,16 +983,18 @@ char *character(int c)
 /*
  *  process the reply string
  */
-void chat_send (register char *s)
+int chat_send (register char *s)
 {
     char file_data[STR_LEN];
+    int ret = 0;
 
     if (say_next) {
 	say_next = 0;
 	s = clean(s, 1);
-	write(2, s, strlen(s));
+	ret = strlen(s);
+	ret = write(2, s, ret) != ret;
         free(s);
-	return;
+	return ret;
     }
 
     if (hup_next) {
@@ -1001,13 +1003,13 @@ void chat_send (register char *s)
            signal(SIGHUP, SIG_IGN);
         else
            signal(SIGHUP, sighup);
-        return;
+        return ret;
     }
 
     if (echo_next) {
 	echo_next = 0;
 	echo = (strcmp(s, "ON") == 0);
-	return;
+	return ret;
     }
 
     if (abort_next) {
@@ -1027,7 +1029,7 @@ void chat_send (register char *s)
 
 	if (verbose)
 	    msgf("abort on (%v)", s1);
-	return;
+	return ret;
     }
 
     if (clear_abort_next) {
@@ -1057,7 +1059,7 @@ void chat_send (register char *s)
         free(s1);
 	if (pack)
 	    pack_array(abort_string,old_max);
-	return;
+	return ret;
     }
 
     if (report_next) {
@@ -1075,7 +1077,7 @@ void chat_send (register char *s)
 	
 	if (verbose)
 	    msgf("report (%v)", s1);
-	return;
+	return ret;
     }
 
     if (clear_report_next) {
@@ -1106,7 +1108,7 @@ void chat_send (register char *s)
         if (pack)
 	    pack_array(report_string,old_max);
 	
-	return;
+	return ret;
     }
 
     if (timeout_next) {
@@ -1120,7 +1122,7 @@ void chat_send (register char *s)
 	if (verbose)
 	    msgf("timeout set to %d seconds", timeout);
 
-	return;
+	return ret;
     }
 
     /*
@@ -1167,6 +1169,7 @@ void chat_send (register char *s)
 
     if (!put_string(s))
 	fatal(1, "Failed");
+    return ret;
 }
 
 int get_char(void)
@@ -1292,9 +1295,10 @@ int put_string(register char *s)
  *	When called with -1, a '\n' character is generated when
  *	the cursor is not at the beginning of a line.
  */
-void echo_stderr(int n)
+int echo_stderr(int n)
 {
     static int need_lf;
+    int ret = 0;
     char *s;
 
     switch (n) {
@@ -1305,15 +1309,17 @@ void echo_stderr(int n)
 	    break;
 	/* fall through */
     case '\n':
-	write(2, "\n", 1);
+	ret = write(2, "\n", 1) != 1;
 	need_lf = 0;
 	break;
     default:
 	s = character(n);
-	write(2, s, strlen(s));
+	ret = strlen(s);
+	ret = write(2, s, ret) != ret;
 	need_lf = 1;
 	break;
     }
+    return ret;
 }
 
 /*
--- a/pppd/main.c
+++ b/pppd/main.c
@@ -753,6 +753,7 @@ void
 detach(void)
 {
     int pid;
+    int ret;
     char numbuf[16];
     int pipefd[2];
 
@@ -774,7 +775,10 @@ detach(void)
 	exit(0);		/* parent dies */
     }
     setsid();
-    chdir("/");
+    ret = chdir("/");
+    if (ret != 0)
+	fatal("Could not change directory to '/'");
+
     dup2(fd_devnull, 0);
     dup2(fd_devnull, 1);
     dup2(fd_devnull, 2);
@@ -1414,7 +1418,7 @@ hup(int sig)
 	kill_my_pg(sig);
     notify(sigreceived, sig);
     if (waiting)
-	write(sigpipe[1], &sig, sizeof(sig));
+	(void) (write(sigpipe[1], &sig, sizeof(sig)) + 1);
 }
 
 
@@ -1434,7 +1438,7 @@ term(int sig)
 	kill_my_pg(sig);
     notify(sigreceived, sig);
     if (waiting)
-	write(sigpipe[1], &sig, sizeof(sig));
+	(void) (write(sigpipe[1], &sig, sizeof(sig)) + 1);
 }
 
 
@@ -1447,7 +1451,7 @@ chld(int sig)
 {
     got_sigchld = 1;
     if (waiting)
-	write(sigpipe[1], &sig, sizeof(sig));
+	(void) (write(sigpipe[1], &sig, sizeof(sig)) + 1);
 }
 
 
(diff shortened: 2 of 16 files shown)


chat/chat.c
```diff
@@ -986,11 +986,12 @@ char *character(int c)
 void chat_send (register char *s)
 {
     char file_data[STR_LEN];
+    int n;
```
pzygielo, line 989: It's possible I'm wrong - but isn't this variable unused (not ever read) now?

enaess: Correct, the compiler warning only requires that the return value is being assigned. It doesn't do any deeper heuristic in analyzing the code ...

pzygielo: So it's not about fixing warnings but silencing them, then?

enaess: Good point, I'll update the PR later. In this case, what do you think the best option is? if (n != len) { then printf(), something else?

pzygielo: I'm not maintainer here, so my comments might be ignored.

I'd rather be bothered with valid warning to be solved in the future, than have it disappeared  with no resolution.

I don't know PPP code very well - so it's hard for me to suggest anything.
It's also possible that proposed changes are already fine.

enaess: You are completely right. There is no need for a fix that only introduces another warning, or just masks the error. I'll have another look.

enaess: The chat.c was contributed more than 20 years ago, and by looking at the git logs; it's only gotten a handful changes related to warnings, and 1 or 2 buffer overflows. I could change the return code, and make the function(s) return -1 if it fails to write to stderr. The changes in chat.c is some instances of this, but for example pppd's main.c has 4 more write() related compiler warnings. It would be up to the caller to handle the return code, and likely discard it. 

These cannot be fixed by adding a log-message (possibly deadlock the application) and in those cases I think these warnings just needs to be ignored. There is a way to do this without introducing new warnings (currently):

https://stackoverflow.com/questions/11888594/ignoring-return-values-in-c

One of the answers are similar to this:
```
(void) (write(...) + 1);
```
The return is used, and then discarded via the void statement. 

In other cases where e.g. the return value of functions such as setuid/setgid requires processing, properly log a message and die() could be the right solution. If you can't execute the ipup script as root, your pretty much f**@# anyway.

pali:
It seems that currently autotools do not enable -Wall nor -W(extra) warnings by default. Should not be these warnings enabled? Otherwise new pull requests could introduce new warnings as developers could not be aware of the fact that warnings are disabled for their local builds...

chat/chat.c
pali: > ```
> (void) (write(...) + 1);
> ```

This is really a hack. If return value is not used and it should be (e.g. for write() to check that it did not failed, e.g. for EINTR) then it is not a good idea to mute warning with some hack. It is better to let warning visible so other people know about issue and maybe in future somebody can properly fix it. Muting warning with hack (instead of fixing it) means that warning stay hidden and probably nobody fix code in future.

chat/chat.c
```diff
@@ -1305,15 +1309,17 @@ void echo_stderr(int n)
...
     case '\n':
-	write(2, "\n", 1);
+	ret = write(2, "\n", 1) != 1;
 	need_lf = 0;
 	break;
     default:
 	s = character(n);
-	write(2, s, strlen(s));
+	ret = strlen(s);
+	ret = write(2, s, ret) != ret;;
```
pali, line 1318: two semicolons at the end.

chat/chat.c
```diff
@@ -1292,9 +1295,10 @@ int put_string(register char *s)
  *	When called with -1, a '\n' character is generated when
  *	the cursor is not at the beginning of a line.
  */
-void echo_stderr(int n)
+int echo_stderr(int n)
```
pali, line 1298: It looks like that caller of `echo_stderr()` ignores return value, so I'm not sure if it makes sense to change return from void to int if it is always ignored...

chat/chat.c
enaess: Updated change

chat/chat.c
enaess: If the function fail to write to stderr, then all bets are off. The two places this is used is in terminate() and getstring(), the former pretty much terminates the application and the later may have some significance. But either way, if you cannot write to stderr something might be very off and trying to call msgf() with a warning may not do you any good either. 

While good coding practices should include both having a function return a value and also processing this at the next level up, I'd say this is more a part of code evolution. Eventually, functions like chat_send() may end up wanting to return a different value for a different scenario and it'll be one less change to make...

Do you have a better idea? Or should we just not fix this? Perhaps let the user (distribution maintainer) specify -Wno-unused-result and then we can go about our day as usual?

enaess:
@pali 

How do you suggest enabling -Wall or -Wextra with Automake in a safe and generic way? Just tried -Wextra and it is very verbose; I could agree to -Wall if we could remove the RCSID[] variables that are put in each file, then eliminate the unused variables. Using AM_INIT_AUTOMAKE([-Wall]) spewed out a bunch of errors.

We could use -Wall by default (e.g. add a AC_COMPILE_IFELSE) and fix-up some of these issues, but let the distribution maintainers specify CFLAGS with configure based on the compiler they have, e.g. specify "-Wall -Wno-unused-result". This way, we can be pretty ignorant about the compiler in use (clang, gcc, or whatever they use on SunOS).

Ideally, if we fix up all (and willfully ignore the ones we chose), then using the right combination (e.g. -Wall -Wno-unused-results -Werror) on the Ubuntu/Linux CI script would cause it to fail and thus  help enforce a better code quality for new patch submissions.

Do you have any thoughts on this patch?

* Fix these for chat / pppd (despite using a hack), this will avoid having us use -Wno-unused-results
* Use -Wall -Wno-unused-results -Werror on Linux/Ubuntu CI builds?
* Take this patch - but leave out changes to pppd/chat 
* Not take this patch

pali:
Well, most changes in this pull requests are fine for me. Printing error when something fails is basically fixing code, something which is really useful (and I would not call it fix warning; rather I would call it, fixing errors about which compiler was complaining via warning reports).

Also change in pppoatm.c is not about "fixing warnings". It is rewriting code from some old GNU C dialect to standard ISO C99. Something which is useful as it allows that code to be compiled also by non-GNU compiler.

What I dislike are mostly changes in chat as they are not fixing any issue. They just try to hack compiler so it cannot deduce (with current logic analyzer) that return values are unused. And in the same way I dislike `+1` hacks in other source files. This pattern does not fix anything (return value is still unused!), just make code harder to read.

About enabling compiler warnings, if it is possible to enable something from -Wall at least on CI, it can really help for development. But not at the cost of adding hacks which do not fix anything.

enaess:
I can certainly remove the `+1` hack from this change in pppd/main.c and revert the chat.c modifications. The compiler warnings in chat.c caused by writing to stderr seems dumb, and trying to do something useful e.g. ```while (try-- > 0) n = write(pipefd, " ", 1); ``` is risky business (got to be able to test the change too. 

Having that said, the gcc compiler will still complain about these. But, if we enable the -Wall -Wno-unused-result -Werror on the "Build and Test" CI script, then we'll capture any warnings in new patches going forward. This would, however, divert the attention to fix other places (like remove unused variables, etc).

We could try to add some additional magic to the configure.ac to add these compiler flags, @thesamesam? But likely, setting them outright would tie the project to be tied to a particular compiler.

enaess:
I would argue for fixing the code in places where it can be fixed. The end goal should be to fail the CI script on new changes that does introduce new compiler warnings (otherwise, this change would not make sense). Allowing -Wno-unused-results would be the tie breaker to make it work.

paulusmack:
Is it possible to use a #pragma to tell gcc not to warn about specific instances where (e.g.) return values are ignored? While there are certainly places where it would be better to check a return value that isn't checked today, there are others where if the call fails there is really nothing we can do about it - like in cases of writing an error message to stderr before exiting. And there are other cases where the effect of the call is checked (e.g. the setuid() followed by getuid()) though the return value of the call isn't.

In the cases where the result arguably should be checked, it usually requires a careful analysis to say what to do about it. For example, just calling fatal() might not be appropriate, depending on what the impact of the failure will be on pppd's ability to run correctly.

Since I'm going to have to look at each change carefully, I think I'd rather see smaller commits with a rationale for each change described in the commit message.

enaess:
Sounds good to me. Though after some thinking, I kind of tilted towards @pali opinion of not fixing them. The #pragma idea could be the solution. I could also re-do the patch set in smaller increments if needed.

pppd/plugins/radius/util.c
```diff
@@ -78,7 +78,7 @@ rc_mksid (void)
   snprintf(buf, sizeof(buf), "%08lX%04X%02hX",
 	   (unsigned long int) time (NULL),
 	   (unsigned int) getpid (),
-	   cnt & 0xFF);
+	   (unsigned short) (cnt & 0xFF));
```
paulusmack, line 81: This really seems like the compiler is being over-zealous. By C argument passing rules, this is going to get coerced back to int anyway. And it is clear that the value fits in 16 bits, so there is no possible loss of information in the hX conversion.

paulusmack (approved):
Despite the comment, overall this looks fine to me.

paulusmack:
Actually I meant that comment and review to apply only to that one commit, not the overall PR.

chat/chat.c
paulusmack: Interesting that the code is using both write(2, ... and fprintf(stderr, ...

Clearly we don't want to call fatal() or terminate() from echo_stderr(), since echo_stderr() is called from terminate(). However, arguably the call to echo_stderr() in get_string() could call fatal() if echo_stderr() returns failure.

pppd/main.c
```diff
@@ -1414,7 +1418,7 @@ hup(int sig)
 	kill_my_pg(sig);
     notify(sigreceived, sig);
     if (waiting)
-	write(sigpipe[1], &sig, sizeof(sig));
+	(void) (write(sigpipe[1], &sig, sizeof(sig)) + 1);
```
paulusmack, line 1421: This (void) (... + 1) thing seems like a horrible hack to me. If we can use something like #pragma no_warn_unused or something here that would be better.

The consequence of the write to sigpipe[1] failing is that pppd may not wake up until another event happens. Would it be better to die here if the write fails? It seems like that would just turn a potential hang into a definite crash. Since we can't tell the user why at this point, it seems better to try to limp along. Are there contrary arguments?

paulusmack: If we can't work out something using #pragma (and with a comment outlining the reasoning), it's probably better to leave the warning in place than do the (void)(... + 1) hack.

pppd/main.c
```diff
@@ -1676,11 +1681,16 @@ device_script(char *program, int in, int out, int dont_wait)
...
 
     /* here we are executing in the child */
-
-    setgid(getgid());
-    setuid(uid);
-    if (getuid() != uid) {
-	fprintf(stderr, "pppd: setuid failed\n");
+    ret = setgid(getgid());
+    if (ret != 0) {
+	fprintf(stderr, "pppd: setgid failed, %s (%d)\n",
```
paulusmack, line 1686: This could be just perror(...) (and also the next fprintf).

pppd/utils.c
```diff
@@ -907,10 +907,13 @@ lock(char *dev)
     pid = getpid();
 #ifndef LOCK_BINARY
     slprintf(lock_buffer, sizeof(lock_buffer), "%10d\n", pid);
-    write (fd, lock_buffer, 11);
+    n = write (fd, lock_buffer, 11);
+    if (n != 11)
```
paulusmack, line 911: hmmm, I don't like leaving an if () ... hanging like this at the #else.

paulusmack (changes requested):
Overall, I would prefer this split into smaller chunks, say one commit per file changed. The commit messages should then discuss the cases where it is not at all clear what the best response to failure is, e.g. the write to the signal pipe in hup().

enaess:
I am closing this PR in favor of the new and updated PR #317

PR #313, opened 2021-08-29
CI: Perform installation step

Chocobo1:
This ensures installation code to be tested.

### DIFF
--- a/.github/workflows/solaris.yaml
+++ b/.github/workflows/solaris.yaml
@@ -15,3 +15,4 @@ jobs:
           pkg install gcc automake autoconf libtool
           ./autogen.sh CFLAGS="-Wno-deprecated-declarations"
           make
+          make install
--- a/.github/workflows/ubuntu.yaml
+++ b/.github/workflows/ubuntu.yaml
@@ -17,8 +17,5 @@ jobs:
     - name: build
       run: make V=1 CFLAGS="-O2 -g -Wno-unused-result"
 
-    - name: check
-      run: make check
-
     - name: distcheck
       run: make distcheck DISTCHECK_CONFIGURE_FLAGS="--enable-multilink --enable-ipxcp"
--- a/.github/workflows/ubuntu.yaml
+++ b/.github/workflows/ubuntu.yaml
@@ -4,6 +4,8 @@ on: [push, pull_request]
 jobs:
   build_and_test:
     runs-on: ubuntu-latest
+    env:
+      configure_flags: --enable-ipxcp --enable-multilink
 
     steps:
     - uses: actions/checkout@v2
@@ -12,10 +14,10 @@ jobs:
       run: sudo DEBIAN_FRONTEND=noninteractive apt-get -y -qq install build-essential autoconf automake pkg-config libtool m4 autoconf-archive libssl-dev libatm1-dev libpcap-dev
 
     - name: configure
-      run: ./autogen.sh --enable-multilink --enable-ipxcp
+      run: ./autogen.sh ${{ env.configure_flags }}
 
     - name: build
       run: make V=1 CFLAGS="-O2 -g -Wno-unused-result"
 
     - name: distcheck
-      run: make distcheck DISTCHECK_CONFIGURE_FLAGS="--enable-multilink --enable-ipxcp"
+      run: make distcheck DISTCHECK_CONFIGURE_FLAGS="${{ env.configure_flags }}"


Neustradamus:
@RICCIARDI-Adrien, @enaess: What do you think?

enaess:
Not opposed to this by any means

RICCIARDI-Adrien:
OK for me too.

Neustradamus:
@paulusmack: Can you look?

paulusmack:
Seems OK if others are happy with it.

I don't know who cancelled the checks or why.

Neustradamus:
@paulusmack: Good?

RICCIARDI-Adrien:
This PR-introduced action seems to be working well : https://github.com/ppp-project/ppp/runs/3643340908?check_suite_focus=true

enaess:
@Chocobo1 - The initial CI script did a "make distcheck" this target does a make install into a directory, then actually tries to build it again to make sure the distribution archive is correct. Do you still think this step is needed?

See the output of "make distcheck" step.

```
 # This is the dist archive (.tar.gz being unzipped into _build directory, then configure, make make install ...
 dc_install_base=`CDPATH="${ZSH_VERSION+.}:" && cd ppp-2.4.10-dev/_inst && pwd | sed -e 's,^[^:\\/]:[\\/],/,'` \
  && dc_destdir="${TMPDIR-/tmp}/am-dc-$$/" \
  && am__cwd=`pwd` \
  && CDPATH="${ZSH_VERSION+.}:" && cd ppp-2.4.10-dev/_build/sub \
  && ../../configure \
     \
    --enable-multilink --enable-ipxcp \
    --srcdir=../.. --prefix="$dc_install_base" \
  && make  \
  && make  dvi \
  && make  check \
  && make  install \
  && make  installcheck \
  && make  uninstall \
  && make  distuninstallcheck_dir="$dc_install_base" \
        distuninstallcheck \
```

While no real objection to the change to me, I just happen to question if it is absolutely necessary.

Chocobo1:
>@Chocobo1 - The initial CI script did a "make distcheck" this target does a make install into a directory, then actually tries to build it again to make sure the distribution archive is correct. Do you still think this step is needed?

I'm less aware of the `distcheck` target and from the [doc](https://www.gnu.org/software/automake/manual/html_node/Checking-the-Distribution.html), seems it is performing the installation to a temporary directory while `sudo make install` is actually installing to the OS and I think this is actually a difference.

A follow up question (not related to this PR) is: `make distcheck` also performs `make check` and I think it make the following step redundant: https://github.com/ppp-project/ppp/blob/7adcfc961d3ccb051da87cc95d76cd3344f00fb8/.github/workflows/ubuntu.yaml#L20-L21

enaess:
I believe you are correct, the "check" is redundant given that "distcheck" is executed. 

For practical reasons, I found that the "make distcheck" have helped me catch multiple mistakes. Especially with regards to targets or files not included in the distribution archive (e.g. a EXTRA_DIST is missing from one of the automake files). 

Ultimately, the "dist" archive (i.e. ppp-<version>.tar.gz) is what we try to ensure the integrity of. Developers will typically do a copy of pppd or "make install" to attempt to test the change/fix before they commit.

enaess:
The VPATH and sub-directory usage of the distcheck target help exercise the uncommon options. 

Is there a compelling reason to add the "make install" target? What was the problem you ran into that prompted you to craft this fix?

Chocobo1:
>What was the problem you ran into that prompted you to craft this fix?

It is more like improving the CI/test coverage than attempting of some fix and I wasn't aware of the details of `make distcheck` at the time I wrote the patch/PR. I suppose if there is really no benefits of having `make install` then I should remove it from ubuntu.yaml. But what about solaris.yaml, is it worth to add `make install` there? at least it is performing CI on a different platform.

Chocobo1:
>Solaris is slightly different as it also bails out of the automake stuff to build a kernel driver. I am not sure the "make distcheck" will correctly handle the extra build output here (or at least complain).

It errors: https://github.com/Chocobo1/ppp/runs/3659611481?check_suite_focus=true#step:3:2375
I'll leave it for others who are interested to fix.

>When I introduced the ubuntu.yaml file, it was more a quick and dirty check for the system to ensure we have a valid package. 

I think it is mostly good. I made two more commits to slightly improve it.

enaess:
@paulusmack Change looks good to me (ready to merge), but it looks like you may need to approve the workflows to run first. Merge after a successful run.

enaess:
@Chocobo1 now as my compiler-warning-fixes is in, it could be possible to enable CFLAGS="-Werror" in the configure statement on the Linux (Fedora/Ubuntu) based CI steps. In that way we'd know if we are trying to merge changes that introduces new compiler warnings. An idea would be to add e.g. Fedora and compile the project using CLANG. Ubuntu already uses GCC.

Neustradamus:
@Chocobo1, @paulusmack: Have you seen @enaess comments?

PR #314, opened 2021-09-08
pppd: Fix usage of BOTHER ioctl API on Linux

pali:
Linux architectures have different content of struct termios2 and also
different value of BOTHER macro. So do not declare any struct termios2 nor
BOTHER macro. Current definitions in ppp were applicable only for x86.

Correct definitions for current architecture are only in <asm/termbits.h>
and <asm/ioctls.h> header files. But Linux header file <asm/termbits.h> is
in conflict with glibc header file <termios.h> and only one can be included
in one source unit. Moreover both header files contains struct termios but
with different content. So it is not possible to use glibc tc* functions
with <asm/termbits.h> definitions.

For this reason provide a new include header file "termios_linux.h" which
provides custom implementation of all glibc's termios.h functions via Linux
ioctl() interface with definitions from Linux <asm/termbits.h> header file.

Thus this "termios_linux.h" is replacement for <termios.h> with additional
support for BOTHER Linux termios API.

Same "termios_linux.h" is going to be used by U-Boot's kwboot utility for
the same reason to use arbitrary baudrate value via BOTHER ioctl API.

Hopefully one day glibc will provide some API functions for functionality
provided currently by BOTHER Linux API.

Cc: @elkablo

### DIFF
--- a/pppd/Makefile.am
+++ b/pppd/Makefile.am
@@ -72,7 +72,7 @@ pppd_LDFLAGS =
 pppd_LIBS =
 
 if LINUX
-pppd_SOURCES += sys-linux.c
+pppd_SOURCES += sys-linux.c termios_linux.h
 pppd_LIBS += $(CRYPT_LIBS) $(UTIL_LIBS)
 endif
 
--- a/pppd/sys-linux.c
+++ b/pppd/sys-linux.c
@@ -96,7 +96,6 @@
 #include <signal.h>
 #include <fcntl.h>
 #include <ctype.h>
-#include <termios.h>
 #include <unistd.h>
 
 /* This is in netdevice.h. However, this compile will fail miserably if
@@ -110,7 +109,6 @@
 
 #if !defined(__GLIBC__) || __GLIBC__ >= 2
 #include <asm/types.h>		/* glibc 2 conflicts with linux/types.h */
-#include <asm/ioctls.h>
 #include <net/if.h>
 #include <net/if_arp.h>
 #include <net/route.h>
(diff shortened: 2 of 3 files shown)


Neustradamus:

paulusmack:
Where did termios_linux.h come from? Its licence is GPL which is incompatible with the rest of pppd. (Yes, I know we have plugins that are GPL'd, but pppd itself does not depend on them, they depend on pppd.)

Having our own header that defines the ioctl interface is fine, but it will have to be reimplemented rather than copied from the Linux kernel source.

(I would love to change pppd's licence to something better than the 4-clause BSD licence it is currently under, but most of it is copyrighted by CMU, and when I have asked CMU for permission to change the licence in the past it has not agreed. I could try again, I guess.)

pali:
> Where did termios_linux.h come from?

Written by me and @elkablo.

> Having our own header that defines the ioctl interface is fine, but it will have to be reimplemented rather than copied from the Linux kernel source.

It does **not** come from Linux kernel nor from any other project. We have written this header file as part of fixing U-Boot's kwboot tool.

This header file is replacement for glibc's `<termios.h>` header file as glibc's does not support BOTHER and arbitrary baudrates.

> Its licence is GPL which is incompatible with the rest of pppd.

What license do you want? I guess that @elkablo would not be against other open source license.

pali:
The aim is to use this header file `"termios_linux.h"` also in other projects which want/need support for arbitrary baudrates on Linux... (until glibc adds this support)

paulusmack:
OK, that's good!

The main part of pppd is under 4-clause BSD (unfortunately), so the licence would need to be something compatible with that - e.g. 2 or 3 (or 4) clause BSD, MIT, etc. Given that you and @elkablo wrote it, you can put on it whatever licence you both agree to (and even license different copies in different ways).

elkablo:
I have no problem with dual licensing GPL / 4-clause BSD

elkablo:
@pali but please change my e-mail to that at kernel.org

pali:
Is this header OK?
`/* SPDX-License-Identifier: GPL-2.0+ OR BSD-4-Clause OR BSD-3-Clause OR BSD-2-Clause */`

pali:
No objection, so I changed it to above variant.

pali:
@paulusmack It is ok now?

Neustradamus:
Thanks to @pali, @elkablo and @paulusmack!

PR #317, opened 2021-09-23
Fix compiler warnings

enaess:
This is take two of PR#308 breaking each change out in a separate commit to make it easier for the reviewer to figure out what changed. Also incorporating feedback from @pali and @paulusmack. Change fixes compiler warnings on Ubuntu w/GCC and CLang installed.

This closes PR#308 in favor of this new PR

Signed-Off: Eivind Naess <eivnaes@yahoo.com>

### DIFF
--- a/pppd/plugins/pppol2tp/pppol2tp.c
+++ b/pppd/plugins/pppol2tp/pppol2tp.c
@@ -514,15 +514,15 @@ void plugin_init(void)
 }
 
 struct channel pppol2tp_channel = {
-    options: pppol2tp_options,
-    process_extra_options: NULL,
-    check_options: &pppol2tp_check_options,
-    connect: &connect_pppol2tp,
-    disconnect: &disconnect_pppol2tp,
-    establish_ppp: &generic_establish_ppp,
-    disestablish_ppp: &generic_disestablish_ppp,
-    send_config: &send_config_pppol2tp,
-    recv_config: &recv_config_pppol2tp,
-    close: NULL,
-    cleanup: NULL
+    .options = pppol2tp_options,
+    .process_extra_options = NULL,
+    .check_options = &pppol2tp_check_options,
+    .connect = &connect_pppol2tp,
+    .disconnect = &disconnect_pppol2tp,
+    .establish_ppp = &generic_establish_ppp,
+    .disestablish_ppp = &generic_disestablish_ppp,
+    .send_config = &send_config_pppol2tp,
+    .recv_config = &recv_config_pppol2tp,
+    .close = NULL,
+    .cleanup = NULL
 };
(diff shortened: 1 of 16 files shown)


pppd/plugins/pppol2tp/pppol2tp.c
```diff
@@ -241,10 +241,10 @@ static void send_config_pppol2tp(int mtu,
 		sprintf(&reorderto[0], "%d ", pppol2tp_reorder_timeout);
 	tid[0] = '\0';
 	if (pppol2tp_tunnel_id > 0)
-		sprintf(&tid[0], "%u ", pppol2tp_tunnel_id);
+		sprintf(&tid[0], "%u ", (short) pppol2tp_tunnel_id);
```
pali: This change is for sure incorrect. C printf modifier `%u` expects `unsigned int` type, but passed type is `short signed int`. Compiler should throw a warning about it.

pppd/plugins/pppol2tp/pppol2tp.c
```diff
@@ -241,10 +241,10 @@ static void send_config_pppol2tp(int mtu,
...
 	sid[0] = '\0';
 	if (pppol2tp_session_id > 0)
-		sprintf(&sid[0], "%u ", pppol2tp_session_id);
+		sprintf(&sid[0], "%u ", (short) pppol2tp_session_id);
```
pali: Same here.

pali (changes requested):
Just two comments below. Otherwise looks good.

pppd/plugins/pppol2tp/pppol2tp.c
enaess: You are right, updated the change list. Must have eyeballed it when copying it from the original patch set. Thanks for the review.

pppd/plugins/pppol2tp/pppol2tp.c
```diff
@@ -241,10 +241,10 @@ static void send_config_pppol2tp(int mtu,
 		sprintf(&reorderto[0], "%d ", pppol2tp_reorder_timeout);
 	tid[0] = '\0';
 	if (pppol2tp_tunnel_id > 0)
-		sprintf(&tid[0], "%u ", pppol2tp_tunnel_id);
+		sprintf(&tid[0], "%hu ", (short) pppol2tp_tunnel_id);
```
pali, line 244: This is still incorrect for two reasons:
* `%hu` expects `unsigned short`, but you are passing `signed short` type cased from `signed int` type
* `pppol2tp_tunnel_id` is user parameter of (32-bit) integer type, so this change do some truncatenation

Anyway, this is used only for logging purposes and moreover L2TP tunnel id is 16-bit unsigned number, so it needs some more work to correctly fix these issues (add validation on input and e.g. disallow specifying too big number).

But in any case for debug / logging purposes, application should print what is really stored in variable, and not truncated content of variable. So I think that correct format should be really `%u` (or better `%d` as `pppol2tp_session_id` is `int`) and `pppol2tp_session_id` should be filled without any truncatenation.

pali: Looking at this code and warning is probably about small array size. So I think it would be better to really increase size of `tid` array instead of juggling with format types, casting and truncating to mute compile warning (which is fragile after 10 years nobody would really understand why value is truncated).

paulusmack:
Thanks @enaess , this looks fine.

enaess:
Wooohooo!!! - Thanks :)

PR #319, opened 2021-10-28
pppol2tp.c add '#ifdef INET6' for optional ipv6

str8fast:
Without it, l2tp daemon can't launch ppp, cuz undefined symbol ipv6_up_notifier.


Neustradamus:

paulusmack:
No sign-off, but I merged it anyway because it is trivial.

PR #320, opened 2021-11-14
add command-line options to set paths of ipv6-{up,down} scripts

telent:
I'd like to be able to specify the pathnames for ipv6-up, ipv6-down just like I can for ip-up and ip-down, so I added some options. Hope this makes sense


telent:
Closed in favour of #321 which is a single commit and has a signed-off-by line

PR #325, opened 2021-12-10
pppd: Allow to use additional Bnnn constants

pali:
These constants are supported by Linux kernel on SPARC architecture.


pali:
I have also sent update to [termios(3)](https://man7.org/linux/man-pages/man3/termios.3.html) manpage. And now _Line speed_ section already contains this fact about additional Bnnn constants on SPARC.

PR #326, opened 2021-12-11
ppp: Drop linux IPX support

rpurdie:
The 5.15 Linux kernel has removed ipx support, along with the userspace
visible header. This support wasn't very well maintained in the kernel
for several years so drop the support from ppp as well since this won't
be usable in future.

Signed-off-by: Richard Purdie <richard.purdie@linuxfoundation.org>


pali:
Why to remove it? It is still supported by LTS kernel versions and at configure time it can be enabled / disabled based on configure option.

rpurdie:
The kernel support has been buggy for a while so whilst it is there, it likely doesn't work too well which is why it was removed entirely. Most LTS situations don't upgrade userspace tools at the same time as using LTS kernels so removing from ppp shouldn't cause problems for most LTS scenarios. If you wish to continue to support IPX in ppp that is fine. I was asked about the patch we carry in Yocto Project/OpenEmbedded to disable it and opted to share our view of what needs to happen. As you say, it can be disabled although not so easily in the released versions of ppp.

pali:
> As you say, it can be disabled although not so easily in the released versions of ppp.

Ehm? It is disabled by default at compile time unless you specify `--enable-ipxcp`. And also it is disabled by at runtime unless you specify `ipx` command line argument (or in config file).

So what is the issue here? I'm feeling that this is an issue in your Yocto Project/OpenEmbedded if it automatically enables ipx support even when pppd has it by default disabled at compile time and also runtime.

pali:
Quick test:
```
$ git clone https://github.com/ppp-project/ppp
Cloning into 'ppp'...
...
$ cd ppp
$ ./autogen.sh 
...
Features enabled
    Microsoft Extensions.: yes
    Multilink............: no
    Plugins..............: yes
    CBCP.................: no
    IPXCP................: no
    EAP-TLS..............: yes
    PEAP.................: yes
```

ipx is really disabled by default. See above output.

rpurdie:
We use the last released version. We added a patch allowing us to disable ipx as for the released version autotools isn't used and we can't make musl builds work without the patch. When a new release is made and we upgrade, we probably can use the configure option. I'm not sure who will actually be able to use that code on any modern system in the future hence this patch. If you want to keep it around, I'm fine with that, I just got the impression discussion was wanted on a patch to remove it. If that isn't the case sorry for wasting your time and mine.

Neustradamus:
@rpurdie: Can you rebase your PR after the Solaris build test change from @RICCIARDI-Adrien?

Neustradamus:
@rpurdie: @paulusmack has commented in the other ticket about it.

Can you rebase your PR, the Solaris test has been failed and the last PR of @RICCIARDI-Adrien solves it.

Thanks in advance.

enaess:
Looks good to me.

paulusmack:
I merged this, but I think I will add back a dummy noipx option that does nothing, just to avoid breaking anyone that has it in some script or in /etc/ppp/options.

PR #327, opened 2021-12-11
Fix Solaris CI build.

RICCIARDI-Adrien:
The Solaris GitHub action is sticking with macOS 10.15 because the new macOS 11 (that now replaces `macos-latest` image) does not contain the needed VirtualBox software. We will migrate the CI to use the latest macOS image when both the GitHub action and the macOS 11 image are ready.
See [actions/virtual-environments#4060](https://github.com/actions/virtual-environments/issues/4060) and [actions/virtual-environments#4010](https://github.com/actions/virtual-environments/pull/4010).

### DIFF
--- a/.github/workflows/solaris.yaml
+++ b/.github/workflows/solaris.yaml
@@ -3,12 +3,12 @@ on: [push, pull_request]
 
 jobs:
   solaris:
-    runs-on: macos-latest
+    runs-on: macos-10.15
     steps:
     - name: Checkout PPP sources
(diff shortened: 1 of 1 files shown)


RICCIARDI-Adrien:
The Solaris GitHub action is sticking with macOS 10.15 because the new macOS 11 (that replaces now `macos-latest` image) does not contain the needed software. We will migrate the CI to use the latest macOS image when the GitHub action is ready.

PR #334, opened 2022-01-18
Need to update the esp->ea_client.ea_namelen variable

enaess:
Need to be able to set the correct length of the username before calling chap/chams make_response.


enaess:
@bootc - I'll post patch separately for debian, but this patch would be very helpful if you cherry-picked

enaess:
Thanks @paulusmack

PR #342, opened 2022-05-16
pppd/lcp: lcp-up and lcp-down scripts.

jkroonza:


enaess:
@jkroonza - Could you please inform us of the use-case for this change?

enaess:
@pali I did checked out the issue, #343 and it is firewall related. I still don't see a particular reason to take it.

jkroonza:
Superceded by #367

PR #349, opened 2022-05-31
Cleaning up header files and updating defines

enaess:
Before I take this any further, I want to post a pull-request and invite people for review on this. 

Exporting pppd/pppd.h is the primary driver for this change, such that any project that picks up the header will know which features was included in the compile. 

Summary:
* Rename all of the features to reasonable defines such as PPP_WITH_FEATURE. This is to avoid overlap with other projects.
* Rely on the system's default includes for <linux/ppp_defs.h>, <linux/ppp-ioctls.h> and <linux/ppp-comp.h> instead of the included headers (which should be considered stale) -- though SunOS still relies on them, but again it also provides the driver for it too.
* Provide a <ppp/pppdconf.h> header that is exported and sets the correct defines, e.g. PPP_WITH_PEAP, etc. 
* Updates to the build system to drop the `${top_srcdir}/include`
* Standard Include guards for all headers


enaess:
@pali You have some time to look through these changes?

paulusmack:
What is the motivation for exporting pppd.h? I see that as an internal header for the pppd program (and for things that hook into it, i.e. plugins).

I think it would be reasonable to have another header that defines things like the exit values, which other programs could reasonably need to know.

enaess:
The question is, how far are we willing to go? 

Locking it down to a very concise API would be ideal, but likely impractical. There is a lot of third party software that depends on ppp project. Do we want to limit it completely and add features to the public API as needed? Or the other way around ?

paulusmack:
I don't want to be in a situation where we're constrained in what names we can use inside pppd by the fear of breaking other projects. Looking through pppd.h, the things that I consider other programs legitimately want to know about are:

- exit values
- configuration (whether various bits of optional functionality are enabled or disabled)
- paths (to scripts, config files, etc.)
- what syslog facility pppd uses

Anything else?

That seems like a small fraction, maybe 10%, of what is in pppd.h.

paulusmack:
A lot of the changes here seem to be about prepending PPP_ to various macro names (particularly HAVE_* and WITH_*). That seems fine to me.

I notice (somewhat at random) that we have a config option MAXOCTETS to enable a small amount of code, which defaults to doing nothing unless explicitly enabled. Maybe we don't need to be able to exclude that code.

enaess:
@paulusmack I'll look into the history of MAXOCTECTS to figure out the best venue here. Is that the only hold-up in merging this change?

enaess:
> I don't want to be in a situation where we're constrained in what names we can use inside pppd by the fear of breaking other projects. Looking through pppd.h, the things that I consider other programs legitimately want to know about are:
>
> * exit values
> * configuration (whether various bits of optional functionality are enabled or disabled)
> * paths (to scripts, config files, etc.)
> * what syslog facility pppd uses
> * Anything else?
> 
> That seems like a small fraction, maybe 10%, of what is in pppd.h.

I do agree with you. I'd like to see this as a direction to go. However, there is already projects depending on looking at some of these internal data structures and the current state of the ppp daemon. Here is one (of many) examples: https://gitlab.gnome.org/GNOME/network-manager-sstp/-/blob/master/src/nm-sstp-pppd-mppe.c

I did remove the PPP_WITH_MAXOCTETS as you requested. It only enables a small part of the code that otherwise wouldn't do anything unless it is enabled via a config option. 

I do have quite a number of changed tied up here and like to be able to continue working on other aspects like fixing openssl 3.0 support, plumbing the missing DES encryption, and others.

enaess:
@paulusmack any update on this?

enaess:
@paulusmack 

Also, I did send a fix upstream to the linux kernel mailing list. I believe you have commit powers for ppp stuff. Would you mind taking a look at this patch?

https://lkml.org/lkml/2022/5/31/635

enaess:
Thank you @paulusmack

enaess:
> ... the things that I consider other programs legitimately want to know about are:
> 
> * exit values
> * configuration (whether various bits of optional functionality are enabled or disabled)
> * paths (to scripts, config files, etc.)
> * what syslog facility pppd uses
> 
> Anything else?
> 
> That seems like a small fraction, maybe 10%, of what is in pppd.h.

I had another look at this. It's unfortunately more complicated than that. Plugins do actually look at the result of e.g. ipcp or ip6cp negotiations to figure out local / remote configuration, i.e. protocol entry structure needs to be exposed. As a result, the option structure must be exposed. Plugins do add their own options, so add_option() must be exposed.

What would be nice if the option related stuff was in e.g. options.h as a standalone separate include which is pulled in when needed (e.g. plugins). Leave the struct option as a forward declaration in pppd as protoent only needs a pointer to it. 

This isn't all of it, like other plugins depends on knowing the state of the signal handler (i.e. got_sigint(), etc). There is a few plugins that depends on a few of the CHAP related functions to be exposed. E.g. ChallengeAuthenticatiorResponse, and friends.

Again, the API is already broken from 2.4.9, so if the next release is 2.5, this would be the time to do this.

If we are going to do this, I think what we should do is to line up the most popular plugins for pppd and see where it breaks, if we break it, then we can submit a PR to that project with the new API, or incrementally move everything to pppd-priv.h which is only used by pppd compilation then add functions needed back into pppd.h.

What do you think @paulusmack

enaess:
@paulusmack To answer your question on what is needed in pppd.h (and moving everything else into pppd-private.h), I started the work here: 

https://github.com/enaess/ppp/blob/ppp-header/pppd/pppd.h

Notice the plugins in pppd/plugins/* still require a number of functions listed in this version of pppd.h between line 233 and 323. If I continue this work, is this something you'd still be willing to accept/merge?

PR #356, opened 2022-07-09
pppd: Fix check for incomplete ppp_stats nlmsgerr structure

pali:
Response structure is incomplete when returned length is less than required
structure length.

CC: @jkroonza

### DIFF
--- a/pppd/sys-linux.c
+++ b/pppd/sys-linux.c
@@ -1595,13 +1595,10 @@ get_ppp_stats_rtnetlink(int u, struct pppd_stats *stats)
     }
 
     if (nlresp.nlh.nlmsg_type == NLMSG_ERROR) {
-	if (nlresplen < offsetof(struct nlresp, __end_err)) {
-	    if (kernel_version >= KVERSION(4,7,0))
-		error("get_ppp_stats_rtnetlink: Netlink responded with error: %s (line %d)", strerror(-nlresp.nlerr.error), __LINE__);
-	} else {
-	    error("get_ppp_stats_rtnetlink: Netlink responded with an error message, but the nlmsgerr structure is incomplete (line %d).",
-		    __LINE__);
-	}
+	if (nlresplen < offsetof(struct nlresp, __end_err))
+	    error("get_ppp_stats_rtnetlink: Netlink responded with an error message, but the nlmsgerr structure is incomplete (line %d).", __LINE__);
+	else if (kernel_version >= KVERSION(4,7,0))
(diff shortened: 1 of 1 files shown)


jkroonza (changes requested):
What will kernels < 4.7.0 return?  If they will return NLMSG_ERROR along with something like ENXIO then we will output an error every time we pass by here (which should only be once or twice, but the reason for the check was specifically to avoid this).

How about a structure like:

```
if (nlresp.nlyh.nlmsg_type == NLMSG_ERROR) {
  if (kernel_version >= KVERSION(4,7,0)) {
    if (nlresplen < offsetof(struct nlresp, __end_err) {
      error("error but message incomplete error");
    } else {
      error("complete error message here");
    }
  }
  goto err;
}
```

pali:
I think that netlink should return complete error message on error even on older kernel versions. Check for length is there only in case of netlink kernel error. So reporting errors to log when incomplete message is received, is probably the correct way.

pppd/sys-linux.c
```diff
@@ -1596,11 +1596,11 @@ get_ppp_stats_rtnetlink(int u, struct pppd_stats *stats)
 
     if (nlresp.nlh.nlmsg_type == NLMSG_ERROR) {
 	if (nlresplen < offsetof(struct nlresp, __end_err)) {
-	    if (kernel_version >= KVERSION(4,7,0))
-		error("get_ppp_stats_rtnetlink: Netlink responded with error: %s (line %d)", strerror(-nlresp.nlerr.error), __LINE__);
-	} else {
 	    error("get_ppp_stats_rtnetlink: Netlink responded with an error message, but the nlmsgerr structure is incomplete (line %d).",
 		    __LINE__);
+	} else {
+	    if (kernel_version >= KVERSION(4,7,0))
```
jkroonza, line 1602: This can be } else if {
... no need for an additional layer of nesting.

If we know the errno on <= 4.7.0 perhaps we should rather test for that specicif errno and filter only that?

pali: I would not recommend to test on specific error returned by pre-4.7 kernels. Same errno code could be returned also in future for some other error. Codes are already too vague and used for many different errors (e.g. EINVAL).

jkroonza: That's fair.  I'm happy.  My one comment then is style related and it's much of a muchness.

jkroonza:
For what it's worth, looks good to me.

Neustradamus:
@paulusmack: Can you look?

Issue #339, opened 2022-04-06
Missing /var/lock directory


hellow554:
I have a custom linux distribution built with yocto. On that, I use systemd.

systemd deprecated the /var/lock directory and doesn't create is per default anymore: https://github.com/systemd/systemd/issues/15668

I had a similar issue with a different program dt-utils from the barebox project, here's a little chat between me and a yocto maintainer: https://bugzilla.yoctoproject.org/show_bug.cgi?id=14186

dt-utils accepted a patch (not by me) that changed the lock directory: https://git.pengutronix.de/cgit/tools/dt-utils/commit/src/barebox-state.c?id=fd48fe4efb40f35b119958495d56b399bec73654

My question: would you think creating a lock directory `/run/ppp/` which contains the lock files would be feasible?

I my systemd environment I would use [tempfiles](https://www.man7.org/linux/man-pages/man8/systemd-tmpfiles.8.html) to generate that directory. I just the question, what people without systemd can do? Should there be a check if that directory exists and if not creating it?

I would try to submit a patch for this, if you like. I just want to hear your opinion

paulusmack:
The lock directory location seems like something that should be configured via the configure script, i.e. perhaps you need to add a stanza to configure.ac. @enaess what do you think?

enaess:
I think it should be doable. Generally, it probably should be tied to one of the already defined directory variables, e.g. localstatedir. 

Though, I do need to better understand the problem first before making any rush recommendation for which variable ...


https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Installation-Directory-Variables.html

hellow554:
I think [`runstatedir`](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html) is a very good candiate:

> The directory for installing data files which the programs modify while they run, that pertain to one specific machine, and **which need not persist longer than the execution of the program**

(emphasis mine)

enaess:
Just to clarify the last comment there:

In autotools:
```
  --localstatedir=DIR     modifiable single-machine data [PREFIX/var]
  --runstatedir=DIR       modifiable per-process data [LOCALSTATEDIR/run]
```

The resulting location should be something along the lines of /usr/var/run unless the maintainer of the package overrides these. Typically, that would be done for a few variables, e.g. --prefix=/usr --sysconfdir=/etc --localstatedir=/var, then if the variable uses runstatedir, the resulting locations will be

* /etc/ppp/pppd.options
[...]

PR #358, opened 2022-07-30
Ppp autotools

enaess:
Fix for issue #339

This allows a user to configure the lock directory via configure variables.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -194,7 +194,7 @@ AS_IF([test "x$enable_openssl_engine" != "xno"], [],
 #
 # Specify runtime directory
 AC_ARG_WITH([plugin-dir],
-	AC_HELP_STRING([--with-plugin-dir=DIR], [Specify the plugin directory for pppd]))
+	AS_HELP_STRING([--with-plugin-dir=DIR],[Specify the plugin directory for pppd]))
 AS_IF([test -n "$with_plugin_dir"],
 	[PPPD_PLUGIN_DIR="$with_plugin_dir"],
 	[PPPD_PLUGIN_DIR="${libdir}/pppd/$VERSION"])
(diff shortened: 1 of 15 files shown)


enaess:
Just a note here. I am not thrilled that the PPP project introduces new variable names like _PATH_XYZ, the <paths.h> from the system already defines e.g. _PATH_VARLOG and others. Changes are that if a new path was added globally with the same name, we'd have to #undef it; technically I kind of wanted to change this to PPP_PATH_XYZ instead (or something similar) to explicitly state the variable came from the PPP project.

wongsyrone:
I think the `PPP_` prefix is more appropriate.

thesamesam:
> Just a note here. I am not thrilled that the PPP project introduces new variable names like _PATH_XYZ, the <paths.h> from the system already defines e.g. _PATH_VARLOG and others. Changes are that if a new path was added globally with the same name, we'd have to #undef it; technically I kind of wanted to change this to PPP_PATH_XYZ instead (or something similar) to explicitly state the variable came from the PPP project. Again, change here was based on the minimum amount of work needed. Let me know if this suffice, or a more extensive fix is needed.

Yeah, no more reserved names please: https://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html. It's far too fragile unfortunately.

thesamesam (changes requested):
lgtm once the reserved names are fixed

enaess:
OK, fixed up the path variables with a PPP_ prefix. Waiting for the CI checks to pass.

pppd/sha1.c
```diff
@@ -23,6 +23,8 @@
 #include <string.h>
 #include <time.h>
 #include <netinet/in.h>	/* htonl() */
+#include <sys/types.h>
```
wongsyrone, line 26: Why not put this line in sha1.h

wongsyrone: There are symbols cannot be found like https://github.com/ppp-project/ppp/blob/c0d579b/pppd/sha1.h#L25

enaess:
What is the error output you get when you compile?



pppd/sha1.c
enaess: I've dropped the change from the patch-set. Seems like the best way is to try to document what compile issue you are seeing and let us figure out the best solution. Please file a bug as @thesamesam suggested.

Generally, I am not a fan of putting includes in header files.

enaess:
Thanks @paulusmack

PR #360, opened 2022-08-04
pppd: Fix compilation on Linux when IPV6 is disabled

paulusmack:
This rearranges the PPP_WITH_IPV6CP guards added in commit
80b8744eb42c ("Changing INET6 to PPP_WITH_IPV6CP and adding configure
option", 2021-08-06) so that we (a) always include the rtnetlink
headers, since we need them for get_ppp_stats_rtnetlink(), and (b)
don't include eui64.h unless we have IPV6 support.

Fixes: 80b8744eb42c ("Changing INET6 to PPP_WITH_IPV6CP and adding configure option")
Signed-off-by: Paul Mackerras <paulus@ozlabs.org>

### DIFF
--- a/pppd/sys-linux.c
+++ b/pppd/sys-linux.c
@@ -128,8 +128,6 @@
 
 #include <linux/ppp-ioctl.h>
 
-
-#ifdef PPP_WITH_IPV6CP
 #include <linux/netlink.h>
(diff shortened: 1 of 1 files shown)


Neustradamus:
cc: @enaess.

enaess:
Looks good to me

PR #361, opened 2022-08-09
pppd: De-duplicate Linux rtnetlink code

pali:
Move common code for sending rtnetlink message into the new
send_rtnetlink_msg() function.

### DIFF
--- a/pppd/sys-linux.c
+++ b/pppd/sys-linux.c
@@ -300,108 +300,143 @@ extern int dfl_route_metric;
 
 
 /*
- * send_rtnetlink_msg - send rtnetlink message and return received error code:
+ * rtnetlink_msg - send rtnetlink message, receive response
+ * and return received error code:
  * 0              - success
  * positive value - error during sending / receiving message
  * negative value - rtnetlink responce error code
  */
-static int send_rtnetlink_msg(const char *desc, void *nlreq, size_t nlreq_len)
+static int rtnetlink_msg(const char *desc, int *shared_fd, void *nlreq, size_t nlreq_len, void *nlresp_data, size_t *nlresp_size, unsigned nlresp_type)
 {
-    struct {
+    struct nlresp_hdr {
         struct nlmsghdr nlh;
         struct nlmsgerr nlerr;
-    } nlresp;
+    } nlresp_hdr;
     struct sockaddr_nl nladdr;
-    struct iovec iov;
+    struct iovec iov[2];
     struct msghdr msg;
     ssize_t nlresp_len;
     int one;
     int fd;
 
-    fd = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
-    if (fd < 0) {
-        error("send_rtnetlink_msg: socket(NETLINK_ROUTE): %m (line %d)", __LINE__);
-        return 1;
-    }
+    if (shared_fd && *shared_fd >= 0) {
+        fd = *shared_fd;
+    } else {
+        fd = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
+        if (fd < 0) {
+            error("rtnetlink_msg: socket(NETLINK_ROUTE): %m (line %d)", __LINE__);
+            return 1;
+        }
 
-    /*
-     * Tell kernel to not send to us payload of acknowledgment error message.
-     * NETLINK_CAP_ACK option is supported since Linux kernel version 4.3 and
-     * older kernel versions always send full payload in acknowledgment netlink
-     * message. We ignore payload of this message as we need only error code,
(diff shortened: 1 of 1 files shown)


pali:
@jkroonza I added second commit into this PR which de-duplicate rtnetlink stats code. Could you please at it?

pppd/sys-linux.c
```diff
@@ -300,108 +300,143 @@ extern int dfl_route_metric;
...
+    } else {
+        iov[0].iov_len = sizeof(nlresp_hdr);
+    }
 
     memset(&msg, 0, sizeof(msg));
     msg.msg_name = &nladdr;
     msg.msg_namelen = sizeof(nladdr);
-    msg.msg_iov = &iov;
-    msg.msg_iovlen = 1;
+    msg.msg_iov = iov;
```
jkroonza, line 389: For example here, in other places you write &iov[0] ... it's the same thing from a pointer arithmetic perspective, but I suggest it be used consistently for the sake of readability, and &iov[0] is probably more clear about the intent (we want the address of element 0).

pppd/sys-linux.c
```diff
@@ -300,108 +300,143 @@ extern int dfl_route_metric;
...
 {
-    struct {
+    struct nlresp_hdr {
         struct nlmsghdr nlh;
         struct nlmsgerr nlerr;
-    } nlresp;
+    } nlresp_hdr;
     struct sockaddr_nl nladdr;
-    struct iovec iov;
+    struct iovec iov[2];
```
jkroonza, line 316: This variable is now used inconsistently in the code, sometimes iov is used, other times iov[i] ...

jkroonza (commented):
In principle this looks good and I'm in agreement.  I've only looked over the code, and I've not actually tested this right now, happy to assist with that if needed, but I don't have a test environment for this set up currently but can do so if needed.

pppd/sys-linux.c
```diff
@@ -300,108 +300,143 @@ extern int dfl_route_metric;
...
+        memset(&nladdr, 0, sizeof(nladdr));
+        nladdr.nl_family = AF_NETLINK;
 
-    if (bind(fd, (struct sockaddr *)&nladdr, sizeof(nladdr)) < 0) {
-        error("send_rtnetlink_msg: bind(AF_NETLINK): %m (line %d)", __LINE__);
-        close(fd);
-        return 1;
+        if (bind(fd, (struct sockaddr *)&nladdr, sizeof(nladdr)) < 0) {
+            error("rtnetlink_msg: bind(AF_NETLINK): %m (line %d)", __LINE__);
+            close(fd);
```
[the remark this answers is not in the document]
pali: This bind syscall is called only if either caller set shared_fd to NULL or caller set *shared_fd to < 0. So at this stage when closing fd due to failed bind syscall, caller cannot have (valid) fd with >= 0 value.

Now I see that it is quite harder to see what is this code doing because of how github show this diff. It is probably better to look into final file, not to diff.

pppd/sys-linux.c
pali: If `iov` is array and you want to reference first member of this array then it is better to write `&iov[0]` to explicitly express array context. If `iov` is just ordinary pointer to structure and you want a reference to this structure then it is better to write `iov` (and not `&iov[0]`) to not confuse people about array. Of course both construction should do same thing.

And I follow these patterns. That is why on the first place I used `&iov[0]` to explicitly pass just one member (indicated by `msg.msg_iovlen = 1;`) and on second place I used `iov` to explicitly express that more members (= array; indicated via larger and non-constant value in `msg.msg_iovlen`) is passed.

pppd/sys-linux.c
pali: I think after all changes, it should be consistent. `iov` is always array and if passing just one member of array I used `iov[i]` and when passing pointer (to array) then I used `iov`.

pali:
> In principle this looks good and I'm in agreement. I've only looked over the code, and I've not actually tested this right now, happy to assist with that if needed, but I don't have a test environment for this set up currently but can do so if needed.

I have tested second commit just by extracting whole rtnetlink code into separate file / executable and run it and checked that returned values are same as statistics reported by `ip -s link` output.

jkroonza:
> > In principle this looks good and I'm in agreement. I've only looked over the code, and I've not actually tested this right now, happy to assist with that if needed, but I don't have a test environment for this set up currently but can do so if needed.
> 
> I have tested second commit just by extracting whole rtnetlink code into separate file / executable and run it and checked that returned values are same as statistics reported by `ip -s link` output.

Good enough for me.  Thank you for your work on this.  I'm also quite happy with your other explanations, so from my side: lgtm.

Issue #242, opened 2021-01-11
openssl3.0.0 and ppp


fedya:
can't build 2.4.9 ppp with latest openssl 3.0.0
```
/usr/include/openssl/macros.h:60:52: note: expanded from macro 'OSSL_DEPRECATED'
#     define OSSL_DEPRECATED(since) __attribute__((deprecated))
        |       |       |       |       |       |  ^
eap-tls.c:520:13: error: function definition is not allowed here
        |   {
        |   ^
eap-tls.c:525:31: error: function definition is not allowed here
        |   int stub (UI* ui) {return 1;};
        |       |       |     ^
eap-tls.c:526:54: error: function definition is not allowed here
        |   int stub_reader (UI *ui, UI_STRING *uis) {return 1;};
        |       |       |       |       |       |    ^
eap-tls.c:528:49: error: use of undeclared identifier 'writer'; did you mean 'write'?
        |   UI_method_set_writer(transfer_pin,  writer);
        |       |       |       |       |       ^~~~~~
        |       |       |       |       |       write
/usr/include/unistd.h:367:16: note: 'write' declared here
extern ssize_t write (int __fd, const void *__buf, size_t __n) __wur
        |      ^
eap-tls.c:529:49: error: use of undeclared identifier 'stub'; did you mean 'stat'?
        |   UI_method_set_opener(transfer_pin,  stub);
        |       |       |       |       |       ^~~~
        |       |       |       |       |       stat
/usr/include/sys/stat.h:205:12: note: 'stat' declared here
extern int stat (const char *__restrict __file,
        |  ^
eap-tls.c:530:49: error: use of undeclared identifier 'stub'; did you mean 'stat'?
        |   UI_method_set_closer(transfer_pin,  stub);
        |       |       |       |       |       ^~~~
        |       |       |       |       |       stat
/usr/include/sys/stat.h:205:12: note: 'stat' declared here
extern int stat (const char *__restrict __file,
```

Neustradamus:
@jjkeijser, @enaess, @tisj: About OpenSSL 3.0 development branch...

tisj:
I tested this against openssl-3.0.0-alpha10, and built successfully with only a small modification (although there are deprecation warnings ofcourse, like with openssl-1.1.x):

```
diff -Nurp ppp-2.4.9.orig/pppd/Makefile ppp-2.4.9/pppd/Makefile
[...]

PR #362, opened 2022-08-10
Create an new API that allows us to abstract the use of the standard crypto functions

### LINKED ISSUE #333: auth.c:893: undefined reference to `crypt'
Hey there !

I'm not able to compile from source

```
./configure
make
```

auth.c:893: undefined reference to `crypt'

Is there a way to solve this issue ? 

Thanks a lot for your support !

enaess:
Create a new API to abstract the crypto functions used by pppd. This re-introduces the missing DES encryption functions copied from Openssl 3.0 project. Incorporates a new API for performing MD4/MD5/SHA and encryption using DES-ECB mode.

Unit tests are included for respective digest/encryption functions using this new API. With this change, you can pass configure --without-openssl to use the internally provided functions. If you do have openssl, then it will default to use these functions. This also provides a framework to allow other vendors to provide additional crypto implementations.

This closes #333, partially addresses #242 (except the pkcs11 engine support). Word has it that openssl is working on support for this, and the libp11 / opensc project are inclined not to do double work on this.


enaess:
@Neustradamus

Neustradamus:
@enaess: Good job!

@paulusmack: Can you look it?

enaess:
This change does pass the build checks, just that it ran into a timeout or something.

pppd/chap-md5.c
```diff
@@ -63,27 +63,43 @@ chap_md5_verify_response(int id, char *name,
...
-
-		/* Test if our hash matches the peer's response */
-		if (memcmp(hash, response, MD5_HASH_SIZE) == 0) {
-			slprintf(message, message_space, "Access granted");
-			return 1;
-		}
+        PPP_MD_CTX* ctx = PPP_MD_CTX_new();
+        if (ctx) {
+
+            if (PPP_DigestInit(ctx, PPP_md5())) {
```
thesamesam, line 80: Whitespace is inconsistent here (and the blank lines look a bit odd IMO)

thesamesam (commented):
I do wonder if it's worth paying some of the cost here and if OpenSSL shouldn't just become a hard dependency.

pppd/chap-md5.c
enaess: I'll fix up the white space inconsistencies

enaess: There is a case for disabling the include of openssl (embedded developers). Also, openssl can be configured not to include certain encryption algorithms and thus we would need to provide our own. Then there is the few other projects (I would guess embedded, that would want to use something else than openssl). The API is modeled after openssl and can easily be search and replaced. So there is that ...

enaess:
@thesamsam The history of the crypto functions in pppd is a bit muddied (we've had a lot of problems, e.g. symbols with same name and linking with OpenSSL caused crashes, etc). 

There is a case for disabling the include of openssl (embedded developers) which also disables eap-tls and peap, but if you are only using pppd for dial-up applications, you likely won't need that. 

Also, OpenSSL can be configured exclude certain encryption algorithms and thus we would need to provide our own. The configure script should detect that. Then there is the few other projects (I would guess embedded, that would want to use something else than openssl to obtain a smaller footprint). 

The API is modeled after OpenSSL and can easily be search and replaced, or if someone wants WolfSSL then they can add their own ifdefs and implementation in a much more non-intrusive manner. Yeah, I agree the change is a bit intrusive, but it is that way for a reason.

paulusmack:
Looks OK to me, but I would like to hear any comments that @carlsonj might have.

carlsonj:
On 8/18/22 19:37, Paul Mackerras wrote:
> Looks OK to me, but I would like to hear any comments that @carlsonj
> <https://github.com/carlsonj> might have.

I'm not sure, but I'm guessing you're referring to this:

https://github.com/ppp-project/ppp/commit/c2b7ad0bb5119ef6a1ada58469a14b1e4070e941

If so, then it looks like the return value from ChallengeResponse() in
chap_ms.c has changed. It seems to have been a flag (0 for failure, 1
for success), and now it's just a hard-coded 0.  If that value is needed
for some reason, I think you might want something like this:

    if (DesEncrypt(challenge, ZPasswordHash + 0,  response + 0) == 1 &&
        DesEncrypt(challenge, ZPasswordHash + 7,  response + 8) == 1 &&
        DesEncrypt(challenge, ZPasswordHash + 14, response + 16) == 1)
      retval = 1;

But that's just a guess. In the pppd code I currently have available to
me, ChallengeResponse() returns void, not int. So I have no idea at all
what that retval is about or who introduced it. It's a bit of a surprise
to me. I'm guessing that this is on some modified branch, but I don't
really understand what's going on.

I'm a little baffled by the changes to the unit tests in ppp-crypto.c. I
had thought those "expected" values came from some authoritative source.
At a minimum, it'd be nice to have comments here explaining the origin
of the values. I'd hope they're not just retrained to whatever the code
now happens to produce. (But I honestly don't know; I'm not the author
there.)

I have no way of knowing whether the changes in ppp-des.c are right. I'm
not familiar with the interfaces involved. But I'd certainly suggest
testing a modified pppd against the existing version very carefully for
interoperability.

-- 
James Carlson     42.703N 71.076W FN42lq08    ***@***.***>

enaess:
@carlsonj 

With regards to ChallengeResponse, it used to returned a void value, I changed it to a integer representing a 0 or 1; the return value isn't used anywhere (but if it could be in the future). OpenSSL 3.0 unless you initialize the legacy crypto sub-system, the DES encryption will fail; this would be a indication that something is busted further down the stack. This code went through several iterations, and I have no removed the unneeded variables (sorry for the confusion). Also, the use of bool should probably be replaced with using the #include <stdbool.h> over typedef bool to unsigned char in pppd.h. Your example could very well be without the == 1 comparison.

```
    if (DesEncrypt(challenge, ZPasswordHash + 0,  response + 0) &&
        DesEncrypt(challenge, ZPasswordHash + 7,  response + 8) &&
        DesEncrypt(challenge, ZPasswordHash + 14, response + 16))
        return 1
```

The main problem with the previous incarnation of the DesEncrypt/DesDecrypt functions were that they required a global variable representing the key-schedule and repeated calls to MakeKey(). This has now been abstracted out in favor of keeping it on the stack and making the code re-entrant.

Overall, this is probably a larger change where you'd need to pull down my code so you can see the bigger picture (a diff makes a review a bit unorganized). Also, the DesEncrypt/DesDecrypt functions with it's strange way of setting up the Key for the crypto operation has been decoupled from the actual encryption operation, i.e. RFC2759 stuff all happens here, while the actual DES ECB encryption happens in ppp-des.c. This makes it a clean cut separation where one can setup unit-tests for RFC2759 encryption in pppcrypt.c while actually testing DES ECB encryption in an isolated environment of ppp-des.c

Frankly, I wasn't able to find a good authoritative source of DES methods w.r.t. test ciphers. NIST declared DES unsecure already back in 1995. And there isn't any FIPS test vectors readily available for this. If you have one, I can plug-in the test vectors and add the comments. Right now, it's a side-by-side comparison with generating the same input / output data with openssl. Same thing with the digests. 

Now, there exist a unit-test for the chap subsection of the code, and it all includes the DES/SHA/MD5/MD4 algorithms, and those test vectors are taken from the respective RFCs (also the one for DES encryption in pppcrypt.c). I once broke the MD4 implementation when using OpenSSL, and added these unit-tests as a result, and they has been instrumental in catching these flaws. Right now, there are 4 test-suites in the areas of chap, pppcrypt, ppp-crypto, and peap.

I do concur with you that we need to be extra careful with testing pppd to ensure we don't have any regressions. The unit-tests aren't full system tests (but they also weren't meant to be). While I see @Neustradamus is rallying up distro-maintainers to submit their patches (PR) in order for a new release, I would probably want to ask for 1 - 2 month code lock-down and test. Give any dependent projects some time to craft patches to work with the new pppd 2.5.0.

enaess:
Also, if it wasn't clear; I've done basic testing by connecting to a client to a Win2K22 server using mschapv2, eap-mschap-v2, and eap-md5. Note that the Microsoft does make it hard(er) to test the legacy chap-md5 and mschap protocols. Though, running through the mschapv2 authentication does test sha1,md4 and des encryption. In addition, the eap-md5 should cover the md5 algorithm. This was using both built-in and by using openssl. 

I'd still give this change a go ...

enaess:
@paulusmack any chance we can merge this?

enaess:
Thanks @paulusmack

PR #366, opened 2022-08-18
pppd: Retry registering interface when on rtnetlink -EBUSY error

pali:
Due to workaround in kernel module ppp_generic.ko in function
ppp_nl_newlink(), kernel may return -EBUSY error to prevent possible
mutex deadlock. In this case userspace needs to retry its request.

Proper way would be to fix kernel module to order requests and mutex
locking, so prevent deadlock in kernel and so never return this error to
userspace. Until it happens we need retry code in userspace.


pali:
Here is link to that ppp_nl_newlink() kernel function and -EBUSY logic:
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/drivers/net/ppp/ppp_generic.c?h=v5.19#n1304

PR #370, opened 2022-09-19
Fix libsystemd detection

Chocobo1:
1. Use uppercase for `prefix` parameter
  `SYSTEMD_CFLAGS` is used elsewhere so `prefix` cannot be lowercase.
  https://autotools.info/pkgconfig/pkg_check_modules.html

2. The module name should be `libsystemd`
   Previously it will result in the following compile error when building pppd/auth.c:
   > /usr/bin/ld: pppd-auth.o: undefined reference to symbol 'sd_notify@@LIBSYSTEMD_209'
   > /usr/bin/ld: /usr/lib/libsystemd.so.0: error adding symbols: DSO missing from command line

   This is due to missing `-lsystemd-daemon` flag which is provided by `libsystemd-daemon-devel` package on Debian or `systemd-libs` on ArchLinux. And the proper .pc file in the package is `libsystemd` not `systemd`.
   https://stackoverflow.com/a/38303241

### DIFF
--- a/.github/workflows/ubuntu.yaml
+++ b/.github/workflows/ubuntu.yaml
@@ -5,13 +5,17 @@ jobs:
   build_and_test:
     runs-on: ubuntu-latest
     env:
-      configure_flags: --enable-ipxcp --enable-multilink
+      configure_flags: --enable-ipxcp --enable-multilink --enable-systemd
 
     steps:
     - uses: actions/checkout@v2
 
     - name: install required packages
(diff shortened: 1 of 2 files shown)


Chocobo1:
@paulusmack 
Please take a look.

paulusmack:
@enaess what do you think of this?

enaess:
Looks like a good fix to me. Typically my workflow sets this up automatically (e.g. apt-get build-dep ppp, or other package) and the dependent files just exist for me. There is in fact, two packages (pkg-config entries) on Ubuntu systemd and libsystemd. I must have grabbed the wrong one.

paulusmack:
I'll take this, despite the bogus signoff, since it is all pretty trivial.

PR #373, opened 2022-09-29
pppd: add noresolvconf option

kurt-vd:
On embedded systems, it is easier to avoid the resolv.conf conflicts than to fix them.
'noresolvconf' option on top of 'usepeerdns' lets pppd request DNS information and forward the info to the scripts, but will not write resolv.conf. The ip-up/ip-down scripts can then deal with the info and no conflict ever existed.

Signed-off-by: Kurt Van Dijck <dev.kurt@vandijck-laurijssen.be>

### DIFF
--- a/pppd/ipcp.c
+++ b/pppd/ipcp.c
@@ -93,6 +93,7 @@ struct notifier *ip_down_notifier = NULL;
 static int default_route_set[NUM_PPP];	/* Have set up a default route */
 static int proxy_arp_set[NUM_PPP];	/* Have created proxy arp entry */
 static bool usepeerdns;			/* Ask peer for DNS addrs */
+static bool resolvconf = 1;		/* deploy peer DNS addrs */
 static bool usepeerwins;		/* Ask peer for WINS addrs */
 static int ipcp_is_up;			/* have called np_up() */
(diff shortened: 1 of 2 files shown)


paulusmack:
The desire to avoid conflicts over /etc/resolv.conf is exactly why pppd already doesn't write to it; it writes to /etc/ppp/resolv.conf (note different path) and lets the ip-up script or something else figure out what to do with it, whether to copy it to /etc/resolv.conf, merge entries in, or whatever else.

So why do we need to stop pppd writing /etc/ppp/resolv.conf? What's the conflict with that file, as distinct from /etc/resolv.conf?

kurt-vd:
I see. I'm mistaken here indeed.

I use buildroot and had not spotted there fix behind `BR2_PACKAGE_PPPD_OVERWRITE_RESOLV_CONF` that replaces `/etc/ppp/resolv.conf` by `/etc/resolv.conf`

kurt-vd:
If I become annoyed by `pppd`'s errors `Failed to create /etc/ppp/resolv.conf: Read-only filesystem`, then I'll create a more accurate fix for that specifically

PR #377, opened 2022-12-12
Remove a few vestiges of IPX support

paulusmack:
Signed-off-by: Paul Mackerras <paulus@ozlabs.org>


Neustradamus:
@enaess: What do you think?

enaess:
I already did this in my PR. Yes, 👍

paulusmack:
Abandoning this now because the ubuntu.yaml change is in another PR and the README change can be done when it is updated for the forthcoming release.

PR #378, opened 2022-12-16
radius: distinguish between User-Request and Admin-Reset.

jkroonza:
For the purposes of our definition:

User-Request - remote side hanging up.
Admin-Reset - local side hanging up.

Reasoning is that typically radius will be used to authentication dial-in users, so if the pppd gets killed locally, that's not the User (client) requesting hangup, but rather the local administrator (be that a manual kill, or as a result of a CoA/Disconnect).

### DIFF
--- a/pppd/plugins/radius/radius.c
+++ b/pppd/plugins/radius/radius.c
@@ -1051,8 +1051,11 @@ radius_acct_stop(void)
     av_type = PW_NAS_ERROR;
     switch( status ) {
 	case EXIT_OK:
-	case EXIT_USER_REQUEST:
 	    av_type = PW_USER_REQUEST;
+	    break;
+
+	case EXIT_USER_REQUEST:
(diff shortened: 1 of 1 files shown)


Neustradamus:
@paulusmack, @enaess: What do you think?

enaess:
I'd be okay with this change, though my header file reorg will have a conflict as I touched this area too. I'd suggest either I add it to my PR or we take this PR after mine

jkroonza:
happy to perform the rebase once ther other PR has been merged, please let me know.

enaess:
I actually did a cherry-pick into my ppp-header branch. If / when that get merged, it will have your fix you committed on that branch.

https://github.com/enaess/ppp/tree/ppp-header

enaess:
Hrm, okay. PR379 also have this commit; I hope git is smart enough to figure that out.

Issue #389, opened 2023-02-15
cppcheck: assorted findings "known condition true or false"


chipitsine:
```
common/zlib.c:3839:13: style: Condition 't<19' is always false [knownConditionTrueFalse]
modules/ppp.c:1502:14: style: Condition 'pass<0' is always false [knownConditionTrueFalse]
modules/ppp.c:1509:16: style: Condition 'pass' is always true [knownConditionTrueFalse]
pppd/auth.c:712:9: style: Condition '!mp_on()' is always true [knownConditionTrueFalse]
pppd/auth.c:723:20: style: Condition '!mp_on()' is always true [knownConditionTrueFalse]
pppd/auth.c:740:9: style: Condition '!mp_on()' is always true [knownConditionTrueFalse]
pppd/auth.c:780:9: style: Condition '!mp_on()' is always true [knownConditionTrueFalse]
pppd/auth.c:827:9: style: Condition '!mp_on()' is always true [knownConditionTrueFalse]
pppd/crypto.c:491:9: style: Condition '!PPP_crypto_deinit()' is always false [knownConditionTrueFalse]
pppd/main.c:316:9: style: Condition 'ifname' is always true [knownConditionTrueFalse]
pppd/main.c:1198:22: style: Condition 'error_count!=errs' is always false [knownConditionTrueFalse]
pppd/main.c:1216:22: style: Condition 'error_count!=errs' is always false [knownConditionTrueFalse]
pppd/main.c:1244:9: style: Condition '!mp_on()' is always true [knownConditionTrueFalse]
pppd/plugins/radius/config.c:207:42: style: Condition '*p=='\0'' is always false [knownConditionTrueFalse]
pppd/plugins/radius/radius.c:748:25: style: Same expression on both sides of '&&' because 'mppe_enc_policy' and 'mppe_enc_keys' represent the same value. [knownConditionTrueFalse]
pppd/sys-linux.c:803:13: style: Condition '!mp_on()' is always true [knownConditionTrueFalse]
pppd/sys-linux.c:3557:13: style: Condition 'sfd<0' is always true [knownConditionTrueFalse]
pppdump/zlib.c:3368:13: style: Condition 't<19' is always false [knownConditionTrueFalse]
solaris/ppp.c:1515:14: style: Condition 'pass<0' is always false [knownConditionTrueFalse]
solaris/ppp.c:1522:16: style: Condition 'pass' is always true [knownConditionTrueFalse]
```

paulusmack:
[...]

PR #396, opened 2023-02-20
Should check name, not ifname in ppp_set_ifname

enaess:
This surfaces in discussion of issue #389


enaess:
Thank you! @paulusmack

Issue #392, opened 2023-02-15
utest_chap: buffer overflow found by address sanitizer


chipitsine:
tests built and ran with

CFLAGS='-fsanitize=address -ggdb'
LDFLAGS='-fsanitize=address'

./pppd/utest_chap

```
=================================================================
==230115==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffc98d210a5 at pc 0x000000409b53 bp 0x7ffc98d20f20 sp 0x7ffc98d20f18
READ of size 1 at 0x7ffc98d210a5 thread T0
    #0 0x409b52 in Get7Bits /home/ilia/ppp/pppd/crypto_ms.c:101
    #1 0x409ddf in MakeKey /home/ilia/ppp/pppd/crypto_ms.c:120
    #2 0x409f2d in DesEncrypt /home/ilia/ppp/pppd/crypto_ms.c:137
    #3 0x40370e in ChallengeResponse /home/ilia/ppp/pppd/chap_ms.c:525
    #4 0x403dc6 in ChapMS_NT /home/ilia/ppp/pppd/chap_ms.c:622
    #5 0x4047d1 in ChapMS /home/ilia/ppp/pppd/chap_ms.c:810
    #6 0x404d23 in test_chap_v1 /home/ilia/ppp/pppd/chap_ms.c:938
    #7 0x40518e in main /home/ilia/ppp/pppd/chap_ms.c:978
    #8 0x7fbc1944a50f in __libc_start_call_main (/lib64/libc.so.6+0x2750f)
    #9 0x7fbc1944a5c8 in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x275c8)
    #10 0x402504 in _start (/home/ilia/ppp/pppd/utest_chap+0x402504)

Address 0x7ffc98d210a5 is located in stack of thread T0 at offset 53 in frame
    #0 0x403586 in ChallengeResponse /home/ilia/ppp/pppd/chap_ms.c:511

  This frame has 1 object(s):
    [32, 53) 'ZPasswordHash' (line 512) <== Memory access at offset 53 overflows this variable
HINT: this may be a false positive if your program uses some custom stack unwind mechanism, swapcontext or vfork
      (longjmp and C++ exceptions *are* supported)
SUMMARY: AddressSanitizer: stack-buffer-overflow /home/ilia/ppp/pppd/crypto_ms.c:101 in Get7Bits
Shadow bytes around the buggy address:
[...]

PR #399, opened 2023-02-27
CI: add automatic tests, enable ASAN build&test

chipitsine:
1. libpam is installed which enables it when running "configure"
2. cbcp is enabled
3. build step for running automatic tests added, i.e. "make check"
4. one more configuration is added, with ASAN (address sanitizer) enabled. it will prevent bugs like https://github.com/ppp-project/ppp/issues/392 (until proper fix is merged, ASAN job will fail)


Neustradamus:
@RICCIARDI-Adrien: What do you think?

enaess:
@chipitsine Could you please add some more notes to the PR description, just to summarize what the changes do?

chipitsine:
I added descr few days ago.

however, maybe I should add inline comment ? people rarely read PR description

enaess:
@chipitsine I think it is okay. It's just that merging changes here are limited by @paulusmack 's time and availability.

Neustradamus:
@paulusmack: Have you looked?

Issue #405, opened 2023-03-05
Add a configure option to support the system default's CA path


enaess:
Typical configuration is to specify 
`capath <path-to-ca-dir>`

A bunch of distributions comes with the default ca path for a lot of applications to use. We should expose a configure options that allows package maintainers to simply set the right path (by default it would be ${prefix}/etc/ssl/certs, or ${sysconfdir}/ssl/certs. 

@jjkeijser

PR #406, opened 2023-03-05
Fix for issue #405, this adds a --with-system-ca-path to configure script

enaess:
It allows package maintainers to set the default location for the ca directory.


enaess:
@jjkeijser I updated the change, please have another look

jjkeijser:
the new version looks fine to me - just a matter of "defensive programming"

enaess:
The Solaris check timed out. Its not an indicator of failure.

PR #412, opened 2023-04-12
Move pppgetpass.8 to EXTRA_DIST

floppym:
automake does not seem to include noinst_*_MANS in the dist tarball.


enaess:
Looks good.

PR #415, opened 2023-04-19
Prefer size_t over int where possible

AZero13:
This avoids needless truncation, which takes up multiple more instructions on some architectures, especially older ones.

### DIFF
--- a/chat/chat.c
+++ b/chat/chat.c
@@ -178,12 +178,14 @@ struct termios saved_tty_parameters;
 
 char *abort_string[MAX_ABORTS], *fail_reason = (char *)0,
 	fail_buffer[50];
-int n_aborts = 0, abort_next = 0, timeout_next = 0, echo_next = 0;
+unsigned long  n_aborts = 0;
+int abort_next = 0, timeout_next = 0, echo_next = 0;
 int clear_abort_next = 0;
 
 char *report_string[MAX_REPORTS] ;
 char  report_buffer[4096] ;
-int n_reports = 0, report_next = 0, report_gathering = 0 ; 
+unsigned long  n_reports = 0;
+int report_next = 0, report_gathering = 0;
 int clear_report_next = 0;
 
 int say_next = 0, hup_next = 0;
@@ -606,9 +608,9 @@ void terminate(int status)
  * Allow the last of the report string to be gathered before we terminate.
  */
     if (report_gathering) {
-	int c, rep_len;
+	int c;
 
-	rep_len = strlen(report_buffer);
+	size_t rep_len = strlen(report_buffer);
 	while (rep_len + 1 < sizeof(report_buffer)) {
 	    alarm(1);
 	    c = get_char();
@@ -1342,7 +1344,8 @@ int echo_stderr(int n)
 int get_string(register char *string)
 {
     char temp[STR_LEN];
-    int c, printed = 0, len, minlen;
+    int c, printed = 0;
+    size_t len, minlen;
     register char *s = temp, *end = s + STR_LEN;
     char *s1, *logged = temp;
 
@@ -1372,7 +1375,7 @@ int get_string(register char *string)
     alarmed = 0;
 
     while ( ! alarmed && (c = get_char()) >= 0) {
-	int n, abort_len, report_len;
+	size_t n, abort_len, report_len;
 
 	if (echo) {
 	    if (echo_stderr(c) != 0) {
@@ -1421,8 +1424,8 @@ int get_string(register char *string)
 	}
 	else {
 	    if (!iscntrl (c)) {
-		int rep_len = strlen (report_buffer);
-		if ((rep_len + 1) < sizeof(report_buffer)) {
+		size_t rep_len = strlen (report_buffer);
+		if (rep_len < sizeof(report_buffer) - 1) {
 		    report_buffer[rep_len]     = c;
 		    report_buffer[rep_len + 1] = '\0';
 		}
--- a/chat/chat.c
+++ b/chat/chat.c
@@ -630,9 +630,9 @@ void terminate(int status)
  * Allow the last of the report string to be gathered before we terminate.
  */
     if (report_gathering) {
-	int c, rep_len;
+	int c;
 
-	rep_len = strlen(report_buffer);
+	size_t rep_len = strlen(report_buffer);
 	while (rep_len + 1 < sizeof(report_buffer)) {
 	    alarm(1);
 	    c = get_char();
@@ -1036,7 +1036,7 @@ int chat_send (register char *s)
 
 	if (strcmp(s, "OFF") == 0)
 	    sa.sa_handler = SIG_IGN;
-        else
+	else
 	    sa.sa_handler = sighup;
 	sigaction(SIGHUP, &sa, NULL);
         return 0;
@@ -1366,7 +1366,8 @@ int echo_stderr(int n)
 int get_string(register char *string)
 {
     char temp[STR_LEN];
-    int c, printed = 0, len, minlen;
+    int c, printed = 0;
+    size_t len, minlen;
     register char *s = temp, *end = s + STR_LEN;
     char *s1, *logged = temp;
 
@@ -1396,7 +1397,8 @@ int get_string(register char *string)
     alarmed = 0;
 
     while ( ! alarmed && (c = get_char()) >= 0) {
-	int n, abort_len, report_len;
+	int n;
+	size_t abort_len, report_len;
 
 	if (echo) {
 	    if (echo_stderr(c) != 0) {
@@ -1445,8 +1447,8 @@ int get_string(register char *string)
 	}
 	else {
 	    if (!iscntrl (c)) {
-		int rep_len = strlen (report_buffer);
-		if ((rep_len + 1) < sizeof(report_buffer)) {
+		size_t rep_len = strlen (report_buffer);
+		if (rep_len < sizeof(report_buffer) - 1) {
 		    report_buffer[rep_len]     = c;
 		    report_buffer[rep_len + 1] = '\0';
 		}
--- a/pppd/auth.c
+++ b/pppd/auth.c
@@ -501,7 +501,7 @@ static int
 setupapfile(char **argv)
 {
     FILE *ufile;
-    int l;
+    size_t l;
     uid_t euid;
     char u[MAXNAMELEN], p[MAXSECRETLEN];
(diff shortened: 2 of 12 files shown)


enaess:
These fixes look mostly good. Though, some cases the refactoring changes the structure ...

pppd/auth.c
```diff
@@ -2154,22 +2155,24 @@ int
...
     /* Allow all if no authorization list. */
     if (!wp)
 	return 1;
 
     /* Allow if we have a match in the authorization list. */
-    while (wp) {
+    do {
 	/* trailing '*' wildcard */
 	l = strlen(wp->word);
+	if (l == 0)
```
enaess, line 2168: Changing from do to while, and checking the length ... should be fine...

pppd/plugins/pppoatm/atm.h
```diff
@@ -103,6 +103,6 @@ int sap2text(char *buffer,int length,const struct atm_sap *sap,int flags);
 int sap_equal(const struct atm_sap *a,const struct atm_sap *b,int flags,...);
 
 int __t2q_get_rate(const char **text,int up);
-int __atmlib_fetch(const char **pos,...); /* internal use only */
+size_t __atmlib_fetch(const char **pos,...); /* internal use only */
```
enaess, line 106: Should be fine, but a possibly oops here: you can install libatm-dev (on debian) and have pppd's plugin compile against that instead of using the code-drop from the library here.

Is this fix also in the upstream version of libatm-dev?

pppd/tls.c
```diff
@@ -186,12 +186,13 @@ static int tls_verify_callback(int ok, X509_STORE_CTX *ctx)
...
         /* Match the suffix of common name */
         if (!strcmp(TLS_VERIFY_SUFFIX, tls_verify_method)) {
-            int len = strlen(ptr1);
-            int off = strlen(cn_str) - len;
+            size_t len1, len2;
             ptr2 = cn_str;
-            if (off > 0) {
-                ptr2 = cn_str + off;
-            }
+
```
enaess, line 191: Did you test this change? 

As am scrolling through, reviewing this to ensure the functionality have not changed is a bit hard. Though the change make the code look cleaner, does it still work?

AtariDreams: Yes, absolutely.

paulusmack:
I have mixed feelings about these changes. It's reasonable to use size_t for things that are the length of a structure or a string, but I don't think that the savings in code size will be significant either in space or time. There's also a lot of changes spread out across the source tree which could be better as separate commits.

AtariDreams:
> I have mixed feelings about these changes. It's reasonable to use size_t for things that are the length of a structure or a string, but I don't think that the savings in code size will be significant either in space or time. There's also a lot of changes spread out across the source tree which could be better as separate commits.

I think they would be best done at once, that's why, since they all have the same theme.

Neustradamus:
@AtariDreams: It is not possible to separate?

paulusmack:
I notice also that the signoff is a noreply.github.com address, which I don't consider adequate.

Neustradamus:
@AtariDreams: Have you seen the @paulusmack comment?
What is the situation with your signoff?

AtariDreams:
@Neustradamus Fixed!

paulusmack (commented):
In general it seems like there is a bunch of refactoring here plus gratuitous whitespace changes, in addition to the changes mentioned in the commit description (i.e. converting int to size_t). So, at the very least the commit description needs to be more comprehensive. If nothing else it should identify any potential bugs fixed here, though in fact I would prefer the bug fixes to be in separate commits.

I'm not convinced by the rationale that we need to do this to reduce the generated code size. As far as I can see it is only going to reduce the code size on 64-bit machines, not 32-bit machines, some of which might arguably be a bit short on memory. If you have a 64-bit processor you probably have a GB or more of memory, and saving at most a few hundred bytes of instruction memory is going to be in the noise.

Neustradamus:
@AtariDreams: Have you seen @paulusmack comments started:
- https://github.com/ppp-project/ppp/pull/415#pullrequestreview-1795621594

Note: The other PR has been merged :)

Neustradamus:
@AtariDreams: Have you looked comments from @paulusmack?

AtariDreams:
I'll take a look.

Neustradamus:
@AtariDreams: Have you looked?
It will be good to have a solution before 2.5.1 release...

paulusmack:
As far as I can see, there is no functional change here (or at least, none intended), so I am not going to wait for this.

chat/chat.c
```diff
@@ -1421,8 +1424,8 @@ int get_string(register char *string)
 	}
 	else {
 	    if (!iscntrl (c)) {
-		int rep_len = strlen (report_buffer);
-		if ((rep_len + 1) < sizeof(report_buffer)) {
+		size_t rep_len = strlen (report_buffer);
+		if (rep_len < sizeof(report_buffer) - 1) {
```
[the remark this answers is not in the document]
AtariDreams: It just makes more sense.

pppd/plugins/pppoatm/atm.h
AtariDreams: I was unaware of this.

Neustradamus:
@paulusmack: Have you seen the @AtariDreams changes?

pppd/main.c
```diff
@@ -1863,9 +1863,9 @@ update_script_environment(void)
     struct userenv *uep;
 
     for (uep = userenv_list; uep != NULL; uep = uep->ue_next) {
-	int i;
+	size_t i;
```
paulusmack, line 1866: For a case like this where the variable is a loop index rather than the size or length of something, I'd prefer plain long or unsigned long rather than size_t. Same comment applies in a few other places.

pppd/plugins/pppoatm/misc.c
```diff
@@ -37,7 +37,7 @@ int __atmlib_fetch(const char **pos,...)
 	}
     }
     va_end(ap);
-    if (best > -1) (*pos) += best_len;
+    (*pos) += best_len;
```
paulusmack, line 40: This is a semantic change; previously the function would return -1 if the string at *pos didn't match any of the argument strings at all, whereas now it will return 0, which callers interpret as a match with the second argument. So now text2qos will parse things incorrectly.

This is an example of why I would have preferred this commit to be split into smaller pieces, for example, four separate commits for chat, pppd, and the pppoatm and pppoe plugins. If you had, I could merge the commits that are OK. As it is, I can't merge any of it.

paulusmack (changes requested):

paulusmack:
Overall, I would take everything except the pppoatm changes, which don't seem to me to be at all necessary, and in fact wrong, as I commented.

shahbaz606 (commented):
l

Neustradamus:
@AreaZR: Have you seen the latest @paulusmack comments?

Neustradamus:
@AZero13: Your PR will be not finished before next version? :/

Neustradamus:
@AZero13: Have you seen the latest @paulusmack comments?

AZero13:
Yes I will address this now

Neustradamus:
@AZero13: I have sent you e-mails about your PR.

Can you look your PR?

Thanks in advance.

AZero13:
Okay, okay okay I will. Yes I will do this

AZero13:
I fixed the issues! Apologies that it took so long!

pppd/main.c
```diff
@@ -2184,8 +2184,10 @@ ppp_script_setenv(char *var, char *value, int iskey)
     char *p, *newstring;
 
     newstring = (char *) malloc(vl+1);
-    if (newstring == 0)
+    if (newstring == 0 || vl > INT_MAX) {
```
jkroonza: Please clarify.

pppd/options.c
```diff
@@ -1673,6 +1675,8 @@ callfile(char **argv)
     }
 
     l = strlen(arg) + strlen(PPP_PATH_PEERFILES) + 1;
+    if (l > INT_MAX)
```
jkroonza: This does NOT look like a sensible way to protect against counter overflows.

pppd/options.c
```diff
@@ -1873,9 +1877,8 @@ user_setenv(char **argv)
 	return 0;
     }
     for (uep = userenv_list; uep != NULL; uep = uep->ue_next) {
-	int nlen = strlen(uep->ue_name);
-	if (nlen == (eqp - arg) &&
-	    strncmp(arg, uep->ue_name, nlen) == 0)
+	size_t nlen = strlen(uep->ue_name);
+	if (nlen == (eqp - arg) && strncmp(arg, uep->ue_name, nlen) == 0)
```
jkroonza: This changes a line without changing it.  I don't object, but my rule of thumb is that unless you're touching the code in a line anyway, don't reformat it either.

pppd/tls.c
```diff
@@ -186,12 +186,13 @@ static int tls_verify_callback(int ok, X509_STORE_CTX *ctx)
...
+            len1 = strlen(ptr1);
+            len2 = strlen(ptr2);
```
jkroonza, line 193: This is a logical breaking change.  There is no chance that this would work correctly.  Specifically tls_verify_method would need to be TLS_VERIFY_SUFFIX, at which point ptr2 won't be initialised by either of the previous two if statements.

strlen(NULL) will segfault.

pppd/tls.c
```diff
@@ -186,12 +186,13 @@ static int tls_verify_callback(int ok, X509_STORE_CTX *ctx)
...
+            if (len2 > len1)
+                ptr2 += len2 - len1;
         }
 
         if (strcmp(ptr1, ptr2)) {
```
jkroonza, line 198: Unrelated to the change, but what happens if tls_verify_method doesn't match any of the if blocks above?  ptr2 == NULL => segfault.

jkroonza (changes requested):

jkroonza:
I saw there was a push, last push seems to change further behaviour not really improving/restoring original behaviour, but I'll need to re-read more thoroughly, but last I checked it essentially broke the behaviour and I highly doubt the last push would restore the original behaviour.

jkroonza (approved):
I did recheck, and I'm no longer seeing functional changes, so from my perspective, this looks good now.

Neustradamus:
@AZero13: Never too late, thanks for your work and thanks to @paulusmack for merging!

PR #416, opened 2023-04-20
Fix calloc calls

AZero13:
Size and number are switched in calloc sometimes. This PR fixes that.

### DIFF
--- a/pppd/tdb.c
+++ b/pppd/tdb.c
@@ -933,7 +933,7 @@ static int tdb_new_database(TDB_CONTEXT *tdb, int hash_size)
 
 	/* We make it up in memory, then write it out if not internal */
 	size = sizeof(struct tdb_header) + (hash_size+1)*sizeof(tdb_off);
-	if (!(newdb = calloc(size, 1)))
+	if (!(newdb = calloc(1, size)))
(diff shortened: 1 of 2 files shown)


enaess:
These fixes look good.

enaess:
We should take this, code as is is using the calloc() api wrongly.

paulusmack:
Hmmm... I would take it if the signoff wasn't a noreply.github.com address. Maybe it's trivial enough that that doesn't matter. In any case there is no functional problem here, given that the arguments are multiplied together and multiplication of integers is commutative.

Neustradamus:
@AtariDreams: Have you seen the @paulusmack comment?
What is the situation with your signoff?

Neustradamus:
@paulusmack: Have you seen the change from @AtariDreams?

paulusmack:
@Neustradamus it's exactly the same change with the same noreply signoff, just rebased.

Neustradamus:
@AtariDreams: Can you add the correct sign-off?

Neustradamus:
@enaess: Can you look to do like previously?
Create a PR and sign the commit?

Neustradamus:
@paulusmack: You can not publish the PR directly with a note about @AtariDreams?

AtariDreams:
Fixed!

PR #418, opened 2023-04-25
create new function "PPPOECloseDiscoverySock", and close DiscoverySock after "the_channel->establish_ppp"

pgson:
Before doing add_fd(ppp_dev_fd) (line 698) in sys-linux.c after PADS, LCP: Conf-Request from the PPP server arrives first and times out. To address this, I moved the close discoverySock() after the cut-off of DiscoverySock by the_channel->establish_ppp to improve the processing.

### DIFF
--- a/pppd/auth.c
+++ b/pppd/auth.c
@@ -663,6 +663,8 @@ void start_link(int unit)
 	goto disconnect;
     }
 
+    the_channel->close_discovery_sock();
+
     if (!demand && ifunit >= 0)
 	set_ifunit(1);
 
--- a/pppd/pppd.h
+++ b/pppd/pppd.h
@@ -235,6 +235,8 @@ struct channel {
 	void (*send_config)(int, uint32_t, int, int);
 	/* set the receive-side PPP parameters of the channel */
 	void (*recv_config)(int, uint32_t, int, int);
+	/* close discovery sock */
+	void (*close_discovery_sock)(void);
 	/* cleanup on error or normal exit */
 	void (*cleanup)(void);
 	/* close the device, called in children after fork */
--- a/pppd/plugins/pppoe/discovery.c
+++ b/pppd/plugins/pppoe/discovery.c
@@ -727,8 +727,6 @@ discovery2(PPPoEConnection *conn)
     }
 
     /* We're done. */
-    close(conn->discoverySocket);
-    conn->discoverySocket = -1;
     conn->discoveryState = STATE_SESSION;
     return;
 }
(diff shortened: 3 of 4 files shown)


pgson (commented):
review request

pgson:
add Signed-off-by.  please review.

pgson (commented):
please review

Neustradamus:

pppd/pppd.h
```diff
@@ -235,6 +235,8 @@ struct channel {
 	void (*send_config)(int, uint32_t, int, int);
 	/* set the receive-side PPP parameters of the channel */
 	void (*recv_config)(int, uint32_t, int, int);
+	/* close discovery sock */
+	void (*close_discovery_sock)(void);
```
enaess, line 239: Why go to the length of adding another callback? It could seem like registering a "cleanup" function would be suficient.

pppd/auth.c
```diff
@@ -663,6 +663,8 @@ void start_link(int unit)
 	goto disconnect;
     }
 
+    the_channel->close_discovery_sock();
```
paulusmack, line 666: This will crash for any channel type other than pppoe. You need to check if the_channel->close_discovery_sock is NULL.

paulusmack:
I'm not sure why the DCO check is failing; the signoff looks fine to me.

paulusmack:
Regarding the commit - the commit description is not sufficient. It needs at least to have the details in the first comment above in the commit message. It would be even better if more details were given to explain exactly what the sequence of events is and why the proposed change makes it work better.

I don't really like having something in the generic code (auth.c) that is so specific to one particular channel type. Given that you add the close_discovery_sock call immediately after the establish_ppp call, you could accomplish the same effect by making pppoe have its own establish_ppp function which calls generic_establish_ppp and then closes the discovery socket. But I'm really not clear on why keeping the discovery socket open for longer fixes anything (and if it does then the commit message needs to explain why it does).

Neustradamus:
@pgson: Have you seen @paulusmack and @enaess comments?

Neustradamus:
@pgson: Have you seen @paulusmack and @enaess comments?

Neustradamus:
@pgson: We wait you...

Can you look here: https://github.com/ppp-project/ppp/pull/418/files?

Thanks in advance.

Neustradamus:
@pgson: Have you seen @paulusmack and @enaess comments?

paulusmack:
Closing because of lack of response to review comments.

Neustradamus:
@pgson: Are you alive?

Can you look comments on your PR?

Thanks in advance.

PR #420, opened 2023-04-29
passwordfd: read password during option processing

floppym:
When configured to detach from the controlling terminal, pppd closes file descriptors 0, 1, and 2 before the passwd hook is called. If the user passes 0, 1, or 2 to the passwordfd option, pppd will fail to read the password.

To work around this, treat passwordfd as a special option and read the password during option processing, before pppd closes it.

Bug: https://bugs.gentoo.org/209294


Neustradamus:

PR #421, opened 2023-05-02
Replace ax_check_openssl with serial 11 from autoconf-archive

floppym:
The current macro checks ssldirs before calling pkg-config. This causes a build failure on multilib-enabled systems, since /usr/lib is searched before /usr/lib64.

The new macro calls pkg-config first, and only falls back to searching ssldirs if that fails.

Bug: https://bugs.gentoo.org/905442

### DIFF
--- a/m4/ax_check_openssl.m4
+++ b/m4/ax_check_openssl.m4
@@ -1,5 +1,5 @@
 # ===========================================================================
-#     http://www.gnu.org/software/autoconf-archive/ax_check_openssl.html
+#     https://www.gnu.org/software/autoconf-archive/ax_check_openssl.html
 # ===========================================================================
 #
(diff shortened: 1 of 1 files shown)


floppym:
Hmm, upon further inspection, it looks like ppp's copy of ax_check_openssl.m4 has been customized to accept "yes" and "no" as valid arguments for `--with-openssl`. The pristine copy from autoconf-archive rejects these values.

Do you want to maintain this customization?

floppym:
The ax_check_openssl.m4 code is actually pretty broken. Closing this until I can submit a better alternative.

Neustradamus:
@floppym: Do not hesitate to link your better alternative ^^

PR #422, opened 2023-05-07
Implement logging the LCP RTT

rfc1036:
This patch adds the lcp-rtt-file configuration option, which instructs pppd to add a timestamp to the data section of each LCP echo request frame and then log their round trip time and any detected packet loss to a circular buffer in that file.

Other programs then can asynchronously read the file and report statistics about the line.

I have also added examples of a simple program which dumps the content of the log file and a Prometheus exporter which can be run as a CGI.

This feature was inspired by a similar one in a commercial PPP implementation, but I have never actually seen that and I do not know how it actually works or it was implemented.

Signed-off-by: Marco d'Itri <md@linux.it>

### DIFF
--- a/pppd/lcp.c
+++ b/pppd/lcp.c
@@ -47,6 +47,11 @@
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
+#include <fcntl.h>
+#include <string.h>
+#include <time.h>
+#include <arpa/inet.h>
+#include <sys/mman.h>
 
 #include "pppd-private.h"
 #include "options.h"
@@ -68,11 +73,21 @@
 static void lcp_delayed_up(void *);
 
 /*
+ * These definitions relate to the measurement and logging of round-trip
+ * time (RTT) of LCP echo-requests implemented in lcp_rtt_update_buffer().
+ */
+#define LCP_RTT_MAGIC 0x19450425
+#define LCP_RTT_HEADER_LENGTH 4
+#define LCP_RTT_FILE_SIZE 8192
+#define LCP_RTT_ELEMENTS (LCP_RTT_FILE_SIZE / sizeof(u_int32_t) - LCP_RTT_HEADER_LENGTH) / 2
+
+/*
  * LCP-related command-line options.
  */
 int	lcp_echo_interval = 0; 	/* Interval between LCP echo-requests */
 int	lcp_echo_fails = 0;	/* Tolerance to unanswered echo-requests */
 bool	lcp_echo_adaptive = 0;	/* request echo only if the link was idle */
+char	*lcp_rtt_file = NULL;	/* measure the RTT of LCP echo-requests */
 bool	lax_recv = 0;		/* accept control chars in asyncmap */
 bool	noendpoint = 0;		/* don't send/accept endpoint discriminator */
 
@@ -152,6 +167,9 @@ static struct option lcp_option_list[] = {
       "Set time in seconds between LCP echo requests", OPT_PRIO },
     { "lcp-echo-adaptive", o_bool, &lcp_echo_adaptive,
       "Suppress LCP echo requests if traffic was received", 1 },
+    { "lcp-rtt-file", o_string, &lcp_rtt_file,
+      "Filename for logging the round-trip time of LCP echo requests",
+      OPT_PRIO | OPT_PRIV },
     { "lcp-restart", o_int, &lcp_fsm[0].timeouttime,
       "Set time in seconds between LCP retransmissions", OPT_PRIO },
     { "lcp-max-terminate", o_int, &lcp_fsm[0].maxtermtransmits,
@@ -197,6 +215,8 @@ lcp_options lcp_hisoptions[NUM_PPP];	/* Options that we ack'd */
 static int lcp_echos_pending = 0;	/* Number of outstanding echo msgs */
 static int lcp_echo_number   = 0;	/* ID number of next echo frame */
 static int lcp_echo_timer_running = 0;  /* set if a timer is running */
+static int lcp_rtt_file_fd = 0;		/* fd for the opened LCP RTT file */
+static volatile u_int32_t *lcp_rtt_buffer = NULL; /* the mmap'ed LCP RTT file */
 
 static u_char nak_buffer[PPP_MRU];	/* where we construct a nak packet */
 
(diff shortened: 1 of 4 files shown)


Neustradamus:

enaess:
Just because a third-party vendor of a commercial ppp implementation does this, doesn't make it a requirement for pppd to do so. First question I have is, is this feature documented in a RFC some place; in which I'd be totally fine with taking this as long as it meets the RFC's specs. 

I am aware of LQP or Link Quality Protocol as described in https://datatracker.ietf.org/doc/html/rfc1333 which is obsoleted by https://datatracker.ietf.org/doc/html/rfc1989.

Where is the "specs" for this change coming from?

rfc1036:
As explained, I have invented and implemented everything from scratch except than taking inspiration from the general idea of measuring the RTT of LCP echo requests.

LQP has a different purpose, semantics and mechanics: it requires to be implemented by both sides and negotiated using LCP, for a start.

enaess:
Let me see if I understand it correctly. This change doesn't change the underlying principals of the PPP/LCP protocol,  but allows one to monitor the RTT of LCP echo / reply packets. It adds a few extra tools so you can monitor and store these in a binary database?

With my knowledge of the LCP protocol, choosing a "magic" number to track may be a bad idea. These are negotiated at random and also prevent loop detection. You may want to change the implementation to measure the negotiated value of his or peer magic within the echo req/rep packets. Unless, my understanding of this is wrong ...

rfc1036:
Yes, I think that you are misunderstanding it. This works with no cooperation at all from the remote, because the standard requires that the complete optional payload of the LCP echo request frame is returned as is in the echo reply. So the magic number is really only relevant in the context of the peer sending and then receiving this data, not of the remote. It could even be argued it is almost superfluous.

pppd sends as usual the LCP echo requests when configured to do so, but attaches a timestamp to each one. When it gets the timestamp back in the LCP echo replies then it will comput the RTT and write it to a ring buffer which can be read by monitoring systems.

rfc1036:
I do not understand why you mention "cloud" and "CGI servers" and what you think that this has to do with this feature.
The first candidates to use this are systems like OpenWRT which implement dashboards with statistics about the connection.

enaess:
Don't worry, you have my vote. Also, I am not the one who merges changes ;-)

pppd/lcp.c
```diff
@@ -67,12 +72,18 @@
 
 static void lcp_delayed_up(void *);
 
+#define LCP_RTT_MAGIC 0x19450425
```
paulusmack, line 75: I haven't seen anywhere where you have said "round trip time" rather than "RTT". If the reader doesn't happen to be familiar with that acronym, this code could be somewhat obscure. I would like to see a comment here (or some other suitable place) explaining that "These definitions relate to the measurement and logging of round trip time (RTT) of LCP echo-requests" (or something similar). That way the acronym is expanded in at least one place.

pppd/lcp.c
```diff
@@ -152,6 +163,8 @@ static struct option lcp_option_list[] = {
       "Set time in seconds between LCP echo requests", OPT_PRIO },
     { "lcp-echo-adaptive", o_bool, &lcp_echo_adaptive,
       "Suppress LCP echo requests if traffic was received", 1 },
+    { "lcp-rtt-file", o_string, &lcp_rtt_file,
+      "Write to this file the RTT of LCP echo requests", OPT_PRIO },
```
paulusmack, line 167: Suggest "Filename for logging the round-trip time of LCP echo requests" or similar.

pppd/lcp.c
```diff
@@ -2225,6 +2240,61 @@ LcpEchoTimeout (void *arg)
...
+    lost = lcp_echos_pending - 1;
+    if (lost > 0xFF)
+	lost &= 0xFF;		/* truncate the lost packets to 256 */
+    if (rtt > 0xFFFFFF)
+	rtt &= 0xFFFFFF;	/* truncate the RTT to 16777216 */
+    /* use bits 24-31 for the lost packets and bits 0-23 for the RTT */
+    ring_buffer[next_entry + 1] = htonl((u_int32_t) ((lost << 24) + rtt));
+
+    /* update the pointer to the (just updated) most current data element */
+    lcp_rtt_buffer[2] = htonl(next_entry);
```
paulusmack, line 2292: Normally with a mmaped ring buffer one should really have some sort of memory barrier here (i.e. before the write to lcp_rtt_buffer[2]), since otherwise, on architectures with a weakly-consistent memory model, a consumer that has mmaped the ring buffer might see the write to lcp_rtt_buffer[2] before the write to the ring buffer entry. If all consumers are simply going to read the whole buffer with the read() syscall, and we assume that read() is atomic with respect to msync(), then perhaps there isn't actually a problem. I don't know of portable constructs for memory barriers short of using a pthread mutex, and arguably this data is not so critical that an occasional bad value is not a real problem, so perhaps it is reasonable not to have any barrier. But at least a comment may be useful.

pppd/lcp.c
paulusmack, line 167: Should this be a privileged option? If it isn't then we need to be careful when opening the file, or restrict its location.

pppd/lcp.c
```diff
@@ -2293,11 +2387,68 @@ LcpSendEchoRequest (fsm *f)
...
     }
 }
 
+static void
+lcp_rtt_open_file (void)
+{
+    if (!lcp_rtt_file)
+	return;
+
+    lcp_rtt_file_fd = open(lcp_rtt_file, O_RDWR | O_CREAT, 0644);
```
paulusmack, line 2412: We either need to do seteuid() before and after this open() call, or else make the option privileged.

pppd/lcp.c
```diff
@@ -67,12 +72,18 @@
 
 static void lcp_delayed_up(void *);
 
+#define LCP_RTT_MAGIC 0x19450425
+#define LCP_RTT_HEADER_LENGTH 4
+#define LCP_RTT_FILE_SIZE 8192
```
paulusmack, line 77: There are Linux machines which use a 64kB page size, so we need to either make this a multiple of 64kB or else use getpagesize() in lcp_rtt_open_file(). (Hopefully Solaris has getpagesize().)

pppd/pppd.8
```diff
@@ -637,6 +637,9 @@ Set the maximum number of LCP terminate-request transmissions to
 Set the LCP restart interval (retransmission timeout) to \fIn\fR
 seconds (default 3).
 .TP
+.B lcp\-rtt\-file \fIfilename
+Sets the file where the RTT of LCP echo-request frames will be logged.
```
paulusmack, line 641: Since this is documentation, please change "the RTT" to "the round-trip time (RTT)".

paulusmack (changes requested):

Neustradamus:
@rfc1036: Have you seen @paulusmack comments?

rfc1036:
Yes, but I am on vacation with no access to my testbed, I will get back to you next month.

pppd/lcp.c
rfc1036: Why this? The Linux mmap(2) man page explicitly describes what happens when the length is not a multiple of the page size, and I could not find anything forbidding this in the Solaris man page either.

pppd/lcp.c
```diff
@@ -2225,6 +2240,61 @@ LcpEchoTimeout (void *arg)
...
+ * containing a pair of u_int32_t in network byte order with this content:
+ * [0] UNIX timestamp
+ * [1] bits 24-31: the number of lost LCP echo replies
+ *     bits 0-23:  the measured RTT in microseconds
+ *
+ * Consumers of lcp_rtt_file are expected to:
+ * - read the complete file of arbitrary lenght
+ * - check the magic number
+ * - process the data elements starting at the index
+ * - discard the last data element, i.e. the one preceding the element
```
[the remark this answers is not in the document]
rfc1036: There was a purpose but it was wrong. :-) I have removed this code.

pppd/lcp.c
```diff
@@ -2225,6 +2240,61 @@ LcpEchoTimeout (void *arg)
...
+    else
+	next_entry = ntohl(lcp_rtt_buffer[2]) + 2;	/* use the next one */
+
+    /* update the data element */
+    /* storing the timestamp in an *unsigned* long allows dates up to 2106 */
+    ring_buffer[next_entry] = htonl((u_int32_t) time(NULL));
+    lost = lcp_echos_pending - 1;
+    if (lost > 0xFF)
+	lost &= 0xFF;		/* truncate the lost packets to 256 */
+    if (rtt > 0xFFFFFF)
```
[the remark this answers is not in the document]
rfc1036: Yes, that was my intention: I have changed the code.

I choose microseconds because it is the natural unit which fits in three octets, and I thought that an LCP latency of over 16 seconds would be highly uncommon.

If you believe that it useful to represent higher latencies then I can easily rework the patch to use tens of microseconds instead.

pppd/lcp.c
rfc1036: That was my reasoning as well. I have added a comment.

If this will show to not be true in practice then I have thought of also setting the timestamp of the oldest sample to zero (i.e. the one which would be overwritten the next cycle): this way even if the index is read out of order then no wrong sample is read and the correct one will still be picked up by the reader the next time.

rfc1036:
I have updated the as discussed in the review.

pppd/lcp.c
paulusmack: You're right, it's OK if length is less than the page size, so ignore this comment.

pppd/lcp.c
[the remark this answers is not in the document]
paulusmack: Generally speaking it would be rare to get latencies greater than 16.7 seconds, but it could happen in pathological situations. I guess we can leave it like this for now and change it later if necessary.

PR #425, opened 2023-05-27
pppd/options.c: handle malloc failure

chipitsine:


Neustradamus:

enaess:
Looks fine to me, but if malloc fails the state of the system is probably not going to make much of a difference by returning 1.

chipitsine:
This one is not recoverable, but returning error is better than segfault

paulusmack:
The more usual thing would be to call novm("environment variable") or some such.

Neustradamus:
@paulusmack: @chipitsine has done changes, what do you think?

PR #427, opened 2023-05-28
Ensure there is a '/' between PPP_PATH_VARRUN and the PID filename

floppym:
Bug: https://bugs.gentoo.org/907311


Neustradamus:

Issue #429, opened 2023-06-08
enable_microsoft_extensions => openssl


icegood:
Because of this line:
https://github.com/ppp-project/ppp/blob/master/pppd/crypto_ms.c#L125

ms extensions should imply openssl. Ideally, should not be depended

Neustradamus:
@paulusmack, @enaess: Have you seen this ticket?

enaess:
Just opened a PR for this. Yeah, remove the boogie men ...

Neustradamus:
Thanks @icegood for this ticket and @enaess for the PR :)

I hope a comeback of @paulusmack soon!

PR #431, opened 2023-06-14
Fixes issue #429, stray include of an openssl header was removed

enaess:
Must have been something I forgot to remove when I was working on this code. After this change, you pppd compiles without openssl as long as you add the following to the configure script: ./configure --disable-peap --disable-eaptls --without-openssl

### DIFF
--- a/pppd/crypto_ms.c
+++ b/pppd/crypto_ms.c
@@ -122,8 +122,6 @@ MakeKey(const unsigned char *key, unsigned char *des_key)
 	DES_set_odd_parity((DES_cblock *)des_key);
 }
 
-#include <openssl/evp.h>
-
 int
 DesEncrypt(const unsigned char *clear, const unsigned char *key, unsigned char *cipher)
 {


Neustradamus:

Neustradamus:
@paulusmack: @enaess has done a PR, can you look?

paulusmack:
One day @enaess will write a decent commit message... :)

I *think* the following is an explanation of what this is all about, but please check and correct.  Proposed commit message:

-----------------------------------------------------------------
pppd: Fix compilation with openssl disabled

If openssl is disabled at configure time but microsoft extensions are enabled,
we get a compilation error due to an unnecessary include. This removes the
unnecessary include.  With this, pppd compiles without openssl as long as you
add the following arguments to the configure script invocation:

        --disable-peap --disable-eaptls --without-openssl

Fixes: https://github.com/ppp-project/ppp/issues/429
-----------------------------------------------------------------

enaess:
Done

PR #432, opened 2023-06-18
pppd/ppp-sha1.c: use uint32_t instead of u_int32_t

bkuhls:
Fixes build with musl-libc toolchains.


Neustradamus:

enaess:
Change looks good to me

enaess:
I am good with this change. I've tried to eliminate the use of u_int32 in public headers. This should come through too.

Issue #419, opened 2023-04-29
lock directory moved in ppp-2.5.0


floppym:
With ppp-2.4.9, serial device lock files were always created in `/var/lock`.

Since 66a8c74c3f73d7480a09923a225b56b8829ae790, the lock files get created in `PPP_PATH_VARRUN "/lock"`, which evaluates to a value like `/var/run/pppd/lock` or `/run/pppd/lock` depending on the value of `runstatedir` passed to configure.

Is this change of location intentional? There are other programs that use serial ports that expect the lock files to be in /var/lock.

cc @enaess

paulusmack:
It wasn't intentional. I suppose some people didn't realize that the lock file is not intended to be private to pppd and other people weren't sufficiently vigilant.

What do distros do these days about coordinating serial port access between different programs? Do modern distros even know what serial ports are? :)

floppym:
On Gentoo at least, /var/lock is a symlink to /run/lock, for compatibility with "legacy" applications.

I would guess this is the case on most systemd-based distros. When sysv compatibility is enabled, systemd installs a tmpfiles drop-in that creates the /run/lock directory and the /var/lock symlink.

https://github.com/systemd/systemd/blob/v253/tmpfiles.d/legacy.conf.in#L13-L14

A couple of alternatives on Linux:

1. Use `ioctl(2)` with `TIOCEXCL` to enable exclusive mode on the tty device, and `TIOCGEXCL` to respect it as root.
2. Use advisory locks on the tty device via `flock(2)`.

enaess:
Q: Looks like the default value for runtime directory was set to /var/run/pppd, this seems to only manage where pppd2.tdb and the lock directory. Would you be able to specify `--with-runtime-dir=/var/run` or just `/run` as the value achieve the same functionality. 

You are probably right that there something buggy in this code. The lock directory should be the same (/var/lock) but not imply the pppd run directory. In this case, only pppd2.tdb would be in this directory (/run/pppd/pppd2.tdb).

floppym:
I think it makes more sense to keep runtime-dir set to `${runstatedir}/pppd`. That path is also used for PID files in main.c.

I am currently working around the lock directory issue with this tmpfiles.d drop-in which creates a symlink from /run/pppd/lock to /run/lock.

[...]

PR #435, opened 2023-08-03
Make lock dir configurable

martinetd:
lock dir changed on linux from /var/lock to /run/pppd/lock with pppd-2.5.0, which makes pppd fail to start if the distribution does not pre-create the directory.

This reverts it back to /var/lock.
    

~~In the line of configurability, also make the lock dir configurable through ./configure --with-lock-dir for distributions that prefer /run/pppd/lock, but change the default for linux back to /var/lock~~

~~This commit is bigger than it should be, because the current configure mechanism for PPPD_RUNTIME_DIR and others is inconditional: pathnames.h has #else cases for when the variable is not defined but that is not possible.~~
~~This does not work for lock dir, because on non-linux platform the path is different, and reproducing this logic at configure time is not trivial.~~

~~Instead of the current method, only set the the PPPD_X_DIR variables in pppdconf.h if it was explicitly set in configure, and otherwise leave it to the else case.~~

~~Note:~~
- ~~the else cases were incorrect for runtime dir and log dirs, and have been fixed~~
- ~~PPPD_PLUGIN_DIR has been kept as is because it is used for rpath in makefiles and needs to be set inconditionnaly at configure time~~

Fixes: 66a8c74c3f73 ("Let ./configure control the paths for pppd")
Fixes: #419

### DIFF
--- a/pppd/Makefile.am
+++ b/pppd/Makefile.am
@@ -83,7 +83,7 @@ pppd_SOURCES = \
     upap.c \
     utils.c
 
-pppd_CPPFLAGS = -DSYSCONFDIR=\"${sysconfdir}\" -DLOCALSTATEDIR=\"${localstatedir}\" -DPPPD_RUNTIME_DIR='"@PPPD_RUNTIME_DIR@"' -DPPPD_LOGFILE_DIR='"@PPPD_LOGFILE_DIR@"'
+pppd_CPPFLAGS = -DSYSCONFDIR=\"${sysconfdir}\" -DPPPD_RUNTIME_DIR='"@PPPD_RUNTIME_DIR@"' -DPPPD_LOGFILE_DIR='"@PPPD_LOGFILE_DIR@"'
 pppd_LDFLAGS =
 pppd_LIBS =
 
--- a/pppd/pathnames.h
+++ b/pppd/pathnames.h
@@ -120,12 +120,12 @@
 #define PPP_PATH_PPPDB          PPP_PATH_VARRUN  "/pppd2.tdb"
 
 #ifdef __linux__
-#define PPP_PATH_LOCKDIR        PPP_PATH_VARRUN  "/lock"
+#define PPP_PATH_LOCKDIR        "/var/lock"
 #else
 #ifdef SVR4
-#define PPP_PATH_LOCKDIR        LOCALSTATEDIR "/spool/locks"
+#define PPP_PATH_LOCKDIR        "/var/spool/locks"
 #else
-#define PPP_PATH_LOCKDIR        LOCALSTATEDIR "/spool/lock"
+#define PPP_PATH_LOCKDIR        "/var/spool/lock"
 #endif
 #endif
 


martinetd:
(ah, note the default for localstatedir is /usr/local/var so that'll make the default lock dir /usr/local/var/lock and not /var/lock, but distros will configure `--localstatedir=/var` and it'll work out in practice... Happy to override this to not use localstatedir if you prefer)

martinetd:
fixed make check (had to add `LOCALSTATEDIR` define to tests cflags as well now as they use `PPP_PATH_LOCKDIR` which now requires `LOCALSTATEDIR` for its default value)

paulusmack:
It looks fine to me, but I would like to hear what @enaess thinks of it (and of #436).

Neustradamus:
@paulusmack: @enaess has done some comments in the other PR recently.
Not in this PR, have you seen?

rfc1036:
I do not think that the locks directory should be configurable: `/var/lock/` is part of the Linux ABI and there is no reason to change it. People with non-standard file system schemes can use local non-standard patches.

martinetd:
Happy to hard-code it back to /var/lock then, I can update this PR ~end of next week if that's preferred; that should simplify this patch quite a bit.

martinetd:
Sorry for the delay I had forgotten to update this to hardcode back the lock path.

I know the standard path is /var/lock, but since we've always been using /var/run/lock I've preferred to revert back to the old path instead of using /var/lock: happy to change it to /var/lock, there's just no standard define for it in paths.h

Also, my original commit had a drive-by fix: "the else cases were incorrect for runtime dir and log dirs, and have been fixed"
This has also been dropped, but can be found in https://github.com/martinetd/ppp/commits/lock_dir_configurable if anyone cares.

paulusmack:
Actually, in ppp-2.4.9 and prior, the lock directory is /var/lock on Linux or /var/spool/locks on Solaris. The lock directory changed in commit 66a8c74c3f73 ("Let ./configure control the paths for pppd", 2022-07-30) to PPP_PATH_LOCKDIR set in pathnames.h to /var/run/lock.

On the whole I would rather we went back to using /var/lock.

martinetd:
Oh, sorry I didn't go far enough.
We had `#define LOCK_DIR    "/var/lock"` indeed; I've reverted back to this.

That also made me notice the non-standard LOCALSTATEDIR define we use again which should indeed be /var, but since it looks like the general consensus is to not make it configurable I've reverted back to explicit /var for all OSes and removed the now-unused variable. (Happy to use LOCALSTATEDIR instead if that's better, just say)

martinetd:
Thanks!

martinetd:
Hmm, sorry for the double-notification to everyone following this, but just a head's up github did something a bit weird here: it apparently rewrote the from of the merged commit?

https://github.com/martinetd/ppp/commit/f8382246ef4a601a7983c6f6946948fe15e4dd70 (original commit that I submitted/was reviewed) had From with my work e-mail (@atmark-techno.com) and doesn't have a co-authored-by;
the commit that was merged https://github.com/ppp-project/ppp/commit/99cbf5e269994482edaf64624be8b1c806f9587c has a from with my personal email (@codewreck.org) and got a co-authored-by added with the work mail... This is a bit weird given there's only one person involved.

(I'm not expecting this to be fixed at this point, but that also means the commit doesn't have a sign off from its "author")

I've never seen this before so not sure what repo setting is involved here, but someone might want to have a look...
(assuming github did this because the commiter of the commit is marked `GitHub <noreply@github.com>`; please ignore this if it was intended)

PR #438, opened 2023-08-06
Fix linking error with lld linkers

listout:
When using lld linker, build fails with

ld.lld: error: /usr/lib/gcc/x86_64-pc-linux-gnu/12/../../../../lib64/Scrt1.o is incompatible with elf32-i386
ld.lld: error: /usr/lib/gcc/x86_64-pc-linux-gnu/12/../../../../lib64/crti.o is incompatible with elf32-i386
ld.lld: error: /usr/lib/llvm/16/bin/../../../../lib/clang/16/lib/linux

The fix is to check pkg-config first, and not force manual -L /usr/lib. If pkg-config succeeded, then we don't bother with -L /usr/lib

Our guess is this what the actual intention was based upon the coments

if pkg-config is installed and openssl has installed a .pc file, then use that information and don't search ssldirs

First found on gentoo linux with llvm profile, please check out Bug: section of the commit for more info and a complete build log.

Bug: https://bugs.gentoo.org/905442

### DIFF
--- a/m4/ax_check_openssl.m4
+++ b/m4/ax_check_openssl.m4
@@ -55,6 +55,20 @@ AC_DEFUN([AX_CHECK_OPENSSL], [
         ])
 
     AS_IF([test "${with_openssl}" != "no"], [
+        # if pkg-config is installed and openssl has installed a .pc file,
+        # then use that information and don't search ssldirs
+        AC_PATH_PROG([PKG_CONFIG], [pkg-config])
+        if test x"$PKG_CONFIG" != x""; then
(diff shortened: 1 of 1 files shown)


Neustradamus:

paulusmack:
@enaess how does this look to you?

enaess:
Fix should be okay.

paulusmack:
Could somebody work out why this breaks the solaris build?

Neustradamus:
@carlsonj: Can you look for Solaris ?

PR #441, opened 2023-08-17
pppd/options.c: fix memory leak or error path

chipitsine:
found by Coverity

```
602err:
603    fclose(f);
604    privileged_option = oldpriv;
   CID 436193 (#1 of 1): Resource leak (RESOURCE_LEAK)10. overwrite_var: Overwriting option_source in option_source = oldsource leaks the storage that option_source points to.
605    option_source = oldsource;
```

PR #444, opened 2023-09-01
pppd/options.c: fix memory leak on error path (#441)

chipitsine:
found by Coverity

1803    if (path != arg)
1804        free(path);

    CID 436214 (#1 of 1): Resource leak (RESOURCE_LEAK)
    12. leaked_storage: Variable handle going out of scope leaks the
	storage it points to.
	1805    return 1;

### DIFF
--- a/pppd/options.c
+++ b/pppd/options.c
@@ -1802,6 +1802,7 @@ loadplugin(char **argv)
     (*init)();
     if (path != arg)
 	free(path);
+	dlclose(handle);
     return 1;
 
  errclose:


paulusmack:
The man page for dlclose() says "If  the  object's  reference count drops to zero and no symbols in this object are required by other objects, then the object is unloaded...". In general a plugin won't have symbols that are required by other objects, so it would get immediately unloaded. And despite the commit headline, the line you are adding is in the success path, not the error path, so unloading the plugin would be entirely the wrong thing to do.

So this patch is quite wrong. I don't see anything needing to be fixed in the existing pppd code.

Issue #446, opened 2023-09-16
ppp 2.5.0 crashed with buffer overfow when device name is too long


Reverier-Xu:
Hi, after I upgraded ppp to 2.5.0 it became unusable.

My env is:

```
Archlinux linux-6.5.2-arch1-1
NetworkManager 1.44.0-1
pppd version 2.5.0
```

this is my `journalctl -b -u NetworkManager`:
```
Sep 16 20:51:44 Reverier-Arch NetworkManager[770]: <info>  [1694868704.6769] policy: auto-activating connection 'XDU PPPoE' (5b298d08-9a67-4c25-8965-5d3f12e9ff71)
Sep 16 20:51:44 Reverier-Arch NetworkManager[770]: <info>  [1694868704.6778] device (enp0s13f0u1u1c2): Activation: starting connection 'XDU PPPoE' (5b298d08-9a67-4c25-8965-5d3f12e9ff71)
Sep 16 20:51:44 Reverier-Arch NetworkManager[770]: <info>  [1694868704.6780] device (enp0s13f0u1u1c2): state change: disconnected -> prepare (reason 'none', sys-iface-state: 'managed')
Sep 16 20:51:44 Reverier-Arch NetworkManager[770]: <info>  [1694868704.6786] device (enp0s13f0u1u1c2): delaying PPPoE reconnect for 0.002 seconds to ensure peer is ready...
Sep 16 20:51:44 Reverier-Arch NetworkManager[770]: <info>  [1694868704.6812] device (enp0s13f0u1u1c2): PPPoE reconnect delay complete, resuming connection...
Sep 16 20:51:44 Reverier-Arch NetworkManager[770]: <info>  [1694868704.6814] device (enp0s13f0u1u1c2): state change: prepare -> config (reason 'none', sys-iface-state: 'managed')
Sep 16 20:51:44 Reverier-Arch NetworkManager[770]: <info>  [1694868704.6821] ppp-manager: starting PPP connection
Sep 16 20:51:44 Reverier-Arch NetworkManager[770]: <info>  [1694868704.6845] ppp-manager: pppd started with pid 29695
Sep 16 20:51:44 Reverier-Arch pppd[29695]: Plugin pppoe.so loaded.
Sep 16 20:51:44 Reverier-Arch NetworkManager[29695]: Plugin pppoe.so loaded.
Sep 16 20:51:44 Reverier-Arch NetworkManager[29695]: PPPoE plugin from pppd 2.5.0
[...]

PR #447, opened 2023-09-19
Fix for #446, don't crash on too long of a device name

enaess:
This change will fix the crash that happens in issue #446 

However, it is worth noting that the device name is typically limited by the struct ifr, and the size of the ifr.ifr_name field (15). It's just that when it is used in context of the sockaddr structure, it's being limited to 14 bytes. The origin of this is the "device name" option which is capped at a really large value. Should probably be set to 14??


Neustradamus:
@paulusmack: Have you seen this @enaess PR?

PR #456, opened 2023-11-06
Fix for failure to detect the presence of sockaddr_ll in linux/if_packet.h

enaess:
Previous conftest failed with a compiler error
```
if (sizeof ((struct sockaddr_ll))) {
   ...
}
```

The double parenthesis failed the evaluation of sizeof and thus caused the AC_CHECK_TYPE fail. Replacing with appropriate AC_COMPILE_IFELSE construct.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -76,7 +76,16 @@ AM_COND_IF([LINUX], [
         linux/if_packet.h       \
         netinet/if_ether.h      \
         netpacket/packet.h])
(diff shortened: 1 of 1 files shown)


enaess:
The solaris job didn't fail because of this change. Looks like the CI/CD pipeline for Solaris is currently broken

Neustradamus:
@RICCIARDI-Adrien currently tries to look for a solution about Solaris because there is a problem with it and it is a general problem...

Issue #465, opened 2023-12-14
PPPD 2.4.9 cannot send PADT message when DISCONNECTED


KuoEuran:
Hi,
     I found that PPPD 2.4.9 delete sendPADT.patch and add conn->discoverySocket=-1 in discovery() function, but after I recover these modifications, but it also cannot sendPADT message when disconneced.
PS: 1. it can send PADT message in PPPD 2.4.7.
      2. PADR AND PADI message successfully sent when connected to pppoe wan.

Neustradamus:
@paulusmack, @enaess: Have you seen this ticket?

KuoEuran:
Hi @Neustradamus ,
     Explain in more detail, discovery() function has assigned -1 to conn->discoverySocket in discovery.c, when I reboot my DUT, the program enters PPPOEDisconnectDevice function, but it can't send PADT message normally in PPPD2.4.9 though I add sendPADT function in front of if(conn->discoverySocket>=0). In this situation, because conn->discoverySocket is -1, sendPADT function will return when  conn->discoverySocket is -1, therefore, it can't send PADT is normal.
     In second experiment, I delete the code which assigned -1 to conn->discoverySocket in discovery.c, and conn->discoverySocket is 14 which makes the program enter if(conn->discoverySocket>=0) in function PPPOEDisconnectDevice. However, I discovered that it return -1 in parameter err in sendPacket() function in if.c when I reboot system or kill the pppd process. Therefore, it cannot send PADT message.
     If you have any ideas on this situation, please give me some advice or maybe we can have some discussions in google meeting, thanks.

Neustradamus:
@paulusmack, @enaess, @jkroonza, @chipitsine: What do you think?

paulusmack:
I don't really follow what the actual problem is. What is the situation where you want pppd to send a PADT and it doesn't?

KuoEuran:
Hi, @paulusmack 
    Thank you for replying. When I reboot or firmware upgrade DUT which installed openwrt, or when I kill the PPPD process in background, it should send PADT but it doesn't.

paulusmack:
I think it was commit 2b4166d02ed0 that broke this.

Please test the fixpppoe branch (or PR #470) and tell me if it fixes the problem.

KuoEuran:
Hi @paulusmack ,
    thank you for replying, this fixpppoe branch fixes the problem.

Neustradamus:
[...]

PR #470, opened 2024-01-02
pppoe: Fix sending of PADT on connection termination

paulusmack:
Commit 2b4166d02ed0 ("Close discovery socket after session completed", 2020-11-26) arranged for the discovery socket to be closed when the PPPOE session negotiation was complete.  However, the discovery socket is used for sending a PADT message when the connection terminates, and now that doesn't work because the socket has been closed.

To fix this, we reopen the discovery socket in order to send the PADT message.


Neustradamus:
Thanks to @KuoEuran about this ticket and @paulusmack for this fix which will be in 2.5.1!
- https://github.com/ppp-project/ppp/issues/465

Neustradamus:
@paulusmack: @KuoEuran has comment here:
https://github.com/ppp-project/ppp/issues/465#issuecomment-1874802744

This PR can be merged?

PR #472, opened 2024-01-10
Restore adding a default route with different metric

### LINKED ISSUE #357: defaultroute-metric doesn't add a new route even if metrics are different
I have the following default route : 

```
default via 10.X.X.X dev enp2s0 proto static metric 50 
```

I am trying to set up a second default route through ppp (version 2.4.9) with higher metric (300) using defaultroute and defaultroute-metric options. 

Here are the options of pppd : 
```
pppd options in effect:
debug		# (from command line)
nodetach		# (from command line)
dump		# (from command line)
defaultroute-metric 300		# (from command line)
noauth		# (from command line)
/dev/ttyUSB2		# (from command line)
230400		# (from command line)
lock		# (from /etc/ppp/options)
connect ./chatscript.sh		# (from command line)
xonxoff		# (from command line)
modem		# (from /etc/ppp/options)
asyncmap a0000		# (from /etc/ppp/options)
lcp-echo-failure 4		# (from /etc/ppp/options)
lcp-echo-interval 30		# (from /etc/ppp/options)
hide-password		# (from /etc/ppp/options)
defaultroute		# (from command line)
usepeerdns		# (from command line)
192.168.1.2:192.168.1.1		# (from command line)
noipx		# (from /etc/ppp/options)
```

But I get an error telling me it is trying to replace the default route : 

```
Script ./chatscript.sh finished (pid 421), status = 0x0
Serial connection established.
using channel 34
Using interface ppp0
Connect: ppp0 <--> /dev/ttyUSB2
rcvd [LCP ConfReq id=0x5c <asyncmap 0xa0000> <magic 0x36f6>]
sent [LCP ConfReq id=0x1 <asyncmap 0xa0000> <magic 0xb61544f6> <pcomp> <accomp>]
sent [LCP ConfAck id=0x5c <asyncmap 0xa0000> <magic 0x36f6>]
rcvd [LCP ConfAck id=0x1 <asyncmap 0xa0000> <magic 0xb61544f6> <pcomp> <accomp>]
sent [LCP EchoReq id=0x0 magic=0xb61544f6]
sent [CCP ConfReq id=0x1 <deflate 15> <deflate(old#) 15> <bsd v1 15>]
sent [IPCP ConfReq id=0x1 <compress VJ 0f 01> <addr 192.168.1.2> <ms-dns1 0.0.0.0> <ms-dns2 0.0.0.0>]
sent [IPV6CP ConfReq id=0x1 <addr fe80::e5cc:aece:d9bf:e2c1>]
rcvd [IPCP ConfReq id=0xce <addr 0.0.0.0>]
sent [IPCP ConfNak id=0xce <addr 192.168.1.1>]
rcvd [LCP EchoRep id=0x0 magic=0x36f6]
rcvd [LCP ProtRej id=0x33 80 fd 01 01 00 0f 1a 04 78 00 18 04 78 00 15 03 2f]
Protocol-Reject for 'Compression Control Protocol' (0x80fd) received
rcvd [LCP ProtRej id=0x33 80 57 01 01 00 0e 01 0a e5 cc ae ce d9 bf e2 c1]
Protocol-Reject for 'IPv6 Control Protocol' (0x8057) received
rcvd [IPCP ConfRej id=0x1 <compress VJ 0f 01>]
sent [IPCP ConfReq id=0x2 <addr 192.168.1.2> <ms-dns1 0.0.0.0> <ms-dns2 0.0.0.0>]
rcvd [IPCP ConfReq id=0xde <addr 192.168.1.1>]
sent [IPCP ConfAck id=0xde <addr 192.168.1.1>]
rcvd [IPCP ConfNak id=0x2 <ms-dns1 8.8.8.8> <ms-dns2 8.8.4.4>]
sent [IPCP ConfReq id=0x3 <addr 192.168.1.2> <ms-dns1 8.8.8.8> <ms-dns2 8.8.4.4>]
rcvd [IPCP ConfAck id=0x3 <addr 192.168.1.2> <ms-dns1 8.8.8.8> <ms-dns2 8.8.4.4>]
Script /etc/ppp/ip-pre-up started (pid 424)
Script /etc/ppp/ip-pre-up finished (pid 424), status = 0x0
not replacing default route to enp2s0 [10.X.X.X]
local  IP address 192.168.1.2
remote IP address 192.168.1.1
primary   DNS address 8.8.8.8
secondary DNS address 8.8.4.4
Script /etc/ppp/ip-up started (pid 427)
Script /etc/ppp/ip-up finished (pid 427), status = 0x0
```

Have I misunderstood the defaultroute-metric option?

sthibaul:
For instance, when using a ppp link as backup link, one would want to add a default route on the ppp link, in *addition* to the existing default route.

d0ccb87156c2 ("pppd: Add replacedefaultroute option (#200)") however broke this case: sifdefaultroute was not passing the metric to defaultroute_exists any more. This commit restores this case.

Fixes #357

### DIFF
--- a/pppd/sys-linux.c
+++ b/pppd/sys-linux.c
@@ -2193,11 +2193,27 @@ int sifdefaultroute (int unit, u_int32_t ouraddr, u_int32_t gateway, bool replac
 	 * - this is normally only the case the doing demand: */
 	if (defaultroute_exists(&tmp_rt, -1))
 	    del_rt = &tmp_rt;
+    } else if (!replace) {
+	/*
+	 * We don't want to replace an existing route.
+	 * We may however add our route along an existing route with a different
+	 * metric.
+	 */
+	if (defaultroute_exists(&rt, dfl_route_metric) && strcmp(rt.rt_dev, ifname) != 0) {
+	   if (rt.rt_flags & RTF_GATEWAY)
+	       error("not replacing existing default route via %I with metric %d",
+		     SIN_ADDR(rt.rt_gateway), dfl_route_metric);
+	   else
+	       error("not replacing existing default route through %s with metric %d",
+		     rt.rt_dev, dfl_route_metric);
+	   return 0;
+	}
     } else if (defaultroute_exists(&old_def_rt, -1           ) &&
 			    strcmp( old_def_rt.rt_dev, ifname) != 0) {
 	/*
-	 * We did not yet replace an existing default route, let's
-	 * check if we should save and replace a default route:
+	 * We want to replace an existing route and did not replace an existing
+	 * default route yet, let's check if we should save and replace an
+	 * existing default route:
 	 */
 	u_int32_t old_gateway = SIN_ADDR(old_def_rt.rt_gateway);
 


jkroonza (commented):
The relevant documentation of the defaultroute options:

       defaultroute
              Add  a  default  route to the system routing tables, using the peer as the gateway,
              when IPCP negotiation is successfully completed.  This entry is  removed  when  the
              PPP  connection  is broken.  This option is privileged if the nodefaultroute option
              has been specified.

       defaultroute-metric
              Define the metric of the defaultroute and only add it if there is no other  default
              route  with the same metric.  With the default value of -1, the route is only added
              if there is no default route at all.

       replacedefaultroute
              This  option  is a flag to the defaultroute option. If defaultroute is set and this
              flag is also set, pppd replaces an existing default  route  with  the  new  default
              route.  This option is privileged.

(An appenddefaultroute option would also be convenient but let's not over-complicate things).

I think we should *ignore* the IPv6 situation, defaultroute6 technically already violates the standards as I understand them, but it's been stated that providers are messing around, so let's just leave that as is.  Also, this could *technically* be handled more flexibly from ipv6-up.  Same for ip-up actually come to think about it.

So there are a few valid combinations of options:

1.  defaultroute
2.  defaultroute+defaultroute-metric
3.  defaultroute+replacedefaultroute
4.  defaultroute+defaultroute-metric+replacedefaultroute

So I believe for these combinations the behaviour should be as follows (Ignoring 3 which is problematic):

1. If any default route exist, don't add one, if none exist, add with metric 0.
2. If a default route with the same metric exist, don't add one, else add with relevant metric.
4. If a default route with the same metric exist, remove it (restore on ipv4 down) and re-add via ppp.

A further problem with replacedefaultroute is that it's possible in linux to have *multiple* exact same routes at the same metric, so technically even 4 is a problem above:

```
plastiekpoot [09:01:41] ~ # ip ro sh
default via 192.168.42.1 dev wlp2s0 proto dhcp src 192.168.42.21 metric 2002 
...
plastiekpoot [09:02:05] ~ # ip route append default via 192.168.42.1 dev wlp2s0 proto static src 192.168.42.21 metric 2002
plastiekpoot [09:02:10] ~ # ip ro sh
default via 192.168.42.1 dev wlp2s0 proto dhcp src 192.168.42.21 metric 2002 
default via 192.168.42.1 dev wlp2s0 proto static src 192.168.42.21 metric 2002 
...
```

So with replacedefaultroute - should we remove *all* matching default routes or only the first matching one?

Which leads into my logic issue with 3 above.  If there is only one other default route it's all good and well, we remove it and re-add later, but what if there are multiple such routes?

I guess the replacedefaultroute option is problematic regardless.  We should only ever add routes and remove our own, the system administrator should manage priority via metrics.

So really the question is if a matching route with the same metric exist, do we append or not?  Does yet another separate option for that even make sense or should we simply append our route *regardless* - as such - nuke the replacedefaultroute option entirely, and append another default route at the specified metric and move on, thus reduce the defaultroute options to "defaultroute" and "defaultroute-metric" where if defaultroute is given we ALWAYS APPEND a default route at the specified defaultroute-metric (which then defaults to 0).

If the system administrator wants more control than that - have them rely on ip-(pre-)up or net-pre-up and let them deal with it there?

Whilst the existing logic is definitely broken, I'm not convinced that the PR brings things closer to the documented behaviour either, nor am I sure the documented behaviour is well defined.

Neustradamus:
@sthibaul: Good, @apompee has confirmed on the ticket that it works with your PR :)

apompee:
Since the defaultroute-metric only works if defaultroute option is set. Wouldn't it also make sense to modify the defaultroute option to accept a metric parameter which defaults to 0 if not set? And then the defaultroute-metric option could be removed.

sthibaul:
> Whilst the existing logic is definitely broken, I'm not convinced that the PR brings things closer to the documented behaviour either

Sure it does: it does fix https://github.com/ppp-project/ppp/issues/357

Indeed, the multi-default routes situation pose question, and it was probably not considered when these options were introduced. But this merge request does fix the (very common) case of an existing default route, and adding another one with a different metric. The case of #357, precisely.

Revamping defaultroute option can be useful, but seems very off-topic for this merge request (which is really about bringing back behavior which *was* working in the past).

PR #478, opened 2024-02-27
pppd: sys-linux.c ioctl cmd SIOCSDSTADDR

jared1980:
the gw/netmask value should be assigned to ifr.ifr_addr. In kernel (net/ipv4/devinet.c), the value was token out from ifr.if_addr.

### DIFF
--- a/pppd/sys-linux.c
+++ b/pppd/sys-linux.c
@@ -3235,7 +3235,7 @@ int sifaddr (int unit, u_int32_t our_adr, u_int32_t his_adr,
  *  Set the gateway address
  */
     if (his_adr != 0) {
-	SIN_ADDR(ifr.ifr_dstaddr) = his_adr;
+	SIN_ADDR(ifr.ifr_addr) = his_adr;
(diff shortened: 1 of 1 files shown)


paulusmack:
The existing code isn't wrong, because struct ifreq has a union in it which means that ifr_addr, ifr_dstaddr and ifr_netmask are all the same bytes of storage. Since this is a kernel/user ABI, the kernel team will ensure that it doesn't get changed in an incompatible way, so the existing code is safe, and arguably using ifr_dstaddr etc. is more expressive.

As to the argument that kernel code uses "ifr.if_addr", that would have some validity if the kernel and pppd were using the same definition of struct ifreq, but they're not. Since pppd is userspace, the definition of struct ifreq that it uses comes from the C runtime (glibc or whatever), and the ioctl() is likewise provided by the C runtime. It may be (and usually is) that ioctl() is a thin wrapper around a system call and that glibc's struct ifreq has all the same things in the same places as the kernel does, but that is not necessarily so.

Lastly, I can't take this commit without a proper signoff.

PR #479, opened 2024-03-05
Use pkg-config to detect PAM when possible

floppym:


Neustradamus:
@paulusmack, @enaess: What do you think?

paulusmack:
I would like the commit message to have at least a sentence saying why this is a good thing and what will be the effect from the point of view of someone running ./configure. For example, are there now any new possibilities for options to ./configure, or do any existing options work any differently now?

As for the substance of the change, I want to hear from @enaess .

floppym:
@paulusmack I updated the commit message with an explanation.

Neustradamus:
@paulusmack: Have you seen @floppym changes?

enaess:
I'll have a look tomorrow, but also worth noting is that there are other ax_***.m4 files that use the same templated pattern.

m4/ax_check_pam.m4
```diff
@@ -26,34 +26,28 @@
...
-            if test -f "$pamdir/include/security/pam_appl.h"; then
-                PAM_CFLAGS="-I$pamdir/include"
-                PAM_LDFLAGS="-L$pamdir/lib"
-                AC_MSG_RESULT([yes])
-                break
-            else
-                AC_MSG_RESULT([no])
-            fi
-        done
+    AS_IF([test "$with_pam" != "no"], [
```
enaess, line 31: Generally, the convention is to use "x${with_pam}" != "xno", why change it?

m4/ax_check_pam.m4
```diff
@@ -26,34 +26,28 @@
...
+        AS_CASE(["$with_pam"],
+            [""|yes], [PKG_CHECK_MODULES([PAM], [pam], [pamdirs=],
+                        [pamdirs="/usr/local /usr/lib /usr"])],
```
enaess, line 34: So you're swapping out the shell script's 'case' with AS_CASE, which in itself is fine and would improve compatibility; but you are eliminating a few of the case options, like what happens if you type --with-pam=ye, or --with-pam=n, or --with-pam=no. I'd like to see them added back with the AS_CASE.

enaess (changes requested):
Generally I am good with the idea of using pkg-config w/fallback of the hard-coded values in case pkg-config doesn't exist. 

I'd additionally like to see similar change in the other .m4 files for pcap, atm, srp, and others to avoid the pam m4 file from deviating from the new "norm" if you like.

m4/ax_check_pam.m4
floppym: I have never understood the convention. Can you explain why the "x" is significant?

m4/ax_check_pam.m4
floppym: Passing `ye` or `n` to --with-pam makes no sense. Valid values should be yes, no, or a path. Trying to accept every possible misspelling seems silly to me.

enaess: I don't think I care much about the 'ye' but the 'y' and 'n' or 'yes' and 'no' should at lest work (short-vs-long). Just hope that nobody like buildroot used a miss-typed 'ye' for yes.

m4/ax_check_pam.m4
enaess: It's a configure / shell compatibility thing. I've seen people insisting on putting it in to make these script consistent. You can look at the changelog for configure.ac

m4/ax_check_pam.m4
floppym: Also, the most common usage is to pass `--with-pam` which implies a value of `yes`, or `--without-pam` which implies a value of `no`.

floppym: I added back handling for "ye", "y", and "n".

m4/ax_check_pam.m4
floppym: I have restored the "x" when passing strings to the test command.

floppym:
> I'd additionally like to see similar change in the other .m4 files for pcap, atm, srp, and others to avoid the pam m4 file from deviating from the new "norm" if you like.

I can sync things up in separate commits once the maintainers approve of what I have done with PAM. No sense in having me do a bunch of extra work if it will be rejected anyway.

enaess:
Could open an issue against the other .m4 files, but okay from me.

PR #493, opened 2024-05-22
Exit early if we strlen on an empty string

AZero13:
In the event we do strlen on an empty string, we should bail.

### DIFF
--- a/pppd/auth.c
+++ b/pppd/auth.c
@@ -2164,6 +2164,8 @@ auth_number(void)
     while (wp) {
 	/* trailing '*' wildcard */
 	l = strlen(wp->word);
+	if (l == 0)
+	    return 1;
 	if ((wp->word)[l - 1] == '*')
 	    l--;
 	if (!strncasecmp(wp->word, remote_number, l))


Neustradamus:
@paulusmack: Have you seen this @AtariDreams PR?

paulusmack:
In the context of the existing code, the change proposed is reasonable, but the commit message is not illuminating, because it doesn't explain the connection between an empty string and bailing out. The mention of strlen doesn't help - doing strlen on an empty string is fine.

I think the existing code is actually buggy and has been buggy since it was first written, in that it treats all allowed number strings as wildcards, whether there is an '*' on the end or not. I don't think that was really the original intention. That means that an empty string is equivalent to "*", and that is why bailing out on an empty string is correct (in terms of fixing the code without changing its behaviour).

Neustradamus:
@AtariDreams: Have you seen the @paulusmack answer?

Neustradamus:
@RSilicon: Have you seen the @paulusmack comments?

paulusmack:
never mind... I pushed a change to make it handle non-wildcards correctly, which also solves the issue this PR was addressing.

PR #494, opened 2024-05-22
Prevent fract and rate from overflowing by exiting early

AZero13:


Neustradamus:
@paulusmack: Have you seen this second @AtariDreams PR?

pppd/plugins/pppoatm/text2qos.c
```diff
@@ -69,9 +71,9 @@ int __t2q_get_rate(const char **text,int up)
 		rate = (rate+(up ? 8*ATM_CELL_PAYLOAD-1 : 0))/8/
 		  ATM_CELL_PAYLOAD;
 		end += 3;
+		if (rate > INT_MAX) return RATE_ERROR;
```
paulusmack, line 74: Is this going to work? Can rate ever be greater than INT_MAX? I think the compiler would be justified in removing this test as being obviously always false. Any overflow that could happen would have already happened by this point.

The only overflow that could occur in line 69 is in the addition of the rounding-up constant, because we are basically dividing by 384 here. So perhaps we should instead divide by 8 * ATM_CELL_PAYLOAD and then increment if there was any remainder and 'up' is true. Or something like that. I think it would be wrong to return an error in the case where rate > INT_MAX - 383 and 'up' is true, since the correct result is perfectly well defined and is less than INT_MAX.

paulusmack (changes requested):
See comments...

Neustradamus:
@AtariDreams: Have you seen the @paulusmack comments?

Neustradamus:
@RSilicon: Have you seen the @paulusmack comments?

paulusmack:
I did the change I suggested above to avoid any possibility of overflow in the division with rounding up. As I said, I don't think the other change is necessary. Hence I'm closing this PR.

PR #501, opened 2024-07-07
Fix options.c resource leak

BluesLion:
Fix loadplugin resource leak dlopen but not close.

### DIFF
--- a/pppd/options.c
+++ b/pppd/options.c
@@ -1802,6 +1802,7 @@ loadplugin(char **argv)
     (*init)();
     if (path != arg)
 	free(path);
+    dlclose(handle);
     return 1;
 
  errclose:


paulusmack:
Doing dlclose() in the success path would unload the module we just loaded. In the success path, we want the code to stay resident until the process exits, so doing dlclose() where you suggest is quite wrong.

paulusmack:
> Thank you very much for your reply. I have a question: the handle is not passed out, then the handle will not be cleared when the process ends. Won't the handle cause a resource leak issue?

No, because the kernel automatically frees all memory and other resources used by the process when it terminates.

BluesLion:
Thank you for your response. I just looked up the information, and you are indeed correct. I apologize for any inconvenience I may have caused.

PR #504, opened 2024-07-12
pppd: workaround sparc <asm/termbits.h> issue

tpetazzoni:
The SPARC version of the <asm/termbits.h> kernel header has an issue: it defines "struct termio", which conflicts with the definition in <bits/ioctl-types.h>, causing a build failure in pppd.

This issue was fixed in the upstream Linux kernel in commit c32d18e7942d7589b62e301eb426b32623366565 ("sparc: move struct termio to asm/termios.h"), which is going to appear in Linux 6.10.

In order to support building with older toolchains, this patch includes in the pppd source code a copy of the fixed SPARC <asm/termbits.h> in the pppd code base, and uses it if we're on SPARC with kernel headers older than 6.10.

Fixes:

In file included from termios_linux.h:39:
/home/thomas/projets/buildroot/output/host/sparc64-buildroot-linux-gnu/sysroot/usr/include/asm/termbits.h:17:8: error: redefinition of ‘struct termio’
   17 | struct termio {
      |        ^~~~~~
  CC       pppd-ipv6cp.o
In file included from /home/thomas/projets/buildroot/output/host/sparc64-buildroot-linux-gnu/sysroot/usr/include/sys/ioctl.h:29,
                 from sys-linux.c:76:
/home/thomas/projets/buildroot/output/host/sparc64-buildroot-linux-gnu/sysroot/usr/include/bits/ioctl-types.h:36:8: note: originally defined here
   36 | struct termio
      |        ^~~~~~

### DIFF
--- a/pppd/sparc-termbits.h
+++ b/pppd/sparc-termbits.h
@@ -1,0 +1,247 @@
+/* SPDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note */
+#ifndef _SPARC_TERMBITS_H
+#define _SPARC_TERMBITS_H
+
+#include <linux/posix_types.h>
+
+typedef unsigned char   cc_t;
+typedef unsigned int    speed_t;
+
+#if defined(__sparc__) && defined(__arch64__)
+typedef unsigned int    tcflag_t;
+#else
+typedef unsigned long   tcflag_t;
+#endif
+
+#define NCCS 17
+struct termios {
+	tcflag_t c_iflag;		/* input mode flags */
+	tcflag_t c_oflag;		/* output mode flags */
+	tcflag_t c_cflag;		/* control mode flags */
+	tcflag_t c_lflag;		/* local mode flags */
+	cc_t c_line;			/* line discipline */
+	cc_t c_cc[NCCS];		/* control characters */
+};
+
+struct termios2 {
+	tcflag_t c_iflag;		/* input mode flags */
+	tcflag_t c_oflag;		/* output mode flags */
+	tcflag_t c_cflag;		/* control mode flags */
+	tcflag_t c_lflag;		/* local mode flags */
+	cc_t c_line;			/* line discipline */
+	cc_t c_cc[NCCS+2];		/* control characters */
+	speed_t c_ispeed;		/* input speed */
+	speed_t c_ospeed;		/* output speed */
+};
+
+struct ktermios {
+	tcflag_t c_iflag;		/* input mode flags */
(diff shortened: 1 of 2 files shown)


Neustradamus:
@paulusmack: Have you seen this @tpetazzoni PR?

paulusmack:
The problem here is the copyright notice on the file you're adding - it's GPL2, which is incompatible with the BSD licence that the rest of pppd code is under (referring specifically to pppd itself as distinct from its plugins). Is there another solution that doesn't involve importing Linux kernel source code?

paulusmack:
Really, this seems to be a conflict between glibc and the kernel. Trying to solve it here in pppd seems the wrong answer.

tpetazzoni:
> Really, this seems to be a conflict between glibc and the kernel. Trying to solve it here in pppd seems the wrong answer.

Thanks for the feedback. As mentioned in my commit message, this bug has been fixed in the kernel, but very recently, so it will take some time for the issue to be fixed in all toolchains (that typically use slightly older kernel headers).

tpetazzoni:
> The problem here is the copyright notice on the file you're adding - it's GPL2, which is incompatible with the BSD licence that the rest of pppd code is under (referring specifically to pppd itself as distinct from its plugins). Is there another solution that doesn't involve importing Linux kernel source code?

Thanks for the feedback. I had not realized the licensing issue. Note that this license issue is quite silly: the pppd code already does "#include <asm/termbits.h>", which means it includes the exact same kernel header file, which is therefore also under GPLv2...

I don't really have a good solution, the kernel header is buggy and the only option is to carry a non-buggy version.

paulusmack:
> Thanks for the feedback. I had not realized the licensing issue. Note that this license issue is quite silly: the pppd code already does "#include <asm/termbits.h>", which means it includes the exact same kernel header file, which is therefore also under GPLv2...

The issue isn't what ultimately gets combined with what, the issue is what is I am distributing.

> I don't really have a good solution, the kernel header is buggy and the only option is to carry a non-buggy version.

I'm quite sure that's not the only possible solution.

paulusmack:
I see that the upstream patch to fix the problem properly (that is, in the Linux kernel) has been accepted into versions v6.9.5, v6.6.34, v6.1.94, v5.15.161, v5.10.219, v5.4.278, and v4.19.316, as well as v6.10. So I think there are plenty of kernel versions for people to use where pppd will now compile without this change.

tpetazzoni:
Indeed, fair enough. Thanks for looking into this, I guess this PR can be closed then.

PR #379, opened 2022-12-24
Header file reorganization and cleaning up the public API for pppd version 2.5.0

enaess:
Summary:
Much of pppd.h used to pile on a bunch of extern declarations for internal use in other files of pppd. This was never meant as a "public API" to be exported and used by others. This change tries to limit that extensively by hiding the extern declarations into pppd-private.h (by virtue of git mv pppd.h to pppd-private.h to retain source control history, and creating a new pppd.h) and migrating only needed functions back into pppd.h for plugins to compile.

I've compiled a few different projects with the new pppd headers. It's is a breaking change, but plugin maintainers will have to do some maintenance here anyways...

This change reorganizes the header files:
 - Unexports the following headers: 
     - chap-md5.h
     - crypto-priv.h
     - eap-tls.h
     - pathnames.h
     - peap.h 
     - pppd-private.h
     - spinlock.h 
     - tdb.h
 - Moves chap-new.h into chap.h
 - Moves ppp-crypto.* into crypto.*
 - Moves pppcrypt.* into crypto_ms.*
 - Uses unsigned char instead of u_char
 - Uses uint32_t instead of u_int32_t
 - Uses bool from <stdbool.h>
 - Rename pppd.h to pppd-private.h, creates another pppd.h with exported API 
 - Creates a new options.h and moving relevant declarations from pppd.h
 - Creates a new multilink.h and moving relevant declarations from pppd.h
 - Documents the API in crypto.h
 - Removed extern variable declarations for MPPE keys in favor of using the new MPPE API
 - Converting to functions and/or renaming existing functions to better align with an API 
   - Don't use "extern int status", it's way too common to use this in code, use ppp_set_status() or ppp_status(). Exit codes moved into an enum for compiler safety
   - Don't use "extern int debug", it's way too commong to use this in code. This is now known as debug_on()
   - Timer functions was renamed and prefixed with "ppp_"
   - Fixed up link statistics, call bool ppp_get_link_stats(&stats) to get the stats
   - Changed add_notifier to ppp_add_notify() and unexported the notifier structures in favor of a enumerated type.
   - Replace use of extern variables in favor of a function call
      - netif_(set|get)_mru -> ppp_(set|get)_mru
      - ppp_available -> ppp_check_kernel_support (just better nameing)
      - generic_(dis)establish_ppp -> ppp_generic_(dis)establish
      - script_(un)setenv -> ppp_script_(un)setenv
      - got_sigterm -> ppp_signaled(int signo)
[...]

PR #505, opened 2024-07-14
fix reliability of print_link_stats (with option persist)

enometh:
* pppd/ipcp.c: (ipcp_down): fix comment
* pppd/main.c: (reset_link_stats): reset print_link_stats to 1, set start_time even if get_ppp_stats fails.

### DIFF
--- a/pppd/main.c
+++ b/pppd/main.c
@@ -1331,9 +1331,9 @@ print_link_stats(void)
 void
 reset_link_stats(int u)
 {
-    if (!get_ppp_stats(u, &old_link_stats))
-	return;
+    get_ppp_stats(u, &old_link_stats);
     ppp_get_time(&start_time);
+    link_stats_print = 1;
 }
 
 /*
--- a/pppd/ipcp.c
+++ b/pppd/ipcp.c
@@ -2026,7 +2026,7 @@ ipcp_down(fsm *f)
     sifvjcomp(f->unit, 0, 0, 0);
 
     print_link_stats(); /* _after_ running the notifiers and ip_down_hook(),
-			 * because print_link_stats() sets link_stats_valid
+			 * because print_link_stats() sets link_stats_print
 			 * to 0 (zero) */
 
     /*


enometh:
This attempt to fix the problem noted in the linux-ppp mailing list on mar-26-2024 and may-03-2024
under the subject  "ppp-2.5.0 sometimes doesn't print stats on terminating on signal 2"

The sent/recv log messages  were being lost, especially with the persist option.
This seems to be an oversight during reorg in commit
ba7f7e053da "Header file reorganization and cleaning up the public API for pppd version 2.5.0 (#379)" around the repurposing of the link_stats_valid variable as link_stats_print.

it also fixes a stray reference to the old variable in a comment.
please review. the reordering in  reset_link_stats

Neustradamus:
@paulusmack: Have you seen this @enometh PR?

paulusmack:
First, this needs a sign-off, see Submitting-patches.md.

Secondly, it needs an explanation that is more than simply saying what the textual change is. Don't make the reader have to reverse-engineer your thought process.

paulusmack:
@enaess is does seem like ba7f7e0 introduced a problem with the link_stats_print variable, in that now print_link_stats will only ever produce output at most once. What was the motivation for introducing link_stats_print? It looks to me like I should just revert that part of ba7f7e0.

enometh:
> First, this needs a sign-off, see Submitting-patches.md. 
> Secondly, it needs an explanation that is more than simply saying what the textual change is. Don't make the reader have to reverse-engineer your thought process.

I explained the change to the extent it could be explained to both on the mailing list and in the comment on the pull request. Since ithe patch  was a trivial 2 line bugfix, of someone else's code, I skipped looking into what the "Sign off" involved. However I am using my real name and real email address. (I go by S.Madhu, officially)

paulusmack:
Not sure why the DCO check is failing...

enaess:
I'd prefer if we filed a bug on what wasn't working. All it says is it is a reliability issue with signal 2??

The get_ppp_stats() function returns 1 on success, then it also configures a timer to call a function repeatedly. The function it calls depends on what is configured and what OS you are on. Debug logs should reveal what is going on here. The get_ppp_stats() returns the value of the function it dynamically resolved, and if success it should be "1". If you get "0" then I'd be very interested in why??

So, again; please file a bug and we can look at your setup and try to figure out what is going on wrong. I am not sure if simply adding a few lines to say "now" it works, to resolve "what" problem in the first place.

enometh:
> I'd prefer if we filed a bug on what wasn't working. All it says is it is a reliability issue with signal 2??
> 
> The get_ppp_stats() function returns 1 on success, then it also configures a timer to call a function repeatedly. The function it calls depends on what is configured and what OS you are on. Debug logs should reveal what is going on here. The get_ppp_stats() returns the value of the function it dynamically resolved, and if success it should be "1". If you get "0" then I'd be very interested in why??
> 
> So, again; please file a bug and we can look at your setup and try to figure out what is going on wrong. I am not sure if simply adding a few lines to say "now" it works, to resolve "what" problem in the first place.

Please see
https://lore.kernel.org/linux-ppp/20240326.220015.873398620206616683.enometh@meer.net/
and
http://lore.kernel.org/linux-ppp/20240503.151110.1561650016112331638.enometh@meer.net/

I believe this is a bonafide bug in the code which my patch fixes, and would verify it if someone
could take the trouble to understand what it fixes.

enometh:
> Not sure why the DCO check is failing...
My author name and email was Madhu <enometh@net.meer> (not routable)
And I signed off as "S Madhu" <enometh@meer.net>, which I thought was acceptable
I used to see commits authored by one person signed-off by another person, but
this didn't work on github.
I've adjusted the GIT_AUTHOR_EMAIL and GIT_COMMITTER_EMAIL to match the
routable email, but I would prefer to keep my original signature (name + email) if possible,
would that be possible in theory?
in the case that this patch is accepted, of course.

paulusmack:
@enaess I think the basic problem is that link_stats_print is only ever set to 1 on initialization, hence print_link_stats() will only ever produce output once (at most). However, there are cases, e.g. when the persist option is used, where we used to get output from print_link_stats more than once, as the one pppd instance could handle a series of connections. This change was introduced in ba7f7e0. The fix proposed here, to set print_link_stats to 1 in reset_link_stats(), seems reasonable to me.

I agree it wants a bit more explanation in the commit message - the fix may be textually simple but that doesn't mean that the reasoning behind it is intuitively obvious.

pppd/main.c
```diff
@@ -1331,9 +1331,9 @@ print_link_stats(void)
 void
 reset_link_stats(int u)
 {
-    if (!get_ppp_stats(u, &old_link_stats))
-	return;
+    get_ppp_stats(u, &old_link_stats);
```
paulusmack, line 1334: Should we initialize old_link_stats to all zeroes if the get_ppp_stats call fails? Or copy it from link_stats?

enaess:
I *think* the use of print_link_stats was intended to avoid double printing the result on exit (though, I am not 100% sure). If there was instances where an update was to be dispatched to a script on a regular interval, then I must have missed that memo. 

I'd have to run this code and troubleshoot any side-effects of making that change. I do recall it wasn't a straight forward shot. Also, do we know if the print_link_stats will ever be called in case of a SIGHUP (and if it is, the value of print_link_stats is invalid?).

enometh:
*  Eivind Næss ***@***.***> ***@***.***>
Wrote on Wed, 21 Aug 2024 09:59:26 -0700

> I *think* the use of print_link_stats was intended to avoid double printing the result on exit (though, I am not 100% sure). If there was instances where an update was to be dispatched to a script on a regular interval, then I must have missed that memo.

I think the patch preserves the logic to avoid double printing the
result.

> I'd have to run this code and troubleshoot any side-effects of making that change. I do recall it wasn't a straight forward shot. Also, do we know if the print_link_stats will ever be called in case of a SIGHUP (and if it is, the value of print_link_stats is invalid?).

I haven't rechecked but I believe the code tries to say: SIGHUP is
supposed to cleanly terminate ppd and print the stats reliably.  This
was not a problem earlier.

enometh:
*  Paul Mackerras ***@***.***> ***@***.***>
Wrote on Wed, 21 Aug 2024 02:29:19 -0700
>  void
>  reset_link_stats(int u)
>  {
> -    if (!get_ppp_stats(u, &old_link_stats))
> -	return;
> +    get_ppp_stats(u, &old_link_stats);
>
> Should we initialize old_link_stats to all zeroes if the get_ppp_stats call fails? Or copy it from link_stats?

I don't know because I couldn't imagine a situation where
get_ppp_stats would fail (and you'd need to know the situation to
decide if the state is to be preserved or blown away). to quote a
fragment in a comment upthread:

	"The get_ppp_stats() returns the value of the function it
	dynamically resolved, and if success it should be "1". [If you
	get "0" then I'd be very interested in why??]

I proposed the change because I thought get_ppp_stats is always
expected to succeed here, and returning immediately is the wrong thing
to do for this function. the failure would probably be catastrophic
but it would "handled" elsewhere.  and there are other calls in the
code where the return value of get_ppp_stats is not checked, so
leaving things as were was probably the best bet,

enaess:
I don't recall correctly, but didn't @paulusmack fix-up some of the signal handling in pppd in the last release (or was that chat.c?). You reported seeing the stats being printed in some cases, and in some cases not? Perhaps the signal handling code has shifted things around that it introduced this bug. 

I'd love getting a repro on this and test this out, but I don't have the time to fiddle with this issue right now.

paulusmack:
> I don't recall correctly, but didn't @paulusmack fix-up some of the signal handling in pppd in the last release (or was that chat.c?). You reported seeing the stats being printed in some cases, and in some cases not? Perhaps the signal handling code has shifted things around that it introduced this bug.
> 
> I'd love getting a repro on this and test this out, but I don't have the time to fiddle with this issue right now.

The key thing here that maybe you missed is that with the persist option, the one pppd instance can handle more than just one connection lifecycle. That is, with persist, pppd will try to reconnect after the connection terminates, and a given pppd instance could handle an unlimited number of connections (in serial fashion, one after another). So the logic where print_link_stats is set to 0 when the stats are printed and never set to 1 again after that is wrong. It's nothing to do with signals.

PR #510, opened 2024-09-01
pppd: Add ip-pre-up-script option

tpaukrt:
This option allow a user to specify the path to the script usually located at /etc/ppp/ip-pre-up, similarly to the existing ip-up-script and ip-down-script options.


paulusmack:
To the extent that this is just making it possible for the admin to override hard-coded paths as they can for other scripts, this seems OK.

@jkroonza @enaess any comments on this one?

jkroonza (approved):
Not tested, but looks good.

PR #367, opened 2022-08-24
pppd: implement net-init, net-pre-up and net-down.

jkroonza:
This no longer differentiates between which sub protocols brings an
interface up or down, merely tracks which sub protocols have brought up
and interface, and which have not yet taken it down.

There is one caveat here, I use pointer comparison on these names, so
the absolutely cannot be anything other than a compile-time constant,
and I'm not sure whether or not at least -O1 is required or not, ie, if
the compiler will eliminate multiple "STRINGS" to a single instance in
memory.

I'm unable to test the solaris code, but I don't see any reason why this
should not work.

Signed-off-by: Jaco Kroon <jaco@uls.co.za>

PR #511, opened 2024-09-01
pppd: Add ipv6-pre-up-script option

tpaukrt:
This option allow a user to specify the path to a script that will be run before the interface is brought up, similarly to the existing ip-pre-up script.

### DIFF
--- a/pppd/ipv6cp.c
+++ b/pppd/ipv6cp.c
@@ -322,7 +322,7 @@ struct protent ipv6cp_protent = {
 };
 
 static void ipv6cp_clear_addrs (int, eui64_t, eui64_t);
-static void ipv6cp_script (char *);
+static void ipv6cp_script (char *, int);
 static void ipv6cp_script_done (void *);
 
 /*
@@ -1235,6 +1235,7 @@ ipv6_demand_conf(int u)
 	eui64_magic_nz(wo->ourid);
     }
 
+    ipv6cp_script(path_ipv6preup, 1);
     if (!sif6up(u))
 	return 0;
     if (!sif6addr(u, wo->ourid, wo->hisid))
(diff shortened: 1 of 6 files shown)


paulusmack:
There have been similar proposals before, and a lot of discussion about whether this kind of thing is the right thing to do or not. See PR #341 and PR #367 for starters.

@jkroonza do you want to comment here?

jkroonza:
net-pre-up in master.

```
commit 39063981f7605a994851eec4a7acae86b2327983 (origin/consolidated-ifup-down-and-scripts)
Author: Jaco Kroon <jaco@uls.co.za>
Date:   Tue Sep 26 14:45:13 2023 +0200

    pppd: implement net-init, net-pre-up and net-down.
    
    net-init executes as a blocking script directly after the unit number
    becomes available.  This can be used to initialise aspects related to
    the ppp connection that lives outside of the ppp connection.  It can
    also be used to clean up (in the author's extremely unlikely case) where
    a previous pppd crashed, and net-down didn't execute in order to clean
    up.
    
    net-pre-up executes as a blocking script after auth, prior to NCPs being
    negotiated.  Unlike ip-pre-up this is guaranteed to execute prior to the
    interface being brought up, and can be used in an NCP agnostic manner to
    pre-initialise aspects of the interface for which it still needs to be
    down (amongst others it's recommended that firewall changes happen
    here).
    
    net-down executes in a non-blocking manner just prior to pppd
    terminating and can be used to clean up actions from previous scripts.
    
    You will notice that I mention ip-pre-up doesn't gaurantee that the
    interface will still be down, this is because in a Linux world all
    protocols runs on the same interface, compared to solaris where I'm
    informed each protocol runs on it's own sub-interface, each of which has
    it's own operational state.  The man page for pppd has also been
    adjusted to indicate as much.
    
```

PR #512, opened 2024-09-03
pppd: Add noresolvconf option

tpaukrt:
This option allows a user to specify that he does not want to create the /etc/ppp/resolv.conf file with the DNS server addresses.


paulusmack:
What is the motivation for this?

Instead of doing this, why not just avoid specifying the usepeerdns option? What's the point of asking the peer for DNS addresses and then not using them?

(This kind of information about the deeper reasons behind what you're doing, and reasons why obvious alternatives weren't chosen, should be in the commit message.)

tpaukrt:
The root filesystem on embedded systems is often read-only, so pppd emits an error message when writing to the resolv.conf file. This new option will prevent that.

DNS servers will still be passed to ip-up/ip-down scripts, where they can be handled much better.

PR #513, opened 2024-09-03
pppd: Add pap-secrets and chap-secrets options

tpaukrt:
These options allow a user to specify paths to pap-secrets and chap-secrets files.

### DIFF
--- a/pppd/auth.c
+++ b/pppd/auth.c
@@ -83,6 +83,7 @@
 #include <grp.h>
 #include <string.h>
 #include <strings.h>
+#include <sys/param.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/socket.h>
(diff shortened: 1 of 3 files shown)


paulusmack:
Letting a non-privileged user control authentication seems like a really bad idea.

tpaukrt:
Added OPT_PRIV to both options.

paulusmack:
Still needs explanation (in the commit message) of why we want such fine-grained control, as opposed to, say, using the `--sysconfdir` option when configuring.

tpaukrt:
Added explanation to the commit message.

Neustradamus:
@tpaukrt, @paulusmack: Can you check the @ua0lnj issue?
- https://github.com/ppp-project/ppp/issues/542

PR #514, opened 2024-09-09
pppd: Add net-init-script, net-pre-up-script and net-down-script options

tpaukrt:
These options allow a user to specify paths to scripts usually located at /etc/ppp/net-init, /etc/ppp/net-pre-up and /etc/ppp/net-down, similarly to the existing ip-up-script and ip-down-script options.

### DIFF
--- a/pppd/main.c
+++ b/pppd/main.c
@@ -363,6 +363,10 @@ main(int argc, char *argv[])
     struct protent *protp;
     char numbuf[16];
 
+    strlcpy(path_net_init, PPP_PATH_NET_INIT, MAXPATHLEN);
+    strlcpy(path_net_preup, PPP_PATH_NET_PREUP, MAXPATHLEN);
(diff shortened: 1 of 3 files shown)


paulusmack:
Can you give an example of where this is useful?

@jkroonza any comments on this PR?

jkroonza (approved):
@paulusmack To get everything consistent?  Other scripts already have similar options (eg ip-up-script - which incidentally I don't see in the manpage but is definitely in pppd/options.c).

In general whilst we don't vary it this way this is useful, eg, if you have incoming PPPoE clients vs L2TP and you want to handle them differently this can be utilized to achieve that in a simple manner.

Perhaps if we could get a single commit to bring all the various script paths up to this spec it would be great, I think it's a waste to do this one by one (or even category by category).

I'm in support of this though.

PR #521, opened 2024-09-23
Export remote_number into REMOTENUMBER environment variable.

jkroonza:


jkroonza:
@paulusmack would have preferred to keep this handling completely in auth.c but could not find something that's guaranteed to execute between options parsing and net-init execution.  Doesn't make sense for our use-case to have this available in net-init ... but I'd prefer to not be short-sighted.

Neustradamus:

jkroonza:
No code changes, just added Signed-off-by.  Works well in production.

PR #522, opened 2024-09-25
ci: Updated the Solaris action to benefit from the faster VM.

RICCIARDI-Adrien:
The total execution decreased by about 20%, sometimes even more.
The best execution time since now has been 2 minutes 52 seconds when the average time with the old VM is about 6 minutes.


Neustradamus:
@paulusmack: Have you seen this @RICCIARDI-Adrien PR?

PR #527, opened 2024-10-18
pppd: Remove promptpass plugin

paulusmack:
This is prompted by a number of factors:

* The handling of privileges and possibly file descriptors is flawed

* In many cases there will be no terminal accessible to the prompter program at the point where the prompter is run

* The passwordfd plugin does much the same thing but does it more cleanly and securely.


paulusmack:
Distro maintainers might want to take note of this, particularly if your distro installs pppd setuid-root. The concern is that if an unprivileged user can run pppd and somehow get the promptpass plugin loaded, the prompter program could perhaps regain root privileges because of the flaws in how the promptpass.c code drops privileges. Fortunately, because the 'plugin' option is privileged, the user can't load the promptpass plugin directly. Hopefully no distros include a file under $sysconfdir/ppp/peers that includes 'plugin promptpass.so'. In any case I suggest that the promptpass.so plugin be removed from being packaged by distros.

@Neustradamus perhaps you could tag this for distro maintainers so they see it.

Neustradamus:
@paulusmack: You have canceled your PR?

paulusmack:
Yes, Jaco Kroon wanted me to be as vague as possible in the public commit, so I'll remove it along with other old stuff and not mention the bug in dropping privileges.

PR #528, opened 2024-10-20
pppd: Fix printing 64-bit counters

tpaukrt:
Add support of format specifiers %lld and %llu to the function vslprintf and use the correct specifiers for printing 64-bit counters.


Neustradamus:
@paulusmack: Have you seen this @tpaukrt PR?

PR #533, opened 2024-11-07
pppd: Fix build without OpenSSL

tpaukrt:
The symbol OPENSSL_VERSION_NUMBER is not defined when pppd is compiled without OpenSSL support, so it evaluates to zero. This results in the following linker error:

crypto.c:241: undefined reference to `ERR_free_strings'

### DIFF
--- a/pppd/crypto.c
+++ b/pppd/crypto.c
@@ -43,7 +43,6 @@
 #ifdef PPP_WITH_OPENSSL
 #include <openssl/opensslv.h>
 #include <openssl/err.h>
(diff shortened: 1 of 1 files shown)


Neustradamus:
@paulusmack: What do you think about @tpaukrt PR?

cc: @rascheel.

Linked to:
- https://github.com/ppp-project/ppp/issues/534

Neustradamus:
Thanks to @tpaukrt for your PR and thanks to @paulusmack for merging!

PR #538, opened 2024-12-12
disable holdoff for reconnect on timeout or link failure

anphsw:
This PR will cover cases where connection must be reestablished as fast as possible to descrease network downtime:
- connect time expired (ex: connection time limit for billing purposes or dynamic IP change)
- user request (ex: ISP peer shutting down and we must reconnect to another available in pool)
- connection lost (ex: we got cable disconnection or packet loss for a few seconds)


Neustradamus:
@paulusmack: Have you seen this @anphsw PR?

anphsw:
Updated.

Issue #519, opened 2024-09-18
Naming of the tarball release: https://github.com/paulusmack/ppp/archive/ppp-2.5.1.tar.gz


rubyFeedback:
Hey guys,

Using this URL:

    https://github.com/paulusmack/ppp/archive/ppp-2.5.1.tar.gz

When I extract this on the commandline, I get this directory:

    ppp-ppp-2.5.1

So the ppp appears twice.

I believe the more natural name would be to use the same name as by the URL,
in other words:

    ppp-2.5.1

Most other projects also do this. I assume some script may be automatically
renaming things, because a human being will probably rarely prepend a
'ppp-'. It also simplifies downstream scripts if the simple rule "basename of
the remote URL is the directory name when distracted".

Neustradamus:
@paulusmack: Can you look?

paulusmack:
I'm not impressed with github's release mechanisms. Besides this naming snafu, it also doesn't include things like the `./configure` script which aren't checked in to git, but get generated by autogen.sh.

I recommend you get ppp-2.5.1.tar.gz from https://download.samba.org/pub/ppp/ and use that.

jkroonza:
Not to mention the .tar.gz at https://download.samba.org/pub/ppp/ppp-2.5.1.tar.gz is double-compressed:

```
jkroon@plastiekpoot ~ $ wget https://download.samba.org/pub/ppp/ppp-2.5.1.tar.gz
--2024-09-19 11:49:33--  https://download.samba.org/pub/ppp/ppp-2.5.1.tar.gz
Resolving download.samba.org... 2a01:4f8:192:486::2:3, 144.76.82.148
Connecting to download.samba.org|2a01:4f8:192:486::2:3|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 994693 (971K) [application/gzip]
Saving to: ‘ppp-2.5.1.tar.gz’

ppp-2.5.1.tar.gz         100%[==================================>] 971.38K   173KB/s    in 5.6s    

2024-09-19 11:49:40 (173 KB/s) - ‘ppp-2.5.1.tar.gz’ saved [994693/994693]

jkroon@plastiekpoot ~ $ file ppp-2.5.1.tar.gz 
ppp-2.5.1.tar.gz: gzip compressed data, max compression, from Unix, original size modulo 2^32 995838
jkroon@plastiekpoot ~ $ gunzip < ppp-2.5.1.tar.gz | file -
[...]

PR #539, opened 2024-12-18
.gitattributes to exclude .git* files from git-archive.

jkroonza:
This relates to #519 and addresses the portion where "we can also exclude things like /.github/ from the release".


Neustradamus:

Issue #523, opened 2024-09-30
Configuring with --enable-microsoft-extensions=no breaks the compile


paulusmack:
After configuring with `--enable-microsoft-extensions=no`, running make gives:

```
  CC       pppd-peap.o
peap.c: In function ‘peap_init’:
peap.c:522:12: error: ‘struct peap_state’ has no member named ‘chap’
  522 |         psm->chap = chap_find_digest(CHAP_MICROSOFT_V2);
      |            ^~
make[2]: *** [Makefile:1520: pppd-peap.o] Error 1
```

@enaess ?

Neustradamus:
@enaess: Have you seen this @paulusmack ticket?

Neustradamus:
@enaess: Have you seen this @paulusmack ticket?

enaess:
If you disable Microsoft extensions then you might as well disable peap in the configuration too. Mschap is the only working peap method at this point.

paulusmack:
So then we should make peap depend on ms-extensions at the configure stage...

enaess:
@paulusmack for the time being, I think that would be appropriate. There might be a time in the future someone would add e.g. PEAP-SHA1 or something similar, which should allow configuring PEAP with whatever it supports. But since MsChapv2 is the only supported PEAP mechanisms now, then yes ...

Neustradamus:
Hope it can be added before a 2.5.1.1 or 2.5.2 release :)

Issue #433, opened 2023-07-03
--disable-microsoft-extensions requires --disable-peap


willyamcts:
When disable Microsoft Extension is required disable PEAP, both are not dependent correct?

if not using `--disable-peap` the error happens in compilation:

` # ./autogen.sh --prefix=/usr/local/pppd --disable-microsoft-extensions`
```
peap.c: In function ‘peap_init’:
peap.c:522:12: error: ‘struct peap_state’ has no member named ‘chap’
  522 |         psm->chap = chap_find_digest(CHAP_MICROSOFT_V2);
      |            ^~
make[2]: *** [Makefile:1492: pppd-peap.o] Error 1
```

Neustradamus:
@paulusmack, @enaess: Have you seen this ticket?

enaess:
Right now, the only peap method supported is mschapv2, so yes they are sort of dependent on each other. Technically, in the future they can be disjoint...

Neustradamus:
@enaess: Can you look it?

Neustradamus:
@enaess: Have you progressed on it?

enaess:
No progress, we can still ship 2.5.1 without this. People would just need to know, PEAP with CHAPv2 is the only thing that woks right now, and if you disable the CHAP extension, PEAP won't authenticate. However, in the future you could probably use PEAP-PAP or something there alike.

paulusmack:
I should add a note to the README about this.

PR #540, opened 2024-12-30
configure: Make PEAP depend on microsoft extensions

paulusmack:
Since our only PEAP mechanism is MS-CHAP v2, the compile fails if we enable PEAP without MS extensions.  Therefore, for now make PEAP depend on MS extensions, so if MS extensions are disabled, PEAP is automatically disabled.


paulusmack:
@enaess does this look right to you? Is there a better way to do it?

Neustradamus:
Linked to:
- https://github.com/ppp-project/ppp/issues/433
- https://github.com/ppp-project/ppp/issues/523

PR #544, opened 2025-01-07
Rework default route handling.

### LINKED ISSUE #473: rethinking the ppp defaultroute* options
The relevant documentation of the defaultroute options:

       defaultroute
              Add  a  default  route to the system routing tables, using the peer as the gateway,
              when IPCP negotiation is successfully completed.  This entry is  removed  when  the
              PPP  connection  is broken.  This option is privileged if the nodefaultroute option
              has been specified.

       defaultroute-metric
              Define the metric of the defaultroute and only add it if there is no other  default
              route  with the same metric.  With the default value of -1, the route is only added
              if there is no default route at all.

       replacedefaultroute
              This  option  is a flag to the defaultroute option. If defaultroute is set and this
              flag is also set, pppd replaces an existing default  route  with  the  new  default
              route.  This option is privileged.

(An appenddefaultroute option would also be convenient but let's not over-complicate things).

I think we should *ignore* the IPv6 situation, defaultroute6 technically already violates the standards as I understand them, but it's been stated that providers are messing around, so let's just leave that as is.  Also, this could *technically* be handled more flexibly from ipv6-up.  Same for ip-up actually come to think about it.

So there are a few valid combinations of options:

1.  defaultroute
2.  defaultroute+defaultroute-metric
3.  defaultroute+replacedefaultroute
4.  defaultroute+defaultroute-metric+replacedefaultroute

So I believe for these combinations the behaviour should be as follows (Ignoring 3 which is problematic):

1. If any default route exist, don't add one, if none exist, add with metric 0.
2. If a default route with the same metric exist, don't add one, else add with relevant metric.
4. If a default route with the same metric exist, remove it (restore on ipv4 down) and re-add via ppp.

A further problem with replacedefaultroute is that it's possible in linux to have *multiple* exact same routes at the same metric, so technically even 4 is a problem above:

```
plastiekpoot [09:01:41] ~ # ip ro sh
default via 192.168.42.1 dev wlp2s0 proto dhcp src 192.168.42.21 metric 2002 
...
plastiekpoot [09:02:05] ~ # ip route append default via 192.168.42.1 dev wlp2s0 proto static src 192.168.42.21 metric 2002
plastiekpoot [09:02:10] ~ # ip ro sh
default via 192.168.42.1 dev wlp2s0 proto dhcp src 192.168.42.21 metric 2002 
default via 192.168.42.1 dev wlp2s0 proto static src 192.168.42.21 metric 2002 
...
```

So with replacedefaultroute - should we remove *all* matching default routes or only the first matching one?

Which leads into my logic issue with 3 above.  If there is only one other default route it's all good and well, we remove it and re-add later, but what if there are multiple such routes?

I guess the replacedefaultroute option is problematic regardless.  We should only ever add routes and remove our own, the system administrator should manage priority via metrics.

So really the question is if a matching route with the same metric exist, do we append or not?  Does yet another separate option for that even make sense or should we simply append our route *regardless* - as such - nuke the replacedefaultroute option entirely, and append another default route at the specified metric and move on, thus reduce the defaultroute options to "defaultroute" and "defaultroute-metric" where if defaultroute is given we ALWAYS APPEND a default route at the specified defaultroute-metric (which then defaults to 0).

If the system administrator wants more control than that - have them rely on ip-(pre-)up or net-pre-up and let them deal with it there?

Whilst the existing logic is definitely broken, I'm not convinced that the PR brings things closer to the documented behaviour either, nor am I sure the documented behaviour is well defined.

_Originally posted by @jkroonza in https://github.com/ppp-project/ppp/pull/472#pullrequestreview-1814815012_
            
In addition, I believe the IPv6 defaultroute6 option should be brought in line with whatever it's decided to do with IPv4.

jkroonza:
1.  replacedefaultroute option is being removed.
2.  default route will now always be appended at defaultroute-metric (both IPv4 and IPv6) if defaultroute{.6} is given.

Closes: #473

### DIFF
--- a/pppd/ipcp.c
+++ b/pppd/ipcp.c
@@ -74,6 +74,8 @@ u_int32_t netmask = 0;		/* IP netmask to set on interface */
 bool	disable_defaultip = 0;	/* Don't use hostname for default IP adrs */
 bool	noremoteip = 0;		/* Let him have no IP address */
 
+unsigned dfl_route_metric = 0;	/* metric of the default route to set over the PPP link */
+
 ip_up_hook_fn *ip_up_hook = NULL;
 ip_down_hook_fn *ip_down_hook = NULL;
 ip_choose_hook_fn *ip_choose_hook = NULL;
@@ -136,6 +138,8 @@ static int setvjslots (char **);
 static int setdnsaddr (char **);
 static int setwinsaddr (char **);
 static int setnetmask (char **);
+static int replacedefaultroute_nonfunctional();
+
 int setipaddr (char *, char **, int);
 static void printipaddr (struct option *, void (*)(void *, char *,...),void *);
 
@@ -196,15 +200,20 @@ static struct option ipcp_option_list[] = {
       "disable defaultroute option", OPT_ALIAS | OPT_A2CLR,
       &ipcp_wantoptions[0].default_route },
 
+    { "defaultroute-metric", o_int, &dfl_route_metric,
+      "Metric to use for the default route (Linux only; default 0)",
+      OPT_PRIV|OPT_LLIMIT|OPT_INITONLY, NULL, 0, -1 },
+
 #ifdef __linux__
-    { "replacedefaultroute", o_bool,
-				&ipcp_wantoptions[0].replace_default_route,
-      "Replace default route", OPT_PRIV | 1
-    },
-    { "noreplacedefaultroute", o_bool,
-				&ipcp_wantoptions[0].replace_default_route,
-      "Do not replace default route", 0 },
+	{ "replacedefaultroute", o_special_noarg,
+		&replacedefaultroute_nonfunctional,
+		"Removed option to replace default route", OPT_PRIV
+	},
+	{ "noreplacedefaultroute", o_special_noarg,
+		&replacedefaultroute_nonfunctional,
+		"Removed option to not replace default route", 0 },
 #endif
+
     { "proxyarp", o_bool, &ipcp_wantoptions[0].proxy_arp,
       "Add proxy ARP entry", OPT_ENABLE|1, &ipcp_allowoptions[0].proxy_arp },
     { "noproxyarp", o_bool, &ipcp_allowoptions[0].proxy_arp,
@@ -284,7 +293,7 @@ struct protent ipcp_protent = {
     ip_active_pkt
 };
 
-static void ipcp_clear_addrs (int, u_int32_t, u_int32_t, bool);
+static void ipcp_clear_addrs (int, u_int32_t, u_int32_t);
 static void ipcp_script (char *, int);	/* Run an up/down script */
 static void ipcp_script_done (void *);
 
@@ -437,7 +446,7 @@ setipaddr(char *arg, char **argv, int doit)
 	return 0;
     if (!doit)
 	return 1;
-  
+
     /*
      * If colon first character, then no local addr.
      */
@@ -459,7 +468,7 @@ setipaddr(char *arg, char **argv, int doit)
 	*colon = ':';
 	prio_local = option_priority;
     }
-  
+
     /*
      * If colon last character, then no remote addr.
      */
@@ -525,6 +534,14 @@ setnetmask(char **argv)
     slprintf(netmask_str, sizeof(netmask_str), "%I", mask);
 
     return (1);
+}
+
+static int
+replacedefaultroute_nonfunctional()
+{
+    ppp_option_error("replacedefaultroute and noreplacedefaultroute route options no longer have any effect.");
+    ppp_option_error("Please refer to the man page for defaultroute and defaultroute-metric.");
+    return 1;
 }
 
 int
@@ -851,7 +868,7 @@ ipcp_addci(fsm *f, u_char *ucp, int *lenp)
     ADDCIWINS(CI_MS_WINS1, go->req_wins1, go->winsaddr[0]);
 
     ADDCIWINS(CI_MS_WINS2, go->req_wins2, go->winsaddr[1]);
-    
+
     *lenp -= len;
 }
 
@@ -1456,7 +1473,7 @@ ipcp_reqci(fsm *f, u_char *inp,	int *len, int reject_if_disagree)
      * Reset all his options.
      */
     BZERO(ho, sizeof(*ho));
-    
+
     /*
      * Process all his options.
      */
@@ -1566,7 +1583,7 @@ ipcp_reqci(fsm *f, u_char *inp,	int *len, int reject_if_disagree)
 		wo->req_addr = 0;	/* don't NAK with 0.0.0.0 later */
 		break;
 	    }
-	
+
 	    ho->neg_addr = 1;
 	    ho->hisaddr = ciaddr1;
 	    break;
@@ -1610,7 +1627,7 @@ ipcp_reqci(fsm *f, u_char *inp,	int *len, int reject_if_disagree)
 		orc = CONFNAK;
             }
             break;
-	
+
 	case CI_COMPRESSTYPE:
 	    if (!ao->neg_vj ||
 		(cilen != CILEN_VJ && cilen != CILEN_COMPRESS)) {
@@ -1629,7 +1646,7 @@ ipcp_reqci(fsm *f, u_char *inp,	int *len, int reject_if_disagree)
 	    ho->vj_protocol = cishort;
 	    if (cilen == CILEN_VJ) {
 		GETCHAR(maxslotindex, p);
-		if (maxslotindex > ao->maxslotindex) { 
+		if (maxslotindex > ao->maxslotindex) {
 		    orc = CONFNAK;
 		    if (!reject_if_disagree){
 			DECPTR(1, p);
@@ -1776,8 +1793,7 @@ ip_demand_conf(int u)
     if (!sifnpmode(u, PPP_IP, NPMODE_QUEUE))
 	return 0;
     if (wo->default_route)
-	if (sifdefaultroute(u, wo->ouraddr, wo->hisaddr,
-					    wo->replace_default_route))
+	if (sifdefaultroute(u, wo->ouraddr, wo->hisaddr))
 	    default_route_set[u] = 1;
     if (wo->proxy_arp)
 	if (sifproxyarp(u, wo->hisaddr))
@@ -1878,8 +1894,7 @@ ipcp_up(fsm *f)
      */
     if (demand) {
 	if (go->ouraddr != wo->ouraddr || ho->hisaddr != wo->hisaddr) {
-	    ipcp_clear_addrs(f->unit, wo->ouraddr, wo->hisaddr,
-				      wo->replace_default_route);
+	    ipcp_clear_addrs(f->unit, wo->ouraddr, wo->hisaddr);
 	    if (go->ouraddr != wo->ouraddr) {
 		warn("Local IP address changed to %I", go->ouraddr);
 		ppp_script_setenv("OLDIPLOCAL", ip_ntoa(wo->ouraddr), 0);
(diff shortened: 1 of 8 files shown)


pppd/options.c
```diff
@@ -332,7 +333,10 @@ struct option general_options[] = {
       OPT_A2PRINTER | OPT_NOPRINT, (void *)user_unsetprint },
 
     { "defaultroute-metric", o_int, &dfl_route_metric,
```
jkroonza, line 335: Need to add o_uint option.

pppd/sys-linux.c
```diff
@@ -2165,72 +2131,72 @@ int have_route_to(u_int32_t addr)
...
-    }
+    nlreq.nlh.nlmsg_len = sizeof(nlreq);
+    nlreq.nlh.nlmsg_type = operation;
+    nlreq.nlh.nlmsg_flags = NLM_F_REQUEST | NLM_F_ACK | NLM_F_CREATE;
+    if (operation == RTM_NEWROUTE)
+	nlreq.nlh.nlmsg_flags |= NLM_F_APPEND;
+
+    nlreq.rtmsg.rtm_family = family;
+    nlreq.rtmsg.rtm_table = RT_TABLE_MAIN;
+    nlreq.rtmsg.rtm_protocol = RTPROT_UNSPEC;
```
jkroonza, line 2165: By default iproute2 uses RTPROT_BOOT, somehow that doesn't seem right, but with this we get:

```
default *proto unspec* scope link metric 10000 
```

strace from ip ro ad shows (reformatted for readability):

```
sendmsg(3, {
  msg_name={
    sa_family=AF_NETLINK,
    nl_pid=0,
    nl_groups=00000000
  },
  msg_namelen=12,
  msg_iov=[
    {iov_base=[
      {
        nlmsg_len=44,
        nlmsg_type=RTM_NEWROUTE,
        nlmsg_flags=NLM_F_REQUEST|NLM_F_ACK|NLM_F_CREATE|NLM_F_APPEND,
        nlmsg_seq=1736279032,
        nlmsg_pid=0
      },
      {
        rtm_family=AF_INET,
        rtm_dst_len=0,
        rtm_src_len=0,
        rtm_tos=0,
        rtm_table=RT_TABLE_MAIN,
        **rtm_protocol=RTPROT_BOOT**,
        rtm_scope=RT_SCOPE_LINK,
        rtm_type=RTN_UNICAST,
        rtm_flags=0
      },
      [
        [{nla_len=8, nla_type=RTA_PRIORITY}, 10000],
        [{nla_len=8, nla_type=RTA_OIF}, if_nametoindex("ppp0")]
      ]
    ],
    iov_len=44
  }],
  msg_iovlen=1, msg_controllen=0, msg_flags=0}, 0) = 44
```

jkroonza (commented):
There is also this reference to replacedefaultroute in pppd/ipcp.c in ipcp_clear_addrs:

```
2055     /* If replacedefaultroute, sifdefaultroute will be called soon
2056      * with replacedefaultroute set and that will overwrite the current
2057      * default route. This is the case only when doing demand, otherwise
2058      * during demand, this cifdefaultroute would restore the old default
2059      * route which is not what we want in this case. In the non-demand
2060      * case, we'll delete the default route and restore the old if there
2061      * is one saved by an sifdefaultroute with replacedefaultroute.
2062      */
```

I *reckon* this means further code adjustments may be required around there.

Neustradamus:
@paulusmack, @sthibaul: Can you look this @jkroonza PR which solves one ticket :)

Neustradamus:
@paulusmack, @sthibaul: Can you look this @jkroonza PR which solves one ticket :)

jkroonza:
@paulusmack I would appreciate some feedback on this, especially in the light off the concept #546.  Given that I'm wondering if the newly added netlink route/add options should take a prefix in addition.  sockaddr_t + socklet_t + some unsigned option.

Then a pair of sifroute functions for adding arbitrary routes which can be called from the radius module.

Currently we handle static routes external to pppd in ip-up and ipv6 up, which is fine, and that can stay there, but we're going to need plumbing for IPv6 PDs anyway, so might just as well handle Framed-Route and IPv6-Framed-Route radius options too.

paulusmack:
Is 0 really a good choice for the default metric? Shouldn't the metric be somewhat reflective of the bandwidth of the network device (which is often lower for PPP links than other types of network)?

jkroonza:
That's the current default.  And how do you know what the bandwidth on the client side of a ppp link will be?

pppd/options.c
paulusmack: Why? Do we need to allow values >= 2^31?

pppd/ipcp.c
```diff
@@ -196,15 +196,6 @@ static struct option ipcp_option_list[] = {
...
 
-#ifdef __linux__
-    { "replacedefaultroute", o_bool,
-				&ipcp_wantoptions[0].replace_default_route,
-      "Replace default route", OPT_PRIV | 1
-    },
-    { "noreplacedefaultroute", o_bool,
-				&ipcp_wantoptions[0].replace_default_route,
-      "Do not replace default route", 0 },
-#endif
```
paulusmack, line 207: In general I prefer not to break existing working setups by removing options, if it is reasonable for the option to become a no-op (or at worst emit a warning). Could these be made no-ops?

pppd/pppd.8
```diff
@@ -121,15 +121,8 @@ This entry is removed when the PPP connection is broken.  This option
...
-is no other default route with the same metric.  With the default
-value of -1, the route is only added if there is no default route at
-all.
-.TP
-.B replacedefaultroute
-This option is a flag to the defaultroute option. If defaultroute is
-set and this flag is also set, pppd replaces an existing default route
-with the new default route.  This option is privileged.
+Define the metric of the \fIdefaultroute\fR.  By default the default route will
+be added with a metric of 0.
```
paulusmack, line 125: Perhaps add here that this option is privileged

paulusmack:
> There is also this reference to replacedefaultroute in pppd/ipcp.c in ipcp_clear_addrs:
> 
> ```
> 2055     /* If replacedefaultroute, sifdefaultroute will be called soon
> 2056      * with replacedefaultroute set and that will overwrite the current
> 2057      * default route. This is the case only when doing demand, otherwise
> 2058      * during demand, this cifdefaultroute would restore the old default
> 2059      * route which is not what we want in this case. In the non-demand
> 2060      * case, we'll delete the default route and restore the old if there
> 2061      * is one saved by an sifdefaultroute with replacedefaultroute.
> 2062      */
> ```
> 
> I _reckon_ this means further code adjustments may be required around there.

I think being able to have multiple default routes in the system makes this easier, actually, in that we can add a default route and remove it (as long as we are sure to remove the one we added) and leave the system as it was before. So that whole comment can probably be deleted.

pppd/ipcp.c
jkroonza: Not sure why I can't reply by the 2^31 issue, the kernel type is unsigned, but sure, as long as we can restrict to positive integer that's fine by me.  If you need metrics >=2^31 you can do the relevant patching.

Re no-op, will handle.  Normally I'd argue for "let it break so that someone knows something need to change", but in this case since it might well relate to something that the admin needs to gain access to the system ... I'm happy to concede.

pppd/pppd.8
jkroonza: Done.

pppd/options.c
jkroonza: To match kernel unsigned option here.  To be realistic I highly doubt a value >=2^31 makes sense, so let's just leave this as is for now.

pppd/sys-linux.c
jkroonza: No opinions, so use use RTPROT_BOOT which I think is "implied" by ioctl mechanism anyway, and this closes to existing behaviour.

pppd/ipcp.c
jkroonza: @paulusmack please check if you're happy with the change.

I took the opportunity to move options out of pppd/options.c and into the NCPs where they belong at the same time (at least those that I touched).  I trust this will be in-order as it honestly makes more sense to have IPCP options reside in ipcp.c and IPv6CP options in ipv6cp.c.

I did (from the initial PR) include some whitespace only changes into the PR.  If that's an issue I can rework to exclude that.  Didn't make the same mistake on other files (vim automatically cleans up trailing whitespace and a few others on my system).

jkroonza:
@paulusmack if/when you have time, further pointers would be appreciated thank you.

paulusmack (changes requested):
Looks pretty good overall, just the details of a message to fix.

pppd/ipcp.c
```diff
@@ -527,6 +536,13 @@ setnetmask(char **argv)
     return (1);
 }
 
+static int
+replacedefaultroute_nonfunctional()
+{
+    ppp_option_error("replacedefaultroute and noreplacedefaultroute route options no longer has any efffect.  Please refer to the man page for defaultroute and defaultroute-metric.");
```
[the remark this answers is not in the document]
jkroonza: Should be better now.

Neustradamus:
Good job, thanks to @jkroonza for this PR and @paulusmack for reviewing/merging!

sithglan:
Thank you. This fixes an issue I had with a 15 minute reconnect after 24 hour disconnect. This was the issue I had with 2.5.2 and was fixed with 2.5.3 which incorporate this pull request:

  /etc/ppp/peers/tng had defaultroute6. Before installing the IPv6 default
route, pppd's defaultroute6_exists() reads /proc/net/ipv6_route line-by-line —
a seq_file where every 4 KB read re-walks the kernel FIB trie across all
tables. Since ~May 22, bird carries a full BGP IPv6 table (~249k routes, kernel
table 100). After each reconnect, bird re-converges and re-inserts routes at
the same time pppd scans — a mutual livelock (verified live: pppd spinning in
userspace state=R/empty kernel stack, bird at 52 % CPU with maxed 212 KB
netlink Recv-Q, table growing ~120 routes/s). The scan took up to 15 minutes;
pppd couldn't answer the BRAS's LCP echoes meanwhile, so the BRAS killed the
session, and the cycle repeated — each round shorter as the table shrank —
until one attempt completed in time.

jkroonza:
> Thank you. This fixes an issue I had with a 15 minute reconnect after 24 hour disconnect. This was the issue I had with 2.5.2 and was fixed with 2.5.3 which incorporate this pull request:
> 
> /etc/ppp/peers/tng had defaultroute6. Before installing the IPv6 default route, pppd's defaultroute6_exists() reads /proc/net/ipv6_route line-by-line — a seq_file where every 4 KB read re-walks the kernel FIB trie across all tables. Since ~May 22, bird carries a full BGP IPv6 table (~249k routes, kernel table 100). After each reconnect, bird re-converges and re-inserts routes at the same time pppd scans — a mutual livelock (verified live: pppd spinning in userspace state=R/empty kernel stack, bird at 52 % CPU with maxed 212 KB netlink Recv-Q, table growing ~120 routes/s). The scan took up to 15 minutes; pppd couldn't answer the BRAS's LCP echoes meanwhile, so the BRAS killed the session, and the cycle repeated — each round shorter as the table shrank — until one attempt completed in time.

Not the motivation behind the change but I can see how that would sort it ... not often I see uplink pppd combined with BGP routing (and especially not full tables) updates.  That is definitely unusual, and especially when combined with defaultroute and defaultroute6 - I'd expect you to not rely on default routes at all then - or get that too via BGP :).

sithglan:
> Not the motivation behind the change but I can see how that would sort
> it ... not often I see uplink pppd combined with BGP routing (and
> especially not full tables) updates.  That is definitely unusual, and
> especially when combined with defaultroute and defaultroute6 - I'd
> expect you to not rely on default routes at all then - or get that too
> via BGP :).

This particular system has a 1 Gbit/s symmetric dark fibre driven by
pppd and relies on at least an IPv4 default route. It has several IPv4
wireguard tunnels which are used for bgp peering. The IPv6 routes end up
in the non default route table and are used with policy based routing.
Anyway I wanted to write a patch for pppd but you had already done it.
:-) I'm now running two of these systems with ppp 2.5.3.

PR #548, opened 2025-02-27
Fixed building with GCC 15

yarda:
GCC 15 defaults to C23 which does not allow K&R declarations.

Credit Yaakov Selkowitz in:
https://src.fedoraproject.org/rpms/ppp/pull-request/12


yarda:
IMHO K&R is now probably not used so it could be probably safely ported to recent standard.

PR #550, opened 2025-03-05
pppd: supress spurious ESRCH on route removal.

jkroonza:

Issue #546, opened 2025-01-23
DHCPv6 proxy prefix delegation/address assignments


jkroonza:
Hi,

I've looked a bit into how to best handle IPv6 requests here.

KEA (ISC replacement) DHCPv6 server is perfectly capable of issuing prefix delegations, and maintaining persistence and all of than, but it really, really sucks at dynamic interfaces.  It's probably OK if you only have a handful of connections, but we had extreme problems just to get it configured in this way.

I then started on a DHCPv6 proxy (yes, there were existing ones, none of which has active maintenance), but then realised that getting PD delegations and all that into radius, whilst it can be done externally, it would overall be simpler if we can build the DHCPv6 proxy (which also knows how to handle radius IPv6 static assignments and delegations) straight into pppd.

I've started playing in https://github.com/jkroonza/dhcpv6-relay towards a standalone relay, but the complexities involved with not being part of the pppd daemon is tremendous once radius interactions becomes involved.

Using strace the pppd daemon sits mainly in this system call.
```
pselect6(13, [8 10 12], NULL, [8 10 12], {tv_sec=29, tv_nsec=999946000}, NULL
```

In this case fd 8 is a pipe whilst 10 and 12 are /dev/ppp.  The write-end of the pipe is at fd=9 and I suspect this is to enable signals to interrupt the select.

This will likely result in work on top of #544 where the function to add/remove the default route will start accepting a specific prefix.  This can happen before #544 but has the obvious rebasing caveats.

It looks like it's fairly trivial to add a file descriptor using add_fd to this set of descriptors, but I see no obvious way to associate an event handler with this file descriptor.

Would the following set of PRs (or commits in a single PR) be acceptable?

1. Adjust add_fd to take two additional arguments:  a function pointer "void (*callback)(int fd, void* pvt)", and a "void* pvt"?  Alterantively a new function in pppd/main.c add_fd_event_handler that has these semantics, and builds on top of add_fd but instead in handle_events calls the callback functions if these file descriptors has input available?  (the latter is probably a cleaner approach, and won't break backwards compatibility with external modules)
1. Adjust the route add/remove code with support for a prefix.
1. In radius module add support Framed-Route (when it's encountered add a route to the interface, should remove too when link goes down).
1. In radius module add support for Framed-IPv6-Route (similar).
1. Make note of any of the following for DHCPv6 proxy module:
    1. Delegated-IPv6-Prefix
    1. DNS-Server-IPv6-Address
[...]

PR #551, opened 2025-03-06
dhcpv6relay plugin.

jkroonza:
After the plugin is loaded, it just needs dhcpv6relay-server option to know where to relay to.

Server will (current planning) need to support option 135 which tells it to which port to send return responses (it will still send from port 547 typically).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -330,6 +330,7 @@ AC_CONFIG_FILES([
     pppd/plugins/pppoatm/Makefile
     pppd/plugins/pppol2tp/Makefile
     pppd/plugins/radius/Makefile
+    pppd/plugins/dhcpv6relay/Makefile
     pppdump/Makefile
     pppstats/Makefile
     scripts/Makefile
--- a/pppd/Makefile.am
+++ b/pppd/Makefile.am
@@ -82,6 +82,7 @@ pppd_SOURCES = \
     lcp.c \
     magic.c \
     main.c \
+    event-handler.c \
     options.c \
     session.c \
     tty.c \
--- a/pppd/event-handler.c
+++ b/pppd/event-handler.c
@@ -1,0 +1,130 @@
+/*
+ * event-handler.c - generic select() based event handler.  Should be system
+ * independent.
+ *
+ * Copyright (c) 1994-2025 Paul Mackerras. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *
+ * THE AUTHORS OF THIS SOFTWARE DISCLAIM ALL WARRANTIES WITH REGARD TO
+ * THIS SOFTWARE, INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS, IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY
+ * SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN
+ * AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING
+ * OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ * Derived from sys-linux.c and sys-solaris.c by Jaco Kroon <jaco@uls.co.za>.
+ */
+#include <limits.h>
+#include <stddef.h>
+#include <errno.h>
+#include <sys/select.h>
+
+#include "pppd.h"
+#include "pppd-private.h"
+
+struct event_handler {
+    struct event_handler* next;
+    int fd;
+    event_cb cb;
+    void* ctx;
+};
+
+static fd_set in_fds;		/* set of fds that wait_input waits for */
+static int max_in_fd;		/* highest fd set in in_fds */
+static struct event_handler* handlers;
+static int called_remove;
+
+/********************************************************************
+ *
+ * wait_input - wait until there is data available,
+ * for the length of time specified by *timo (indefinite
+ * if timo is NULL).
+ */
+
+void wait_input(struct timeval *timo)
+{
+    fd_set ready, exc;
+    int n;
+    struct event_handler* h = handlers, *nh;
+
+    called_remove = 0;
+    ready = in_fds;
+    exc = in_fds;
+    n = select(max_in_fd + 1, &ready, NULL, &exc, timo);
+    if (n < 0 && errno != EINTR)
+	fatal("select: %m");
+
+    while (h) {
+	nh = h->next;
+	if (FD_ISSET(h->fd, &ready)) {
+	    FD_CLR(h->fd, &ready); /* clear so that if we need to re-iterate we won't call again */
+	    h->cb(h->fd, h->ctx);
+
+	    if (called_remove) {
+		nh = handlers;
+		called_remove = 0;
+	    }
+	}
+	h = nh;
+    }
+}
+
+/*
+ * add_fd - add an fd to the set that wait_input waits for.
+ */
+void add_fd(int fd)
+{
+    if (fd >= FD_SETSIZE)
+	fatal("internal error: file descriptor too large (%d)", fd);
+    FD_SET(fd, &in_fds);
+    if (fd > max_in_fd)
+	max_in_fd = fd;
+}
+
+void add_fd_callback(int fd, event_cb cb, void* ctx)
+{
+    struct event_handler *n = malloc(sizeof(*n));
+    n->next = handlers;
+    n->fd = fd;
+    n->cb = cb;
+    n->ctx = ctx;
+    handlers = n;
+    add_fd(fd);
+}
+
+/*
+ * remove_fd - remove an fd from the set that wait_input waits for.
+ */
+void remove_fd(int fd)
+{
+    struct event_handler** h, *t;
+    FD_CLR(fd, &in_fds);
+
+    for (h = &handlers; *h; h = &(*h)->next) {
+	if (fd == (*h)->fd) {
+	    called_remove = 1;
+	    t = (*h)->next;
+	    free(*h);
+	    *h = t;
+	    return;
+	}
+    }
+}
+
+void event_handler_init()
+{
+    FD_ZERO(&in_fds);
+    max_in_fd = 0;
+}
--- a/pppd/main.c
+++ b/pppd/main.c
@@ -497,6 +497,8 @@ main(int argc, char *argv[])
 	fd_devnull = i;
     }
 
+    event_handler_init();
+
     /*
      * Initialize system-dependent stuff.
      */
--- a/pppd/plugins/Makefile.am
+++ b/pppd/plugins/Makefile.am
@@ -17,5 +17,5 @@ winbind_la_LDFLAGS = $(PLUGIN_LDFLAGS)
 winbind_la_SOURCES = winbind.c
 
 if !SUNOS
-SUBDIRS = pppoe pppoatm pppol2tp radius
+SUBDIRS = pppoe pppoatm pppol2tp radius dhcpv6relay
 endif
--- a/pppd/plugins/dhcpv6relay/Makefile.am
+++ b/pppd/plugins/dhcpv6relay/Makefile.am
@@ -1,0 +1,8 @@
+pppd_plugin_LTLIBRARIES = dhcpv6relay.la
+pppd_plugindir = $(PPPD_PLUGIN_DIR)
+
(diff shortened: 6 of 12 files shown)


jkroonza:
This relates to #546 

I'm hoping that seeing the binding work here explains why I think this is easier "in-pppd" than outside.

Turns out here exists an DHCPv6 "option 135" which allows us to specify the target port for return traffic from the server.  I suspect KEA isn't going to like us sending from a port other than 546, not sure about dhcpd -6 (yet).  So might need to bind two "client" sockets, one for sending from port 546 and one for receiving (on some other port).  Will need to play a bit with that.

I will say that multicast is a pain in the backside.

@paulusmack if this provides you any insight/ideas, or raises questions, please let me know.

This is intended for pppd as a dial-in (server) service.  Client side is easily manageable already and there are hordes of tools to deal with that side of the equation.

Neustradamus:
@jkroonza: Nice :)

jkroonza:
> @jkroonza: Nice :)

Conceptually perhaps, but not there yet.

paulusmack:
> Would it be highly-objectionable to consolidate that into using the same mechanism for _both_ linux and solaris, and into say pppd/event_handler.c which still presents the add_fd(), remove_fd() and wait_input() calls, but also adds the following additional call:
> 
> ```
> typedef int (*event_cb)(int fd, void* ctx);
> add_fd_callback(int fd, event_cb cb, void* ctx);
> ```

I wouldn't object, if that would be helpful. I assume the single implementation would use poll().

jkroonza:
select() would be easier, but I can use poll() too if there's an objection to select?

jkroonza:
```03:02:42.499658 ppp0  In  IP6 (hlim 1, next-header UDP (17) payload length: 58) fe80::14.546 > ff02::1:2.547: [udp sum ok] dhcp6 solicit (xid=48614a (client-ID hwaddr type 1 18fd74159342) (option-request DNS-server) (elapsed-time 0) (rapid-commit) (IA_PD IAID:20 T1:1800 T2:2880))
	0x0000:  86dd 0000 0000 0198 0200 0000 0000 0000  ................
	0x0010:  0000 0000 6000 0000 003a 1101 fe80 0000  ....`....:......
	0x0020:  0000 0000 0000 0000 0000 0014 ff02 0000  ................
	0x0030:  0000 0000 0000 0000 0001 0002 0222 0223  .............".#
	0x0040:  003a 67ec 0148 614a 0001 000a 0003 0001  .:g..HaJ........
	0x0050:  18fd 7415 9342 0006 0002 0017 0008 0002  ..t..B..........
	0x0060:  0000 000e 0000 0019 000c 0000 0014 0000  ................
	0x0070:  0708 0000 0b40                           .....@
03:02:42.499749 lo    In  IP6 (flowlabel 0xd8a5e, hlim 64, next-header UDP (17) payload length: 102) ::1.53032 > ::1.547: [bad udp cksum 0x0079 -> 0xbda4!] dhcp6 relay-fwd (linkaddr=:: peeraddr=fe80::14 (opt_135) (relay-message (dhcp6 solicit (xid=48614a (client-ID hwaddr type 1 18fd74159342) (option-request DNS-server) (elapsed-time 0) (rapid-commit) (IA_PD IAID:20 T1:1800 T2:2880))))
	0x0000:  86dd 0000 0000 0001 0304 0006 0000 0000  ................
	0x0010:  0000 0000 600d 8a5e 0066 1140 0000 0000  ....`..^.f.@....
	0x0020:  0000 0000 0000 0000 0000 0001 0000 0000  ................
	0x0030:  0000 0000 0000 0000 0000 0001 cf28 0223  .............(.#
	0x0040:  0066 0079 0c00 0000 0000 0000 0000 0000  .f.y............
	0x0050:  0000 0000 0000 fe80 0000 0000 0000 0000  ................
	0x0060:  0000 0000 0014 0087 0002 cf28 0009 0032  ...........(...2
	0x0070:  0148 614a 0001 000a 0003 0001 18fd 7415  .HaJ..........t.
	0x0080:  9342 0006 0002 0017 0008 0002 0000 000e  .B..............
	0x0090:  0000 0019 000c 0000 0014 0000 0708 0000  ................
	0x00a0:  0b40                                     .@
```

We can relay already towards the server already.  This was with:

```
plugin dhcpv6relay.so
dhcpv6relay-server ::1
```

Other variations should work too.  If option 135 ends up not working I have two ideas, the simplest of which is to implement a pure proxy that can action the relevant relay on our behalf that does support option 135.  This should not be too difficult, and is definitely easier than implementing some "management agent" like previous suggestions.

Next up is to get KEA / ISC DHCPd to respond with a PD so that I can get working on:

1.  Decapsulating and forwarding the response.
2.  Adding relevant routes via sys-* (will need help on solaris side).  Routing will need to support both PD as well as address allocation.
3.  Potentially need to keep track of multiple addresses/prefixes issued, and set expiry timers to remove the routes when valid_lifetime expires (assuming that the client doesn't renew the addresses).
4.  Add option 37 aka remote-id (REMOTENUMBER) and 38 aka subscriber-id (PEERNAME).

Is there a maximum length on either of those options or should I consider further complicating the iovec structure?

jkroonza:
Option 37 + 38 added.  Note the string values inside the hex dump of the packet makes a lot more sense than the tcpdump decoded values.

```
13:50:42.706067 lo    In  IP6 (flowlabel 0x29224, hlim 64, next-header UDP (17) payload length: 125) ::1.42205 > ::1.547: [bad udp cksum 0x0090 -> 0x9f53!] dhcp6 relay-fwd (linkaddr=:: peeraddr=fe80::14 (opt_135) (Remote-ID 825831982 3136382e312e31...) (Subscriber-ID 74657374...) (relay-message (dhcp6 solicit (xid=ef5a63 (client-ID hwaddr type 1 18fd74159342) (option-request DNS-server) (elapsed-time 300) (rapid-commit) (IA_PD IAID:20 T1:1800 T2:2880))))
	0x0000:  86dd 0000 0000 0001 0304 0006 0000 0000  ................
	0x0010:  0000 0000 6002 9224 007d 1140 0000 0000  ....`..$.}.@....
	0x0020:  0000 0000 0000 0000 0000 0001 0000 0000  ................
	0x0030:  0000 0000 0000 0000 0000 0001 a4dd 0223  ...............#
	0x0040:  007d 0090 0c00 0000 0000 0000 0000 0000  .}..............
	0x0050:  0000 0000 0000 fe80 0000 0000 0000 0000  ................
	0x0060:  0000 0000 0014 0087 0002 a4dd 0025 000b  .............%..
	0x0070:  3139 322e 3136 382e 312e 3100 2600 0474  192.168.1.1.&..t
	0x0080:  6573 7400 0900 3201 ef5a 6300 0100 0a00  est**...2..Zc.....
	0x0090:  0300 0118 fd74 1593 4200 0600 0200 1700  .....t..B.......
	0x00a0:  0800 0201 2c00 0e00 0000 1900 0c00 0000  ....,...........
	0x00b0:  1400 0007 0800 000b 40                   ........@
```

jkroonza:
Tested working up to this point with kea on server side, bound to ::1, and Mikrotik router as client over incoming L2TP connection (via xl2tpd).

DHCP itself works, just need to add routing to resolve this section from the RFC:

https://datatracker.ietf.org/doc/html/rfc8415#section-19.1.3

For the moment I'm going to stick with the assumption that in the pppd use-case we're looking for a PD, and not an address assignment, as in my mind at least it doesn't make sense to merely request a single address on a ppp link, since the CPE device router will invariably want a prefix to assign to the LAN, I'm going to focus on that use-case.  A single address request may make sense in VPN type scenarios where a host dials in from a remote location.  That may well end up being better served using RAs though, but a DHCP configuration *might* make more sense, eg, to serve such requests from a single /64

I have spotted one case (ISC dhcpcd -6) where the client expects to first see an route-advertisement in response to a route-solicitation.  This can be dealt with with a tool such as radvd or even frr.  That said, it may be useful to have a ipv6ra module added to pppd as well in addition to this dhcpv6relay, just want to look into radvd - the problem is our use case is of such a nature that even if radvd will work with patterns it may well end up being in-discriminant, but frr should do the trick by disabling ra announce for the interface during net-pre-up, and then selectively enabling in ipv6-up.  If that'll work for us, I'm obviously not going to dig too much further into this.

As an aside, due to the design of ISC dhcpd, it's unlikely going to work well alongside pppd in this configuration on the same host, and not at all on it's own (due to binding specific *broadcast ethernet* interfaces only during startup, whilst still having a [::]:547 bound unicast socket).

KEA is more configurable, but crashes when you try to bind it to a ppp interface, and having tried to engage with upstream, this doesn't seem likely to get fixed.  So my recommendation is KEA bound to a specific address, eg ::1 if you're on a single-host, and then use this module to relay to that.

jkroonza:
@paulusmack @Neustradamus 

I think I'm going to pause here for a day or two just to give time for feedback.  I *suspect* we can treat these two commits as:

1. Core changes - need proper review, and ideally testing beyond just me.
2. The extra module - probably less of an issue, suspect not too many users will need this, and those that would would like me action their own full blown testing regardless.  Not to say "be reckless", but this module will likely not be used by a great many number of people as it's focused very much on the ISP type industry, most of which probably use CISCO or some other highly paid commercial alternative (although I've started seeing a drive in the local industry to get away from those).  It *might* (with some further modifications) be useful to VPN solutions building on top of ppp (eg, ppp over l2tp over ipsec)

jkroonza:
My testing pans out:

1.  select() implementation hasn't failed in any way (on Linux at least).  Happy to redo to poll() but implementation will be more complicated.
2.  changes to sys-linux was tested again with client with defaultroute and defaultroute6, this showed a bug in the changes which was rectified.  sys-solaris needs an implementation but I've no way to test that, so would rather leave it with stub implementations that just generates and error indicating that it's non-functional.
3. dhcpv6relay module covers my requirements.  Things that could still be added (That I can think off):
3.1. deal with IA_ADDR delegation (single /128 address assignment).  This makes more sense for a VPN environment, and there I'm betting RAs are sufficient regardless.  Shouldn't be hard to add.
3.2. Multiple upstream DHCPv6 addresses (but you might just as well give this module a multicast address, plus logic of "send to all vs ).
3.3. Specific bind() address for upstream socket, rather than determining by way of connect().
3.4. External route manipulation scripts.

Possible change:  Turns out I don't need to remember until when a route is valid courtesy of the way ppp_(un)timeout() works.  So this field can be removed from the relevant structure, leaving just the prefix and length as the effective fields.  We do still need to know what all the routes are, since if IPV6CP gets torn down, we do need to clean up all the routes still.

Where to add documentation?  How are modules documented?  And how do people know which man pages to get to?  It seems the only plugin that's documented in the pppd man page is pppoe.so ... surely that's something we should improve?  Either have a separate man page for each pppd plugin (eg, pppd-dhcpv6relay), or document all of them in the main man page?

Everything else (eg, setting DNS servers, NTP servers, or any other information providable by DHCP for that matter) can be handled on KEA side (or ISC dhcpd if you're that way inclined, KEA seems to be more manageable, especially in a clustered environment due to options for shared SQL back-ends - which itself can be clustered again).  This module solves a few problems for us:

1.  DHCPv6 server not on the same host.
2.  KEA crashes when trying to bind to a ppp interface.  Bug logged, no response for years now.
3.  Dynamic reconfiguration of DHCP servers is best avoided based on testing.
4.  Avoids complicated dhcpv6-proxy process which does what this module does, but for several interfaces, and monitoring for interface addition/removal as it turns out is very, very complicated (which is probably why existing DHCPv6 servers doesn't do it).

jkroonza:
FWIW:  Just went into production a few hours ago, and other than the fix from o_wild => o_special (which was braindead stupid) things seems to be holding up so far.

jkroonza:
Rebased after #550 and if the core changes are acceptable, then I think this can be merged at this point.  As stated, this went into production yesterday, and we already have a small percentage of users that's obtaining IPv6 through thanks to this module.

Neustradamus:

paulusmack:
@jkroonza The code all looks OK to me. The only comment I have is that there should be copyright notices in the new files. For event-handler.c, since it's all code from sys-linux.c that I wrote, it should have my copyright notice from sys-linux.c in it, but not the CMU copyright. For the other new files, that's all your code, so you should put your copyright notice and license on them, preferably BSD-style for consistency with the rest of the package, but it's your choice.

One thing in the code that I noticed was that in dhcpv6relay_real_release_route(), if sifdelroute fails, we print an error saying that we failed to remove the route, but then we also print the notice() level message that the route was uninstalled. That seems like it could be confusing. Maybe there's a subtle difference between removing and uninstalling here, in which case the messages should be clearer, or else the notice() should be skipped in the error case.

jkroonza:
@paulusmack please check if that works better for you?

Whilst this is in play in production and works currently, I do suspect I need to add an additional RA component to this.  But lets get this merged so long, I need to action some more research on that front and then architect a design (ie, what functionality all from RA to support).  I'm inclined to keep it basic.

jkroonza:
Added code to deal with receiving Router Solicitation messages and responding with "Managed" Router Advertisement.

We bumped into a case where we're seeing incoming RS messages from the ppp interface, so cooked this up.  Turns out it's a somwhat misconfigured Linux router on the other end.  Will address this with the client, but regardless, this is useful.  Also suspect other routers my be sitting idly waiting to see RAs before initiating IPv6, but that remains to be tested (which isn't going to happen tonight).

tcpdump is happy with the captured RAs, and I've adhered to spec, but none of the routers I've got available to test with will accept RAs and switch to DHCPv6, they all require explicit DHCPv6 client configuration.

Pushed as a separate commit for easier review (rest of the code remains unchanged).  Happy to squash these two commits if that's preferred.

jkroonza:
Thanks for the review and the merge!  Appreciated.

PR #561, opened 2025-07-09
Stricter checks on scripts.

jkroonza:
Perform symbolic link path resolution before performing checks (avoid symlink change attacks).

Ensure the path leading to the executable is root:* owned, and not writable by anyone else.

Ensure the executable itself is:

Owned by root.
Executable for root.
Not writable by group or other.

Doesn't look like we need to check FACLs (the bits for additional users are mixed into group permission bits according to my testing).

### DIFF
--- a/pppd/main.c
+++ b/pppd/main.c
@@ -1914,10 +1914,52 @@ update_script_environment(void)
  * reap_kids) iff the return value is > 0.
  */
 pid_t
-run_program(char *prog, char * const *args, int must_exist, void (*done)(void *), void *arg, int wait)
+run_program(char *_prog, char * const *args, int must_exist, void (*done)(void *), void *arg, int wait)
 {
     int pid, status, ret;
     struct stat sbuf;
+    char prog[PATH_MAX];
+    char *slash;
+
+    /*
+     * Resolve symlinks such that we eliminate a few potential ToCToU
+     * attack avenues (eg, changing symlinks).
+     */
+    if (!realpath(_prog, prog)) {
+	if (must_exist || errno != ENOENT)
+	    warn("Can't execute %s: %m", prog);
+	return 0;
+    }
+
+    /*
+     * full check the entire path, must be root: owned, and NOT be writable
+     * to group/other (which incorporates FACL bits somehow, so we can ignore
+     * explicit FACL checks).
+     */
+    do {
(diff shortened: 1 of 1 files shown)


paulusmack:
> I suspect we can create a stub realpath() using readlink() repeatedly until we have something that's NOT a symbolic link.

I wasn't saying we had to have a feature check, I was just raising the question. It seems like realpath() has been in POSIX etc. since at least 2004, so we're probably OK to assume it's available. It's certainly in glibc and musl.

> fstatat is only used here to allow use of flags=AT_SYMLINK_NOFOLLOW, since we resolve the symlinks first, not having AT_SYMLINK_NOFOLLOW would constitute another ToCToU issue, however, that's probably the best we can do if we don't have fstatat available for use.

Looks like fstatat() dates back to 2008, so it's probably OK to assume it too. Or we could use lstat, which goes back to at least 2004.

> Regarding other checks, sure, I could implement those too. It does actually make sense to ensure that options and secrets files too are similarly restricted in terms of access, so perhaps just split this out into a separate function:
> 
> ```
> int ppp_check_access(const char* path, char** use_path, int executable) { ...}
> ```
> 
> So that's it becomes part of the ppp API too such that plugins can utilize it too?

Sounds great.

> Call with something like:
> 
> ```
> const char * ip_up_path = "/etc/ppp/ip-up";
> char* real_ip_up_path;
> 
> ...
> if (!ppp_check_access(ip_up_path, &real_ip_up_path, 1)) {
>    failure handling here;
> }
> ... success path
> free(real_ip_up_path);
> }
> ```
> 
> Alternatively, the caller needs to supply a storage buffer for real_ip_up_path and provide that?

Either seems fine.

> The required handling is then external to ppp_check_access?

Sorry, what do you mean by "required handling"?

> In failure case we could rig use_path to point to a constant error string such that logging happens by the caller?

Probably cleaner to have an extra parameter to pass back an error string, and set it to NULL if there is no error.

jkroonza:
> @jkroonza: Thanks for your PR :) There are a lot of needed changes to be perfect?

No, but I'm catching up with $$$ work after a short holiday :).

Neustradamus:
@jkroonza: Hope to have a complete solution for your good work ^^

PR #565, opened 2025-08-13
Fix typo in pppoe-discovery.c ( Signed-off-by: Gregor Heuer <gregwarrior@gmx.de> )

Fallguy0815:
Fix typo in usage(void) of pppoe-discovery
"-Q" should be described as Quiet (as in silent) mode, not Quit (as in exit) mode

Signed-off-by: Gregor Heuer <gregwarrior@gmx.de>


paulusmack:
Thanks for the patch, it looks fine. I don't know why the DCO check failed; I think I'll ignore that since there are actually two sign-offs. I'll edit the subject line since we only need the sign-off at the end of the description.

Issue #311, opened 2021-08-20
Add VRF binding support


c-po:
Hi Paul,

I am one of the mainters of VyOS an open source, Linux (Debian) based network operating system. We utilize pppd to provide dial-up (pppoe) connections for the users. We have added VRF support in our upcoming 1.3 release by using ip-up/ip-down scripts to move an interface into the appropriate VRF - that somehow works but could be nicer.

# Why?

The flakyness now comes into play when routes should be installed by pppd, pppd will always install the route into the main routing table - no problem - we can workaround this in an ip-up/down helper, too.

# Feature Request

Could a new configuration item named `vrf` be added into the peer configuration file? If that vrf is defined, and the vrf exists, pppd can then automatically assign the interface into the appropriate VRF without the need for additional external helpers.

Neustradamus:
@paulusmack, @enaess: What do you think?

paulusmack:
Sounds reasonable. Do you have a patch?

c-po:
I do not have a patch but I would provide one. As I am not familiar with the codebase maybe you can point me to the file which needs to be extended.

* I would implement this only for Linux
* I would name the config option for the peer file `vrf` which defaults to `master` which is the Linux default.

Neustradamus:
@c-po: Have you looked?

c-po:
Hi @Neustradamus,

I have workarounded the issue by calling a custom script which does all the VRF handling once the interface was dialed. IMHO not that nice but it works

Neustradamus:
@EasyNetDev: Can you look it?

EasyNetDev:
Hi all,

I can take a look in the near future for a patch. Now I'm busy with creating a script to build custom Debian OS for Banana Pi R4 and not only.

Neustradamus:
@EasyNetDev: It will be too late for Debian 13?

PR #566, opened 2025-08-27
pppd: add VRF binding support

martin-olivier:
Hello,

This PR introduces vrf support to pppd and will resolve https://github.com/ppp-project/ppp/issues/311

Tested on Ubuntu 24.04, using accel-pppd for the ppp server:

Setup:
```sh
sudo ip link add sessions type vrf table 1234
sudo ip link set sessions up
```

Client config:
```
plugin pppoe.so
nic-veth1
ifname PPPOE
vrf sessions
user "testuser"
noauth
defaultroute
usepeerdns
```

Start PPPoE session:
```sh
sudo pppd call test debug dump nodetach
```

Check that the ppp interface belongs to the vrf 'sessions':
```sh
ip link show PPPOE 
```
```
128: PPPOE: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1492 qdisc fq_codel master sessions state UNKNOWN mode DEFAULT group default qlen 3
    link/ppp
```
^^^ The ppp interface belongs to the vrf sessions (master sessions)

Check that the routes installed by pppd are located in the routing table of the vrf 'sessions':
```sh
ip route show table 1234
```
```
default dev PPPOE scope link                         ### default route installed by pppd
10.0.0.1 dev PPPOE proto kernel scope link src 10.0.0.10         ### route installed by the kernel
local 10.0.0.10 dev PPPOE proto kernel scope host src 10.0.0.10       ### route installed by the kernel
```

The same test has been performed successfully without setting the ifname to
be able to test both when the ppp interface is created thought netlink or via ioctl on the ppp driver.

```
plugin pppoe.so
nic-veth1
### ifname PPPOE
vrf sessions
user "testuser"
noauth
defaultroute
usepeerdns
```

### DIFF
--- a/pppd/main.c
+++ b/pppd/main.c
@@ -846,6 +846,10 @@ set_ifunit(int iskey)
 	slprintf(ifname, sizeof(ifname), "%s%d", PPP_DRV_NAME, ifunit);
     info("Using interface %s", ifname);
     ppp_script_setenv("IFNAME", ifname, iskey);
+#ifdef __linux__
+    if (req_vrf[0] != '\0')
+        ppp_script_setenv("VRF", req_vrf, iskey);
+#endif
     slprintf(ifkey, sizeof(ifkey), "%d", ifunit);
     ppp_script_setenv("UNIT", ifkey, iskey);
     if (iskey) {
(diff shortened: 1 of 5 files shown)


paulusmack:
The code looks OK as far as I can see. The commit lacks an update to pppd.8 describing the new option that you have added. The commit message is a bit terse and needs to be expanded to describe more of the motivation, including at least an expansion of the 'VRF' acronym and a brief (1-sentence) explanation of how/why/when it's useful for pppd to be able to bind the PPP interface to a VRF.

martin-olivier:
Hello @paulusmack,

Thank you for your feedback, I just pushed an amendement.

I have added in `pppd.8` the documentation of the new option and the new env variable.
I have also improved the `get_vrf_table_id` function signature.
I have re-writed the commit message.

PR #569, opened 2025-10-16
dhcpv6relay plugin: Add support for NA.

jkroonza:
As it turns out some (I won't mention names) routers ends up using LL addresses when originating IPv6 traffic in the absense of IA_NA issued address (In spite of having IA_PD space available).  This allows the DHCPv6 server to issue same and have routing function.


jkroonza:
For what it's worth:  Already in production.

Neustradamus:
@paulusmack: @jkroonza has done the change quickly! :)
What do you think now?

Issue #541, opened 2025-01-03
2.5.2 - link failure when multilink is disabled


kanavin:
Specifically, this happens:

```
[...]

PR #570, opened 2025-10-22
pppd/Makefile.am: correctly install-but-not-distribute pppdconf.h

kanavin:
BUILT_SOURCE statement had no effect, and is replaced by the correct statement (verified with 'make dist' and 'make install').

This should resolve https://github.com/ppp-project/ppp/issues/541


Neustradamus:
@paulusmack, @jkroonza, @enaess: Can you look this @kanavin PR?

jkroonza (approved):

PR #572, opened 2025-10-22
dhcpv6relay plugin: Make all sockets non-blocking.

jkroonza:
$ strace -p 3701
strace: Process 3701 attached
read(14^C^Cstrace: Process 3701 detached

$ ls -lah /proc/3701/fd/14
lrwx------ 1 root root 64 Oct 22 02:52 /proc/3701/fd/14 -> 'socket:[2478479691]'

$ grep 2478479691 /proc/net/{udp,tcp,raw}*
/proc/net/raw6:  118: 000002FF000000000000000002000100:003A 00000000000000000000000000000000:0000 07 00000000:00000000 00:00000000 00000000     0        0 2478479691 2 00000000c0af204a 0

So it's blocking trying to receive RS.  This seems bogus, as the only raw socket is dhcpv6relay_sock_rsra, and this only ever gets read from dhcpv6relay_router_solicitation which should only ever be called if event-handler determines there's data available.

All sockets are datagram (and by implication already deemed "unreliable") based, so making them all non-blocking should have no adverse effect, however, blocking pppd process most certainly has undesired effects.

### DIFF
--- a/pppd/plugins/dhcpv6relay/dhcpv6relay.c
+++ b/pppd/plugins/dhcpv6relay/dhcpv6relay.c
@@ -416,7 +416,7 @@ void dhcpv6relay_server_event(int fd, void*)
     socklen_t slen = sizeof(sa);
     bool valid_source = true;
     int hlim = 0;
-    ssize_t r = recvfrom(fd, buffer, sizeof(buffer), MSG_DONTWAIT,
+    ssize_t r = recvfrom(fd, buffer, sizeof(buffer), MSG_DONTWAIT | MSG_TRUNC,
 	    (struct sockaddr*)&sa, &slen);
     if (r < 0) {
 	error("DHCPv6 relay: Failed to read from %s socket: %s",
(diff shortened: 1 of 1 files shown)


Neustradamus:

jkroonza:
> With these sockets in non-blocking mode, are we still guaranteed to get complete packets returned by recvfrom()? Or could we get partial packets and need to reassemble them?

These are all either UDP or RAW sockets, so to the best of my understanding yes I believe so.  The read will either succeed, or fail.  In the case of buffer too small, question really is, will we end up in a situation where we receive a partial packet or will we get -1 with some appropriate error.

Based on experience with work in DNS I believe we end up with a truncated packet in the case of read().

In the case of ppp we have two cases:

1.  Router Solicitation messages, in which case we don't care, the content of the RS doesn't change the content of the Router Announcement we respond with.  We're using read() here, and truncated packet is fine.

2.  In the case of DHCP messages it gets a LOT more tricky, based on the description of MSG_TRUNC we will end up with a truncated packet, MSG_TRUNC there would allow us to detect it at least.

PR #573, opened 2025-10-27
Remove MRU limit on PPPoE

### LINKED ISSUE #331: Any technical limitation for MTU > 1500 bytes?
I see that for some reason, `ppp` is hardcoded a limit for MTU/MRU at max 1500 here:
https://github.com/ppp-project/ppp/blob/master/include/net/ppp_defs.h#L55

Even if I specify the MTU/MRU number higher than 1500 then it will just jump back to 1500
I know that a standard PPPoE MTU is supposed to be 1492 to add space for the overheads, I tried to raise those numbers to work with 1600 bytes and it seems to work just fine (my BRAS supported it).

My question is that why can't we make those MTU MRU dynamically fit with whatever the lower layer support and not limit it down to 1500? I feel like there's a reason for that but the .h file didn't tell me what it is... I need a larger MTU to help with the fragmentation issue if the client is using 4in6/DSLite.

mateusz834:
Fixes #331

### DIFF
--- a/pppd/plugins/pppoe/plugin.c
+++ b/pppd/plugins/pppoe/plugin.c
@@ -453,11 +453,6 @@ void pppoe_check_options(void)
     lcp_allowoptions[0].neg_pcompression = 0;
     lcp_wantoptions[0].neg_pcompression = 0;
 
-    if (lcp_allowoptions[0].mru > MAX_PPPOE_MTU)
-	lcp_allowoptions[0].mru = MAX_PPPOE_MTU;
-    if (lcp_wantoptions[0].mru > MAX_PPPOE_MTU)
-	lcp_wantoptions[0].mru = MAX_PPPOE_MTU;
-
     /* Save configuration */
     conn->storedmtu = lcp_allowoptions[0].mru;
     conn->storedmru = lcp_wantoptions[0].mru;


jkroonza:
I'll see if I can get a test environment set up, need to test the pppoe-ia code for pppoe-server (rp-pppoe) as well.  Might just as well do both at the same time.

mateusz834:
So... (i think) that maybe only changing this:

https://github.com/ppp-project/ppp/blob/13291f4c08ebe23afc5427c0e66f8ebb56305fbe/pppd/plugins/pppoe/plugin.c#L456-L459

would solve the issue?

Neustradamus:

jkroonza:
> So... (i think) that maybe only changing (removing) this:
> 
> https://github.com/ppp-project/ppp/blob/13291f4c08ebe23afc5427c0e66f8ebb56305fbe/pppd/plugins/pppoe/plugin.c#L456-L459
> 
> would solve the issue?

I believe you're right, under the *assumption* that we're using kernel-mode helpers (I don't think userspace is supported any more at all, at least, pppoe-server side ditched that code a while back.  I do not believe that any discovery frames at least can possibly exceed even 500 bytes (from what I recall of the protocol, but it's possible that some network uses multi-KB line tags along with pppoe-ia then tecnically it's possible I guess).  In practise though I highly doubt we'll ever exceed even 100 byte frames during discovery (longest line tags I've seen in my environments are probably 20 characters long).

@mateusz834 would you mind updating the PR to removing those lines?

@paulusmack def not ready to merge yet, your insight would be appreciated.

@dfskoll if you don't mind jumping in here as well please to just confirm our reasoning?

Would be absolutely tremendous if we can get to at least 1500 *inner* MTU for most of our clients - inter-AS traffic is (and probably will be for a very long time still) be 1500 byte MTU, which would just mean less crap with routers that even in this day and age don't do proper MSS adjustments for TCP SYN frames, and perhaps slightly less fragmentation for the odd 1500 byte UDP frame.

Neustradamus:
@paulusmack, @enaess, @chipitsine, @jow-, @systemcrash: Have you seen this PPP PR?

It is linked, for example, to:
- PPP file: https://github.com/ppp-project/ppp/blob/master/pppd/plugins/pppoe/discovery.c#L723
- OpenWrt file: https://github.com/openwrt/openwrt/blob/main/package/network/services/ppp/files/ppp.sh#L243
- https://github.com/ppp-project/ppp/issues/331
- https://github.com/ppp-project/ppp/pull/573
- https://github.com/openwrt/openwrt/issues/5265
- https://github.com/openwrt/openwrt/issues/8190
- https://github.com/openwrt/openwrt/issues/14024
- https://github.com/openwrt/openwrt/pull/20606
- https://github.com/opnsense/core/issues/9398

MTU on wiki.wireshark.org:
- https://wiki.wireshark.org/MTU

RFCs MTU list:

RFC 1191 - Path MTU discovery
- https://datatracker.ietf.org/doc/html/rfc1191
```
1500   Ethernet Networks             RFC 894
1500   Point-to-Point (default)      RFC 1134
1492   IEEE 802.3                    RFC 1042
```

RFC 894 - A Standard for the Transmission of IP Datagrams over Ethernet Networks
- https://datatracker.ietf.org/doc/html/rfc894

RFC 1042 - Standard for the transmission of IP datagrams over IEEE 802 networks
https://datatracker.ietf.org/doc/html/rfc1042

RFC 1134 - Point-to-Point Protocol: A proposal for multi-protocol transmission of datagrams over Point-to-Point links
- https://datatracker.ietf.org/doc/html/rfc1134

RFC 1981: Path MTU Discovery for IP version 6
- https://datatracker.ietf.org/doc/html/rfc1981

RFC 2923 - TCP Problems with Path MTU Discovery
- https://datatracker.ietf.org/doc/html/rfc2923

RFC 3988: Maximum Transmission Unit Signalling Extensions for the Label Distribution Protocol
- https://datatracker.ietf.org/doc/html/rfc3988

RFC 4459 - MTU and Fragmentation Issues with In-the-Network Tunneling
- https://datatracker.ietf.org/doc/html/rfc4459

RFC 4638 - Accommodating a Maximum Transit Unit/Maximum Receive Unit (MTU/MRU) Greater Than 1492 in the Point-to-Point Protocol over Ethernet (PPPoE)
- https://datatracker.ietf.org/doc/html/rfc4638

RFC 4821 - Packetization Layer Path MTU Discovery
- https://datatracker.ietf.org/doc/html/rfc4821

RFC 7690 - Close Encounters of the ICMP Type 2 Kind (Near Misses with ICMPv6 Packet Too Big (PTB))
- https://datatracker.ietf.org/doc/html/rfc7690

RFC 8201 - Path MTU Discovery for IP version 6
- https://datatracker.ietf.org/doc/html/rfc8201

RFC 8899 - Packetization Layer Path MTU Discovery for Datagram Transports
- https://datatracker.ietf.org/doc/html/rfc8899

RFC 9268 - IPv6 Minimum Path MTU Hop-by-Hop Option
- https://datatracker.ietf.org/doc/html/rfc9268

mateusz834:
That commit seems to be pre-rfc4638, so i guess the behavior back then was right, the commit basically  allowed `MRU < 1492`.

chipitsine:
> That commit seems to be pre-rfc4638, so i guess the behavior back then was right, the commit basically allowed `MRU < 1492`.

thus we can now drop that behaviour.

(or maybe add config  option to keep pre-rfc4638, but I doubt it worth that)

mateusz834:
> I wonder if there's some implementation that count on MRU < 1492

I doubt, but even if there are such, then this is still is configurable.

jkroonza:
Just to feedback - we've been busy on our end with some pre-bleak-friday preparation and changes which kept us busy.  We're discussing a test environment for this and other patches on pppoe-server too, which closely correlate.

paulusmack:
I never included this kind of thing before because the original PPPoE RFC said that the MRU must be limited to 1492. Now that RFC 4638 gives us a way to lift that limit, we should implement that rather than just blindly removing the limit. In other words, I won't take this but I would take an implementation of RFC 4638.

mateusz834:
Aren't we already supporting rfc 4638, but up to 1500 mru? This only removes that restriction and allows >1500 mru.

mateusz834:
If RFC 4638 does not happen in the discovery, then we still limit the MRU to 1492, see: https://github.com/ppp-project/ppp/blob/1affa97637a405b1557b3e7f1c23239bfe65889d/pppd/plugins/pppoe/discovery.c#L722-L728

mateusz834:
Per my understanding, we already have a RFC 4638 compatible implementation, but limited to 1500 MRU. I think we only need to remove that 1500 MRU limit (this PR).

paulusmack:
> Per my understanding, we already have a RFC 4638 compatible implementation, but limited to 1500 MRU. I think we only need to remove that 1500 MRU limit (this PR).

You're right. Sorry about the delay (distracted with other things).

PR #577, opened 2026-02-13
Strict access checks for scripts, secrets files and call option files

paulusmack:
This is a reworking and extension of @jkroonza's work in PR #561 to pull out the checking code into a separate function and use it for secrets files and call option files as well as scripts. I moved the checking code into a new function called ppp_check_access() and I used lstat() instead of fstatat(..., AT_SYMLINK_NOFOLLOW), updated the error messages, and reworked the main loop, but apart from that, the body of ppp_check_access() is basically @jkroonza's code.

I would appreciate review and testing - in particular, are there things being checked here that shouldn't be, or things that aren't and should be?

### DIFF
--- a/pppd/utils.c
+++ b/pppd/utils.c
@@ -75,6 +75,102 @@ struct buffer_info {
 };
 
 /*
+ * Check that a file is owned by root, not writable by group or other,
+ * and that all the directories in the path leading to it is likewise
+ * owned by root and not writable by group or other.  The real path to
+ * the file is returned in *path_to_use, which should be freed after
+ * use (real meaning not containing any symlinks or ".." components).
+ * If exec is true, check for execute permission, otherwise for read
+ * permission.
+ * Returns 1 if OK; if not, prints an error message and returns 0.
+ */
+int
+ppp_check_access(const char *path, char **path_to_use, int must_exist, int exec)
+{
+    char *rpath, *slash, *part;
+    struct stat sbuf;
+    int perm;
+
+    /*
+     * Resolve symlinks such that we eliminate a few potential ToCToU
+     * attack avenues (eg, changing symlinks).
+     */
+    rpath = realpath(path, NULL);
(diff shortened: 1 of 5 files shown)


Neustradamus:
@jkroonza: Have you seen this PR?

pppd/utils.c
```diff
@@ -74,6 +74,104 @@ struct buffer_info {
...
+    char *rpath, *slash, *part;
+    struct stat sbuf;
+    int perm;
+
+    /*
+     * Resolve symlinks such that we eliminate a few potential ToCToU
+     * attack avenues (eg, changing symlinks).
+     */
+    rpath = realpath(path, NULL);
+    if (rpath = NULL) {
```
jkroonza: Well, this will always be false, rpath == NULL here, or more succinctly, !rpath (which is also my preference to be honest).

jkroonza (changes requested):
One bug, two comments/questions/suggestions for possible improvement.  The loop code duplication issue is probably best served by:

```
while (true) {
  /* tests on part */
  if (!slash)
    break;
  /* adjustments for slash */
}

/* last component explicit tests */
```

Which is a construct I also don't like, but I like the code duplication even less.

jkroonza (commented):
Just to be clear:  In theory this still suffers a TOCTOU issue whereby if somehow the attacker has the ability to change ownership/permisssions between check time and use time ... which would require at least some level of root involvement in which case I believe there's already far larger issues.

jkroonza:
> > One thing I miss from gerrit to github is a diff between diffs ... ie, what changed between versions of this patch series :(
> 
> Git can do it easily, and in gitk it's a couple of mouse clicks. Here's the diff.

Thank you.  Yea, if you've got refs on both it is indeed easy - but github doesn't expose that *as far as I know*.

LGA1150:
#591

PR #583, opened 2026-04-21
pppd: Make set and unset options privileged [security]

paulusmack:
Since the "set" and "unset" options affect the environment passed to scripts that run as root, and environment variable settings could affect or disrupt the execution of those scripts, these options should be privileged.


Neustradamus:
@jkroonza, @enaess: Have you seen this @paulusmack PR?

PR #585, opened 2026-05-09
pppd: Remove dead check for existence of route to remote IP

paulusmack:
The have_route_to() call in auth_ip_addr() is dead code, so remove it. It can never be reached because, to reach it, we would have to have !auth_required && !allow_any_ip && !privileged, but auth_check_options() sets auth_required to 1 in that case.

In other words, the logic established by previous commits is this:

If pppd is run by non-root and the noauth option isn't in effect, or if the auth option is given, the peer has to authenticate, and the IP addresses are controlled by the secrets file, with no reachability check.

If pppd is run by root and authentication isn't required (i.e. the auth option is not given), the peer can use any IP address, with no reachability check.

In neither case is the reachability check done by have_route_to() used.

### DIFF
--- a/pppd/auth.c
+++ b/pppd/auth.c
@@ -2143,7 +2143,7 @@ auth_ip_addr(int unit, u_int32_t addr)
 
     if (auth_required)
 	return 0;		/* no addresses authorized */
-    return allow_any_ip || privileged || !have_route_to(addr);
+    return allow_any_ip || privileged;
 }
 
 static int


Neustradamus:
@jkroonza: Have you seen this @paulusmack PR?

jkroonza:
I have not, but the logic sounds OK.  It does make me think we should require an explicit auth or noauth to be provided though, or auth should be default.

PR #588, opened 2026-05-19
Update README, add warning if CBCP is enabled, remove warnings

paulusmack:
Update README
Add warning if CBCP is enabled at configure time
Remove warnings

### DIFF
--- a/README
+++ b/README
@@ -67,8 +67,61 @@ use any IP address.  (This only applies where the peer is
 authenticating itself to you, of course.)
 
 
-What's new in ppp-2.5.2
+What's new in ppp-2.5.3
 ***********************
+
+* Several security improvements:
+  - Some options are now privileged: 'set', 'unset',
+    'defaultroute', and 'defaultroute6'.  If a non-root user
+    running a setuid-root pppd needs to use these options,
+    the system administrator will have to make a 'call' file
+    in /etc/ppp/peers containing the required option(s) for
+    the user's use.
+
+  - Scripts, privileged options files and secrets files now are
+    subject to a path check, which checks that the file and each
+    directory in the real path to the file are owned by root and
+    not writable by non-root.
+
+  - If pppd is installed setuid-root and run by a non-root user,
+    the peer will be required to authenticate itself; previously
+    this requirement only applied if the system had a default
+    IPv4 route.
+
+* Default route handling has changed; pppd no longer checks for
+  an existing default route before adding its default route.  The
+  defaultroute and defaultroute6 options are now privileged, and
+  if used, the default route will always be added.  The metric of
+  the default route can be controlled with new defaultroute-metric
+  and defaultroute6-metric options, which are privileged.
+  The replacedefaultroute and noreplacedefaultroute options
+  are no longer functional, and just cause an error message to
+  be printed.
+
+* There is now a dhcpv6relay plugin, which provides a DHCPv6
+  relay for the local system inside pppd.
+
+* VRF (Virtual Routing and Forwarding) support has been added
+  to pppd on Linux.  There is now a 'vrf' option which tells
+  pppd to bind the PPP interface to a specific VRF, so that
+  routes are installed in the VRF's routing table rather than
+  the main routing table.
+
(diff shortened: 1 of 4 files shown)


README
```diff
@@ -67,9 +67,62 @@ use any IP address.  (This only applies where the peer is
...
+
+* VRF (Virtual Routing and Forwarding) support has been added
+  to pppd on Linux.  There is now a 'vrf' option which tells
+  pppd to bind the PPP interface to a specific VRF, so that
+  routes are installed in the VRF's routing table rather than
+  the main routing table.
+
+* The pppoe (PPP over ethernet) plugin now supports maximum
+  packet sizes greater than 1492 bytes if configured to do so
+  and the server agrees.
```
jkroonza, line 112: and the *peer* agrees.

Sorry, I primarily use pppd on the service provider side :).

README
```diff
@@ -67,9 +67,62 @@ use any IP address.  (This only applies where the peer is
...
+  defaultroute and defaultroute6 options are now privileged, and
+  if used, the default route will always be added.  The metric of
+  the default route can be controlled with new defaultroute-metric
+  and defaultroute6-metric options, which are privileged.
+  The replacedefaultroute and noreplacedefaultroute options
+  are no longer functional, and just cause an error message to
+  be printed.
+
+* There is now a dhcpv6relay plugin, which provides a DHCPv6
+  relay for the local system inside pppd.
```
jkroonza, line 102: This is for service provider side to generate RA's indicating managed mode and then to forward the incoming DHCPv6 client request to the real DHCPv6 server, as well as manage the routing around the relevant delegations that the DHCPv6 server then provides.  Essentially it just appends a route for the delegations (whether it's IA_NA or IA_PD) to the local routing table and to again remove them upon expiry.  This is probably too much detail for here.

For client side it's always been possible to use for example dhcpcd -6 on the ppp interface to obtain addressing information.  Possibly in combination with something like radvd to forward advertise.

How about:

There is now a dhcpv6relay plugin, which can be used to provide IPv6 RAs to the remote side and relay the subsequent incoming DHCPv6 requests to a DHCPv6 server.  Note:  This is to delegate IPv6 to the remote side, not to configure IPv6 locally - for that you can look at projects like dhcpcd and/or radvd as needed.

configure.ac
```diff
@@ -362,3 +362,6 @@ Features enabled
     EAP-TLS..............: ${enable_eaptls:-yes}
     systemd notifications: ${enable_systemd:-no}
 "
+
+AM_COND_IF([PPP_WITH_CBCP],
+    AC_MSG_WARN([CBCP support is likely to be removed in future]))
```
jkroonza, line 367: I'd make this more aggressive and point to the specific issue

[CPCP is going to be removed in version (one of) 2.5.X (or) 2.6.0 unless motivation to keep it is given to the issue at https://....]

paulusmack:
I'm about to merge this, update configure.ac to change the version number to 2.5.3, and do the 2.5.3 release. If anyone knows of anything else that really should be in 2.5.3, speak now...

Neustradamus:
@jkroonza: Can you look this PR and the @paulusmack comment?

jkroonza:
I need to start working on an alternative for radius for another project too, but that's at least two months out.  Other than the potential concerns raised on the README, I have no objections to a new version.  Once you've released I can see if Gentoo be the first to package :).

jkroonza (commented):
Use it, don't use it :).  Other than the dhcpv6relay plugin rewording I'm mostly neutral.  The issue I have there is makes it sound like you need it on the client side when it's really aimed at the SP side.

jkroonza:
@paulusmack I note I never actually submitted my comments ... it's been stuck in the half-review.  So I missed it, not sure if they're al that significant, I suspect for >99% of folks it's "good enough" as it stands.

Neustradamus:
@paulusmack: Have you seen the @jkroonza comment done after merging?

paulusmack:
> @paulusmack: Have you seen the @jkroonza comment done after merging?

PR #590 addresses his comments, hopefully.

PR #590, opened 2026-05-23
README: Improve wording of new feature descriptions

paulusmack:


jkroonza (approved):
Thanks.  I'd add the URL for CBCP to the ./configure message as well, but I don't think that it's absolutely critical.

Issue #591, opened 2026-05-27
Regression: ppp_check_access() may break existing configurations


LGA1150:
https://github.com/openwrt/openwrt/pull/23540#issuecomment-4552677593

ppp-project/ppp@60392cf89b2fd5dc386406f36770e156ef60e07b
The check requires all paths from the root dir down to not have group write or others write permissions, which is too strict and may break existing configurations.

There are better ways to solve the TOCTOU bug. One may be using `openat2() + fexecve()` on fd.

jkroonza:
@LGA1150 

The moment there is write anywhere on the path it becomes potentially possible for normal users to modify the target file.  At least, that was our reasoning.  openat2 isn't exposed via libc.  It's also Linux specific, not POSIX, so can't rely on that.

others write I agree with, but if group is root then group-write IMHO is acceptable.

We should probably also check facls in the check process too:

```
jkroon@plastiekpoot /tmp $ mkdir foo
jkroon@plastiekpoot /tmp $ setfacl -m u:jkroon2:rwx foo
jkroon@plastiekpoot /tmp $ ls -la foo
total 2496
drwxrwxr-x+ 2 jkroon jkroon    4096 May 27 11:23 .
drwxrwxrwt  9 root   root   2547712 May 27 11:23 ..
jkroon@plastiekpoot /tmp $ getfacl foo
# file: foo
# owner: jkroon
# group: jkroon
user::rwx
user:jkroon2:rwx
group::r-x
mask::rwx
other::r-x
```

Or maybe that's why we did the w filter on group?

@paulusmack I wasn't aware of fexecve - that certainly looks to be an interesting option.  Combined with fstat() to get permissions on the file itself.  So open(), then fstat() on the file, check permissions (and ACLs) on the file itself.  Then fexecve() or read()/mmap() as the case may be after that?  Then I *suspect* we don't even need to check the rest of the path.  There is only so much we can do to protect here, and for files containing secrets we can also ban read for anybody other than root ... (just to force the admin to be a little bit more thorough).

Neustradamus:
@LGA1150: Thanks for your report and the quick @jkroonza answer!

paulusmack:
> [openwrt/openwrt#23540 (comment)](https://github.com/openwrt/openwrt/pull/23540#issuecomment-4552677593)

Interesting, thanks. I see that /lib/netifd has mode 0755 on the OpenWRT instance I have here (OpenWRT 23.05.5), do we know why it got changed to 0775?

[...]

PR #592, opened 2026-05-27
pppd: relax and simplify permission check

LGA1150:
The additional check added in v2.5.3 is intended to mitigate symlink attacks, but it checks the permissions of every component of a path and is too strict, breaking existing configurations. Relax and simplify the check by only checking the opened file descriptor of a path, so the path is resolved only once.

Related issue: #591

### DIFF
--- a/pppd/auth.c
+++ b/pppd/auth.c
@@ -307,7 +307,7 @@ static int  setupapfile (char **);
 static int  privgroup (char **);
 static int  set_noauth_addr (char **);
 static int  set_permitted_number (char **);
-static void check_access (FILE *, char *);
+static void check_access (int, const char *);
 static int  wordlist_count (struct wordlist *);
 static void check_maxoctets (void *);
 
@@ -532,7 +532,7 @@ setupapfile(char **argv)
         free(fname);
 	return 0;
     }
-    check_access(ufile, fname);
+    check_access(fileno(ufile), fname);
     uafname = fname;
 
     /* get username */
@@ -1543,8 +1543,7 @@ check_passwd(int unit,
      * Open the file of pap secrets and scan for a suitable secret
      * for authenticating this user.
      */
-    if (!ppp_check_access(path_upapfile, &filename, 0, 0))
-	    return UPAP_AUTHNAK;
+    filename = path_upapfile;
     addrs = opts = NULL;
     ret = UPAP_AUTHNAK;
     f = fopen(filename, "r");
@@ -1552,7 +1551,13 @@ check_passwd(int unit,
 	error("Can't open PAP password file %s: %m", filename);
 
     } else {
-	check_access(f, filename);
+	int fd = fileno(f);
+
+	if (!ppp_check_access(fd, filename, 0)) {
+	    fclose(f);
+	    return UPAP_AUTHNAK;
+	}
+	check_access(fd, filename);
 	if (scan_authfile(f, user, our_name, secret, &addrs, &opts, filename, 0) < 0) {
 	    warn("no PAP secret found for %s", user);
 	} else {
@@ -1586,7 +1591,6 @@ check_passwd(int unit,
 	}
 	fclose(f);
     }
-    free(filename);
 
     if (ret == UPAP_AUTHNAK) {
         if (**msg == 0)
@@ -1646,21 +1650,24 @@ null_login(int unit)
      * Open the file of pap secrets and scan for a suitable secret.
      */
     if (ret <= 0) {
-	if (!ppp_check_access(path_upapfile, &filename, 0, 0))
-	    return 0;
+	int fd;
+
+	filename = path_upapfile;
 	addrs = NULL;
 	f = fopen(filename, "r");
-	if (f == NULL) {
-	    free(filename);
+	if (f == NULL)
+	    return 0;
+	fd = fileno(f);
+	if (!ppp_check_access(fd, filename, 0)) {
+	    fclose(f);
 	    return 0;
 	}
-	check_access(f, filename);
+	check_access(fd, filename);
 
 	i = scan_authfile(f, "", our_name, secret, &addrs, &opts, filename, 0);
 	ret = i >= 0 && secret[0] == 0;
 	BZERO(secret, sizeof(secret));
 	fclose(f);
-	free(filename);
     }
 
     if (ret)
@@ -1701,7 +1708,7 @@ get_pap_passwd(char *passwd)
     f = fopen(filename, "r");
     if (f == NULL)
 	return 0;
-    check_access(f, filename);
+    check_access(fileno(f), filename);
     ret = scan_authfile(f, user,
 			(remote_name[0]? remote_name: NULL),
 			secret, NULL, NULL, filename, 0);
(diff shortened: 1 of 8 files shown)


jkroonza (changes requested):
Using FILE* where not needed.  I don't like the approach, especially for the fexec side, will see if I can find cycles on this.

LGA1150:
I've refactored `ppp_check_access` and `check_access`. They now check the file descriptor instead of file name.

jkroonza:
@LGA1150 I'm going to unsubscribe myself for the moment, please do @jkroonza me once you believe you're done.

Thank you for moving this in a better direction.  I see you're still pushing quite a lot so please @jkroonza when you're looking for specific input or believe you're done.  I don't have commit rights but I'll happily help review.

You'll note the original patch introduced a few extra char* variables to hold the absolute filename and to eliminate symlink attacks (which with your patch is no longer needed at all).

In order to simplify things long-term (not for the patch itself), you may want to eliminate those again as well.  Eg, filename = path_upapfile.  There should no longer be a reason to copy the pointer at all, just use path_upapfile as is, for example.

LGA1150:
@jkroonza The `filename` local variable was _not_ introduced in said commit, so the simplification is unrelated to this pull request.

pppd/utils.c
```diff
@@ -75,98 +75,50 @@ struct buffer_info {
...
 	      exec? "execut": "read");
 	goto err;
     }
 
-    *path_to_use = rpath;
     return 1;
 
  err:
-    free(rpath);
     return 0;
```
jkroonza, line 122: Side note: Since we no longer cleanup anything this goto can go away, but I'm not hard objecting to the idiom used here.

pppd/auth.c
```diff
@@ -307,7 +307,7 @@ static int  setupapfile (char **);
 static int  privgroup (char **);
 static int  set_noauth_addr (char **);
 static int  set_permitted_number (char **);
-static void check_access (FILE *, char *);
+static void check_access (int, const char *);
```
jkroonza, line 310: I know this was still in place after previous patch set, but since this is now already checked by ppp_check_access I think this function can go away entirely.  I think you modified this now based on my previous review noting that ppp_check_access was then taking a FILE*.

I think in order to not just fix the patch but actually make progress too regarding simplification, kill this function entirely too please unless I'm being stupid and it does actually check something which I'm missing, but as far as I can tell it doesn't actually enforce anything other than *warning* about the file being group/other writeable - something which is now enforced already by ppp_check_access.

pppd/auth.c
```diff
@@ -1543,16 +1543,21 @@ check_passwd(int unit,
      * Open the file of pap secrets and scan for a suitable secret
      * for authenticating this user.
      */
-    if (!ppp_check_access(path_upapfile, &filename, 0, 0))
-	    return UPAP_AUTHNAK;
+    filename = path_upapfile;
```
jkroonza, line 1546: I get that this is a straight up revert, so if you don't mind, please fix in this commit - so either you can clean up now or I'll post a follow-up clean up patch where I can find this idiom.

Cases like this is not ideal either so a follow up commit trying to trace and fix all of them is possibly better:

```
2546 have_eaptls_secret_client()

2550 char *filename;

2567 filename = PPP_PATH_EAPTLSCLIFILE;
```

Plainly we're assigning const char* there to char* which can lead to hard to troubleshoot issues.

pppd/auth.c
```diff
@@ -1543,16 +1543,21 @@ check_passwd(int unit,
...
     ret = UPAP_AUTHNAK;
     f = fopen(filename, "r");
     if (f == NULL) {
 	error("Can't open PAP password file %s: %m", filename);
 
     } else {
-	check_access(f, filename);
+	int fd = fileno(f);
+
+	if (!ppp_check_access(fd, filename, 0)) {
```
jkroonza, line 1556: Once check_access() goes away I think fileno() inline here is fine avoiding the extra variable.

jkroonza (approved):
None of my comments are blockers @paulusmack I think this is good but can be improved - I'm happy to perform the cleanups in a follow-up commit which I think would be in order anyway to find and sort out a few cases which really aught to be const char* where char* is currently used.

pppd/auth.c
LGA1150: A subsequent clean-up patch could constify all path variables as long as parameters.

pppd/auth.c
paulusmack: Actually it warns about the file being group/other _readable_ so that is different to `ppp_check_access`. But this function could easily go away, making a nice simplification, with an extra parameter to `ppp_check_access` telling it to check group/other read permissions.

pppd/main.c
```diff
@@ -1923,22 +1923,31 @@ update_script_environment(void)
...
-    int pid, status, ret;
-    char *rpath;
+    int fd, pid, status, ret;
 
     /*
      * First check if the file exists and is executable by root,
      * and couldn't have been modified by a non-root process.
      */
-    if (!ppp_check_access(prog, &rpath, must_exist, 1))
+    fd = open(prog, O_RDONLY);
```
paulusmack, line 1937: O_RDONLY is not really correct here, since the file could have execute permission but not read permission. But since we have euid = 0, we can probably open it anyway. May be worth a comment.

pppd/main.c
```diff
@@ -1978,14 +1986,24 @@ run_program(char *prog, char * const *args, int must_exist, void (*done)(void *)
...
-	syslog(LOG_ERR, "Can't execute %s: %m", rpath);
-	closelog();
+#ifdef HAVE_FEXECVE
+    fexecve(fd, args, script_env);
+#else
+    {
+	char fdpath[32];
+
+	snprintf(fdpath, sizeof(fdpath), "/dev/fd/%d", fd);
+	execve(fdpath, args, script_env);
```
paulusmack, line 2002: I'm not sure this trick would work on Solaris. But fortunately Solaris has fexecve().

pppd/main.c
LGA1150: According to `fexecve` manpage:

The file descriptor fd must be opened read-only (`O_RDONLY`) _or_ with the `O_PATH` flag and the caller must have permission to execute the file that it refers to.

However `O_PATH` isn't portable so we have to use `O_RDONLY` here.

LGA1150: I just read `fexecve` manpage of Solaris, and it mentioned the `O_EXEC` flag:

The file descriptor passed to the `fexecve()` function need not have been opened with the `O_EXEC` flag. However, if the file to be executed denies read and write permission for the process preparing to perform the exec, the only way to provide the file descriptor fd to `fexecve()` is to specify the `O_EXEC` flag when opening fd.

So maybe we need an `#ifdef` here.

paulusmack:
As discussed in issue #591, the premise of the commit message is wrong, in that there is no TOCTOU issue. The original series was not about fixing any TOCTOU bug, it was about trying to detect certain potentially dangerous misconfigurations of the filesystem. So the commit message at least would need to be reworded.

In fact this patch is about weakening the checks introduced in the earlier series so that only the file itself is checked, not the path leading to it, because of concerns about breaking existing setups. If that is what we decide to do, then the code changes seem mostly OK, except that it doesn't update the man page (pppd.8). It could be done a bit more cleanly in that we could avoid having the various functions in auth.c that read secrets files each calling both `check_access()` and `ppp_check_access()` --- either the call to `ppp_check_access()` could be put inside `check_access()`, or we could eliminate `check_access()` entirely and have `ppp_check_access()` optionally check group/other read permissions.

pppd/main.c
LGA1150: I think this will do:

```suggestion
#define _GNU_SOURCE 1 // may be set in configure.ac
#if defined(O_PATH)
    int flags = O_PATH;
#elif defined(O_EXEC)
    int flags = O_EXEC;
#else
    int flags = O_RDONLY; // require read permission on Linux version < 2.6.39
#endif
    fd = open(prog, flags);
```

PR #596, opened 2026-06-13
add ppp testsuite

tridge:
modeled on the rsync 3.5 test framework


Neustradamus:
@paulusmack, @jkroonza: What do you think about this PR?

jkroonza:
I like the direction, would need to dig into details to say yes/no, but I'm definitely pro any/all improvement in testing.

paulusmack:
Thanks Tridge!

PR #597, opened 2026-06-16
fix: free ets->data on error paths in eaptls_receive and eaptls_free_…

xuammu:
## Summary

  `eaptls_receive()` allocates `ets->data` to reassemble fragmented EAP-TLS messages, but three early-return error paths fail to free it.`eaptls_free_session()` also leaks `ets->data` when cleaning up thesession.
  Each triggered error leaks up to 64KB. An attacker can repeatedly send malformed EAP-TLS packets (LI=1, MF=0, tlslen=65536, tiny payload) to exhaust memory and cause denial of service.

  ## Affected code
  `pppd/eap-tls.c` — `eaptls_receive()` and `eaptls_free_session()`

### DIFF
--- a/pppd/eap-tls.c
+++ b/pppd/eap-tls.c
@@ -802,6 +802,7 @@ void eaptls_free_session(struct eaptls_session *ets)
     if (ets->info)
         tls_free_verify_info(&ets->info);
 
+    free(ets->data);
     free(ets);
 }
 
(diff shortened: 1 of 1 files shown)


paulusmack:
Thanks for this. @jjkeijser, @enaess any comments / review?

Fortunately with this kind of thing the attacker can really only deny service to themselves, since each connection has a separate pppd process.

jjkeijser:
The patch looks good to me. Ack

Neustradamus:
@jkroonza: Have you seen this PR?

jkroonza (approved):
LGTM.  Eyeball only.

@paulusmack yea, if they rack up enough concurrency and is quick enough about this to prevent ppp timeout they can potentially consume a reasonable amount of system RAM, but I agree, this is generally not a major issue in all likilihood (I wouldn't roll a new version just for this kind of not important).

PR #599, opened 2026-06-24
pppd: suppress unconditional "Plugin ... loaded" message before daemonization

jhoustek:
### Problem

`pppd` currently emits:

```
Plugin <plugin>.so loaded.
```

via `info()` immediately after successfully loading a plugin.

This happens very early during startup, before the daemon has fully detached. As a result, the message is not only sent to syslog but may also appear on the caller's terminal/stdout/stderr path, depending on how `pppd` was started.

A practical example is the common `pon` workflow:

```
pon myprovider
```

Even when running in a non-debug, non-interactive setup, the user sees:

```
Plugin sstp-pppd-plugin.so loaded.
```

printed during startup.

### Why this is undesirable

Successful plugin loading is a normal internal initialization step, not an operational event that typically requires user attention.

The message is especially surprising because:

* it is emitted even when debug logging is disabled;
* it appears before daemonization is complete;
* it can leak into the output of higher-level orchestration tools (`pon`, network managers, init systems, ifupdown integrations, etc.);
* it creates noise for scripts that expect startup to be silent unless an error occurs.

In my case, an SSTP connection managed through `pon`/`ifupdown-ng` produces a visible line on startup even though the connection succeeds and debug logging is not enabled.

### Suggested change

Move the message from:

```c
info("Plugin %s loaded.", filename);
```

to:

```c
dbglog("Plugin %s loaded.", filename);
```

This preserves the information for troubleshooting while avoiding unsolicited output during normal operation.

### Rationale

Plugin load success is primarily useful for debugging startup issues. In normal production operation, administrators are generally interested in failures to load a plugin, not successful loads of expected plugins.

Making the message debug-only would align it with the principle that non-debug startup should remain quiet unless there is an actual warning or error.

### DIFF
--- a/pppd/options.c
+++ b/pppd/options.c
@@ -1839,7 +1839,7 @@ loadplugin(char **argv)
 		     arg, vers, VERSION);
 	goto errclose;
     }
-    info("Plugin %s loaded.", arg);
+    dbglog("Plugin %s loaded.", arg);
     (*init)();
     if (path != arg)
 	free(path);


Neustradamus:
@paulusmack, @jkroonza: What do you think?

jkroonza:
Don't particularly mind, but given it's one line of log per plugin per connection, I'd personally leave things as is, extremely helpful for troubleshooting problems at higher level without needing to resort to the debug option.

jhoustek:
After looking into this more closely, I think the root cause is the default early logging behavior rather than this message itself.

`pppd` logs to stdout until `detach()`, and plugin options are processed before that point, so the successful plugin load message ends up on the terminal. This can be suppressed with `nolog`, but since both `nolog` and `plugin` are handled during option parsing, their ordering matters.

I realize changing the default logging behavior is probably too invasive for a long-standing project. Given the current behavior, though, I think only messages that are actually important should be emitted before `detach()`. A failed plugin load `warn()` certainly is; a successful plugin load feels more like debug information.

So I still think changing this particular message from `info()` to `dbglog()` is a reasonable, narrowly scoped improvement. If that change is not desirable, it might at least be worth documenting that logging-related options such as nolog need to appear before options that may emit log messages during option parsing (e.g. plugin), otherwise they won't affect those early messages.

PR #601, opened 2026-07-14
Remove atm_tcpip_port_mapping

AZero13:
It is completely unused and buggy too.


paulusmack:
Actually, it looks like that function is entirely unused. The better fix is probably just to delete it.

AZero13:
> Actually, it looks like that function is entirely unused. The better fix is probably just to delete it.

Ok

AZero13:
> Actually, it looks like that function is entirely unused. The better fix is probably just to delete it.

Done!

AZero13:
Why is it saying no DCO? I signed the commit off in the patch itself?

PR #602, opened 2026-07-15
Fix unhandled unknown tls-verify-method and uninitialized buffers in TLS verification

AZero13:
If the user provides an unknown tls-verify-method option, the certificate verification logic would previously fall through and leave ptr2 uninitialized (NULL), leading to a segmentation fault when attempting to call 'strcmp'.

### DIFF
--- a/pppd/tls.c
+++ b/pppd/tls.c
@@ -94,8 +94,8 @@ static inline int SSL_CTX_set_max_proto_version(SSL_CTX *ctx, long tls_ver_max)
  */
 static int tls_verify_callback(int ok, X509_STORE_CTX *ctx)
 {
-    char subject[256];
-    char cn_str[256];
+    char subject[256] = {0};
+    char cn_str[256] = {0};
     X509 *peer_cert;
     int err, depth;
     SSL *ssl;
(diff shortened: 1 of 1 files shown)


jkroonza (approved):
Looks like a sensible improvement.

Overall it looks like we can possibly optimise a bit more, specifically, if on entry ok = 0 we will always return 0.  But we still jump through various pieces of code (presumably to log errors) to achieve this.  If on entry ok=0 shouldn't we just log the verification error and outright return 0?  Separate PR I reckon, to relook at the logic of this function.  There are other aspects of this which confuses me too (eg, the error depth, which is only relevant if ok==0 since otherwise there wasn't an error to begin with.  I highly recommend someone with more intimate knowledge of OpenSSL compared to me please re-look at this function in particular.

Directly related, the possible values for tls_verify_method should be one of (from pppd-private.h):

```
229 #define TLS_VERIFY_NONE     "none"
230 #define TLS_VERIFY_NAME     "name"
231 #define TLS_VERIFY_SUBJECT  "subject"
232 #define TLS_VERIFY_SUFFIX   "suffix"
```

The behaviour to me seems (until now) undefined if tls-verify-method gets set to anything else.  So in full support of this change.

Thank you.

PR #604, opened 2026-07-22
pppd/EAP: Don't send EAP Nak packets as an authentication server

paulusmack:
In the EAP negotiation, when we're authenticating the peer, we should never send an EAP Request-Nak packet; in fact such a packet is not defined in the EAP RFC.  We were doing that in two cases; (a) when the peer sends an EAP Response-Nak to suggest that we ask it for MS-CHAPv2, but we don't have MS-CHAPv2 support compiled in, or (b) if we receive an EAP-MSCHAPv2 response that we don't recognize.  Instead of sending a Nak, in the first case we restart the authentication process from the beginning, and in the second case we fail the authentication.

### DIFF
--- a/pppd/eap.c
+++ b/pppd/eap.c
@@ -2501,7 +2501,8 @@ eap_response(eap_state *esp, u_char *inp, int id, int len)
 			/* If MSCHAPv2 digest was not found, NAK the packet */
 			if (!esp->es_server.digest) {
 				error("EAP MSCHAPv2 not supported");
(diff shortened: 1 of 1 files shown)


Neustradamus:
@jkroonza, @enaess, @jjkeijser, @Evengard, @h-milz: Have you seen this @paulusmack PR?

jkroonza:
I don't have enough knowledge to confirm - but looks like @paulusmack has done his research.  Code itself looks fine based on description, so IMHO let it fly.

PR #605, opened 2026-07-23
pppd: Remove EAP-SRP code

paulusmack:
The EAP-SRP support isn't enabled by default, and doesn't seem to see much if any use.  Remove it so we don't have to worry about potential bugs or vulnerabilities in it.


paulusmack:
@carlsonj have I missed anything here?

Neustradamus:
@jkroonza, @enaess, @jjkeijser: What do you think?

jkroonza:
Please refresh my memory, SRP is the one where you need the plaintext password on the client, some intermediate hash on the server and can then use some form of challenge-response (similar to (MS)CHAP) to verify that both sides have the same credentials?

I don't think I've ever seen SRP in use, although I found the principles very interesting indeed.

I do think this can be laid to rest.  If someone complains we can always revert and then deal with the implied vulnerabilities & bugs in it.  I'm assuming @paulusmack got issues reported which is I'm guessing why this PR was generated.

I'm not aware of any clients supporting this.  Lots of EAP-PEAP (Usually combined with MSCHAPv2) that I'm seeing in the field, also EAP-TTLS and EAP-TLS.

I also doubt SRP will stand up to quantum compute due to the reliance on primes.  Whereas TLS based protocols that's moving away from RSA based certificates are more likely to.  But I'm no cryptographer and honestly haven't dug into these kinds of details just yet.  So please take this statement with a bag of salt.

jjkeijser:
I've never been able to compile pppd with EAP_SRP enabled - the SRP code itself no longer compiles or runs on modern Linux distros.  So I say , let's remove it (unless someone does know how to bulid pppd with EAP_SRP enabled)

paulusmack:
> I'm not aware of any clients supporting this. Lots of EAP-PEAP (Usually combined with MSCHAPv2) that I'm seeing in the field, also EAP-TTLS and EAP-TLS.

What's the context for "lots of EAP-PEAP"? Presumably those are Windows-using customers of yours, doing tunneled PPP (i.e. pptp or similar) rather than using PEAP to authenticate to your servers?

The reason I ask is that I would be quite happy to rip out the PEAP code we have in pppd because it's incomplete and buggy and I have no way to test it.

Neustradamus:
It is not possible to fix PEAP?

jkroonza:
Come to think of it, the PEAP I'm seeing is all WiFi related, so no ppp involvement.  Will do no harm to *me* to remove it.  Can't speak for others.

paulusmack:
> It is not possible to fix PEAP?

It's possible, but a lot of work.  However, the more difficult question is how to test it.

Neustradamus:
@enaess, @jjkeijser: Have you possibility to help @paulusmack to improve PEAP and to test it?

carlsonj:
On 7/22/26 10:26 PM, Paul Mackerras wrote:
> *paulusmack* left a comment (ppp-project/ppp#605) <https://github.com/ 
> ppp-project/ppp/pull/605#issuecomment-5053636804>
> 
> @carlsonj <https://github.com/carlsonj> have I missed anything here?

No, that looks pretty comprehensive to me.

Sorry for the delay; I was out at Oshkosh for Airventure.

-- 
James Carlson     42.703N 71.076W FN42lq08    ***@***.***>

carlsonj:
On 7/24/26 4:04 AM, Jaco Kroon wrote:
> *jkroonza* left a comment (ppp-project/ppp#605) <https://github.com/ppp- 
> project/ppp/pull/605#issuecomment-5067573289>
> 
> Please refresh my memory, SRP is the one where you need the plaintext 
> password on the client, some intermediate hash on the server and can 
> then use some form of challenge-response (similar to (MS)CHAP) to verify 
> that both sides have the same credentials?

Yes. It has a number of improvements over the usual CHAP mechanism. 
MS-CHAP is a comical protocol where the actual password in use is an MD4 
hash, and both sides must store (or obtain) that fixed hash.  It's no 
more secure than regular CHAP that requires both sides to have access to 
the cleartext password.  In SRP, if the authenticator database is 
compromised, the attacker still can't get data that would allow 
successful authentication except by either a dictionary attack or 
compromise of a secure hash algorithm.  It's asymmetric, so only the 
authenticatee needs the password (and unlike MS-CHAP, only that password 
will do).  And it performs very well.

But it definitely did not catch on among PPP users.

> I don't think I've ever seen SRP in use, although I found the principles 
> very interesting indeed.

I've used it, but not for a couple of decades. On the other hand, it's 
apparently in use for other protocols besides PPP-EAP.

> I also doubt SRP will stand up to quantum compute due to the reliance on 
> primes. Whereas TLS based protocols that's moving away from RSA based 
> certificates are more likely to. But I'm no cryptographer and honestly 
> haven't dug into these kinds of details just yet. So please take this 
> statement with a bag of salt.
If that ever becomes reality, we're in way deeper than PPP link 
authentication.

-- 
James Carlson     42.703N 71.076W FN42lq08    ***@***.***>

jjkeijser:
> > It is not possible to fix PEAP?
> 
> It's possible, but a lot of work. However, the more difficult question is how to test it.

Testing PEAP would require settting up a Freeradius server and confiugre it to provide PEAP. Also a lot of work, and the real question is:  does anybody out there actually use pppd+PEAP to begin with?  Which cornercase are we trying to cover here?

enaess:
For what it is worth, the simplest way to test the PEAP stuff is if you have a working SSTP/L2TP/PPTP setup on a Windows 2022 Server. I'd say, if you are able to configure it from a user-perspective. It's not only a WiFi protocol, it's used with the Microsoft RAS service with their backend radius server. I'd say, it's work to get setup and verified, but there are other protocols in pppd I would rather deprecate and rip out than the PEAP implementation.

Neustradamus:
@jjkeijser, @enaess: Have you possibility to create a test VM for @paulusmack?
It will be nice to test different features and maybe to fix problems...
Of course, access sent privately to @paulusmack.

jjkeijser:
Nope, no resources, no time.

Like enaess , I am fully in favor of ripping it out

jkroonza:
Also, I think testing should (whilst it can rely on external tools, like the freeradius binary) be self-contained.

If Windows VPN actively uses PEAP protocol for VPN authentication that would be a significant motivation to restore a fixed version of the code, but I'm with @paulusmack - if we can't test it, let's rather axe it - however, I'm not against restoring and fixing if there is a good reason to do so.  I don't see one.

enaess:
I have no bandwidth to help here, test or provide VMs - Sorry ... 

As for the title of this ticket, remove EAP-SRP code, I am all in favor of it. 

As for the EAP-PEAP code, I would protest removing it. It's still a very good / common use-case. At best, you can leave in the code as an option to configure it (default turn it off and call it experimental). If any of the distro-maintainers wanting to include it, they can.

I do have other work pending the SSTP project, but again .. I haven't had the time lately to do anything there, but eventually I hope to be able to set this up and test it again.

paulusmack:
> As for the EAP-PEAP code, I would protest removing it. It's still a very good / common use-case. At best, you can leave in the code as an option to configure it (default turn it off and call it experimental). If any of the distro-maintainers wanting to include it, they can.

My difficulty is that I have received a report of a vulnerability in the PEAP code. (I can send you a copy privately if you're interested.) It looks to be not particularly straightforward to fix, and as mentioned above, I have no way to do any kind of test of PEAP (since we only implement the client side, I can't even test it against itself). That all inclines me to removing it.

Of course, if the code were removed, it would still be available via git, and could be brought back easily if it could be fixed and tested.

PR #606, opened 2026-07-30
Radius plugin hardening

paulusmack:
This is a collection of fixes for the radius plugin code, most of which are about checking length values correctly to avoid buffer overflows. Some of these buffer overflows are user-triggerable, that is, caused by command-line options given to pppd, and some of them could be triggered by the RADIUS server sending a badly-formatted RADIUS reply. The first set are handled in part by making all of the radius plugin command-line options privileged, and also by adding suitable length checks. For the second set, suitable length checks are added. But note that the RADIUS server is trusted and the network between pppd and the RADIUS server is assumed to be secure.

### DIFF
--- a/pppd/plugins/radius/avpair.c
+++ b/pppd/plugins/radius/avpair.c
@@ -471,11 +471,13 @@ void rc_avpair_free (VALUE_PAIR *pair)
  * Purpose: Copy a data field from the buffer.  Advance the buffer
  *          past the data field.
  *
+ * 'string' is assumed to have space for AUTH_ID_LEN characters.
  */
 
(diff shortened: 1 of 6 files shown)


jkroonza (approved):
looks good, I'd like to test in a dev environment prior to final yes, but eyeball looks good.

paulusmack:
> looks good, I'd like to test in a dev environment prior to final yes, but eyeball looks good.

Please do test and let me know how it goes.

Neustradamus:
@enaess, @jjkeijser, @EasyNetDev: Have you seen this @paulusmack PR?
What do you think?

PR #608, opened 2026-08-06
pppd: allow running without root when CAP_NET_ADMIN is present

JuliusBairaktaris:
pppd refuses to run when `geteuid() != 0`, from the setuid-root era: the message even says "since it is not setuid-root". On Linux the privilege pppd actually needs is a capability, not a uid — the kernel gates `ppp_open()` on `ns_capable(CAP_NET_ADMIN)` (see `drivers/net/ppp/ppp_generic.c`, whose own comment notes this "could (should?) be enforced by the permissions on /dev/ppp") and the interface ioctls require the same. A sandbox that grants that capability (procd/ujail on OpenWrt, systemd) can run pppd fully unprivileged, but the gate stops it before it can try.

Three changes, because clearing the gate alone does not get a non-root pppd to a working link:

**1. The startup gate.** Replace it with `ppp_privileged()`: euid 0, or CAP_NET_ADMIN in the effective set, checked via `capget` so the binary needs no setuid bit and no libcap dependency. Solaris keeps the euid check.

**2. The privileged option flag.** `parse_args()` copies `privileged` into `privileged_option`, and `process_option()` refuses any `OPT_PRIV` option unless it is set. `privileged` was `getuid() == 0`, so an unprivileged pppd that passed the new gate still died on its first privileged option — `ifname` for any caller that names the interface, and `plugin` for pppoe, which is how OpenWrt loads `pppoe.so`. So the flag has to follow the capability too.

It must not follow whatever the *exec* handed over, though. `ppp_privileged()` returns true whenever `geteuid() == 0`, so wiring `privileged` straight to it would set the flag for an unprivileged user invoking a **setuid-root** pppd — exactly the install mode the old message names — and equally for a `setcap cap_net_admin=ep` pppd, where the kernel grants the capability to whoever runs the binary. `plugin` is `OPT_PRIV`, so either would be `dlopen` of an attacker-chosen object with those privileges, and the `OPT_PRIV` file options would additionally skip the `seteuid(getuid())` that `option_open()` does for unprivileged callers. The guard is therefore:

```c
privileged = uid == 0 || (uid == geteuid() && !ppp_secure_exec() && ppp_privileged());
```

`ppp_secure_exec()` is `getauxval(AT_SECURE)` on Linux and 0 on Solaris, where the `uid == geteuid()` term still covers setuid. AT_SECURE is the kernel's own verdict on whether this exec raised privileges: `cap_bprm_creds_from_file()` sets it for a setuid or setgid bit and for file capabilities, and leaves it clear when a non-root process is exec'd with ambient capabilities and no file caps, since `pP' = pA'` and nothing grew. So a sandboxed pppd gets the flag; a setuid-root or setcap pppd invoked by a normal user does not, exactly as before this patch.

**3. `run_program()`.** It re-roots the ip-up/ip-down child with `setuid(0)` before exec. Without `CAP_SETUID` that fails with `EPERM` and is fatal, so the link-up scripts never run under a capability sandbox. Only re-root when already euid 0 (the setuid-installed case); otherwise the child keeps its uid and any ambient capabilities reach the scripts unchanged.

## Testing

Validated on an ipq807x router (aarch64, musl), pppd under procd/ujail on a live PPPoE uplink:

- pppd as an unprivileged user (uid 454) with `CAP_NET_ADMIN`/`CAP_NET_RAW` effective + ambient passes the gate, accepts `ifname` and `plugin pppoe.so`, and brings up a real PPPoE session (v4 + delegated v6)
- the ip-up/ip-down scripts execute as uid 454 and inherit the ambient set
- the same user with an empty capability set exits `EXIT_NOT_ROOT` with `<argv[0]> must be run as root or with CAP_NET_ADMIN`

The three ways a pppd can end up holding CAP_NET_ADMIN were then measured on the built binary — `privileged` is observable, since it flips `must_exist` on the system options file:

| how the capability was obtained | AT_SECURE | `privileged` |
|---|---|---|
| ambient set, unprivileged uid, no file caps (the sandbox case) | 0 | yes |
| `setcap cap_net_admin=ep`, invoked by a normal user | 1 | no |
| setuid-root, invoked by a normal user | 1 | no |

The setuid-root row also reduces to the pre-patch `uid == 0` through the `uid == geteuid()` term, so that case is unchanged by construction either way.

### DIFF
--- a/pppd/main.c
+++ b/pppd/main.c
@@ -401,7 +401,11 @@ main(int argc, char *argv[])
     umask(umask(0777) | 022);
 
     uid = getuid();
-    privileged = uid == 0;
+    /* Privileged options come from the invoking user's own authority, so a
+     * pppd that gained privileges from the exec itself must not hand them
+     * to whoever ran it. */
+    privileged = uid == 0 || (uid == geteuid() && !ppp_secure_exec() &&
+                              ppp_privileged());
     slprintf(numbuf, sizeof(numbuf), "%d", uid);
     ppp_script_setenv("ORIG_UID", numbuf, 0);
 
@@ -452,10 +456,11 @@ main(int argc, char *argv[])
     }
 
     /*
-     * Check that we are running as root.
+     * Check that we have the privileges needed to open the ppp device
+     * and configure interfaces.
      */
-    if (geteuid() != 0) {
-	ppp_option_error("must be root to run %s, since it is not setuid-root",
+    if (!ppp_privileged()) {
+	ppp_option_error("%s must be run as root or with CAP_NET_ADMIN",
 		     argv[0]);
 	exit(EXIT_NOT_ROOT);
     }
@@ -1975,9 +1980,11 @@ run_program(char *prog, char * const *args, int must_exist, void (*done)(void *)
     if (ret != 0) {
         fatal("Failed to change directory to '/', %m");
     }
-    ret = setuid(0);		/* set real UID = root */
-    if (ret != 0) {
-        fatal("Failed to set uid, %m");
+    if (geteuid() == 0) {
+	ret = setuid(0);	/* set real UID = root */
+	if (ret != 0) {
+	    fatal("Failed to set uid, %m");
+	}
     }
     ret = setgid(getegid());
     if (ret != 0) {
--- a/pppd/sys-linux.c
+++ b/pppd/sys-linux.c
@@ -107,6 +107,9 @@
 
 #if !defined(__GLIBC__) || __GLIBC__ >= 2
 #include <asm/types.h>		/* glibc 2 conflicts with linux/types.h */
+#include <sys/syscall.h>
+#include <sys/auxv.h>
+#include <linux/capability.h>
 #include <net/if.h>
 #include <net/if_arp.h>
 #include <net/route.h>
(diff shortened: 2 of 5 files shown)


pppd/sys-linux.c
```diff
@@ -2851,6 +2853,27 @@ ppp_registered(void)
...
+
+    if (geteuid() == 0)
+	return 1;
+
+    hdr.version = _LINUX_CAPABILITY_VERSION_3;
+    hdr.pid = 0;
+    if (syscall(SYS_capget, &hdr, data) < 0)
+	return 0;
+
+    return data[CAP_TO_INDEX(CAP_NET_ADMIN)].effective & CAP_TO_MASK(CAP_NET_ADMIN);
```
paulusmack: Possibly better not to assume that CAP_TO_MASK(CAP_NET_ADMIN) is inside the range of an int here. I suggest adding "!!" to the start of this expression.

pppd/main.c
```diff
@@ -452,10 +454,11 @@ main(int argc, char *argv[])
...
 
     /*
-     * Check that we are running as root.
+     * Check that we have the privileges needed to open the ppp device
+     * and configure interfaces.
      */
-    if (geteuid() != 0) {
-	ppp_option_error("must be root to run %s, since it is not setuid-root",
+    if (!ppp_privileged()) {
+	ppp_option_error("must be run as root or with CAP_NET_ADMIN to run %s",
```
paulusmack: perhaps clearer to say "%s must be run as root or with CAP_NET_ADMIN"

paulusmack (changes requested):
Mostly looks fine, just some minor comments.

paulusmack:
This looks quite interesting, thanks.

JuliusBairaktaris:
Hi thanks for the review!

paulusmack:
Looks good, the only thing remaining is that the man page (pppd/pppd.8) needs updating. Either I can do that in a later commit, or you can add it here, whichever you prefer.

paulusmack (approved):
Changes look good.

JuliusBairaktaris:
@paulusmack updated it. Thanks for the review

JuliusBairaktaris:
One more fix on top of the approval, sorry.

The file-capability guard I added last round reads every `getxattr()` failure as "no file capabilities", so it hands the privileged-option grant straight back out whenever the probe cannot answer: `/proc` not mounted, a filesystem without `security.*` xattrs, or a path component the caller cannot traverse.

Replaced it with `getauxval(AT_SECURE)` — the kernel's own verdict on whether this exec raised privileges. It needs no filesystem access, and it also covers setgid and LSM-driven secure exec, which the xattr probe did not.

A/B on the built binary, `setcap cap_net_admin=ep pppd` invoked by an unprivileged user (`privileged` is observable, since it flips `must_exist` on the system options file):

| | old (`getxattr`) | new (`AT_SECURE`) |
|---|---|---|
| `/proc` mounted | unprivileged | unprivileged |
| `/proc` unmounted | **privileged** | unprivileged |

The sandbox case is unaffected: ambient capabilities on a non-root uid with no file caps leave AT_SECURE clear, so `privileged` is still granted there.

Man page reworded to name setgid alongside setuid-root and file capabilities. Nothing else changed.

paulusmack:
Using AT_SECURE also seems less likely to have possible TOCTOU races.

Neustradamus:
@JuliusBairaktaris: Thanks for your work and thanks to @paulusmack for merging!

It replaces the @a-andreyev PR:
- https://github.com/ppp-project/ppp/pull/136

PR #610, opened 2026-08-14
Updates and security improvements for the EAP-TLS and PEAP code

paulusmack:
A few updates to SSL-related code in pppd, mostly pretty minor.

- Disable OpenSSL Engine support by default, since it causes warnings with recent OpenSSL versions.
- Print a better message when the OpenSSL config file can't be opened
- Make most EAP-TLS/PEAP options privileged, for security reasons.


paulusmack:
@jjkeijser @enaess comments welcome

jjkeijser (approved):
this looks fine to me

jjkeijser (approved):
Ack

jjkeijser (approved):
Better safe than sorry about these - so ACK

PR #612, opened 2026-08-18
Add a .editorconfig which sets the default editor config.

jkroonza:
If specific sub-folders needs different settings they can be overridden there by adding similar to those folders.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+root = true
+
+[*]
+indent_style = tab
+indent_size = 4
+tab_width = 8
+end_of_line = LF
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true


jkroonza:
@paulusmack maybe you're aware of sub-folders that use details different from this.  Managed to get this working on vim too, so should be able to not commit whitespace-only change on the dhcpv6relay plugin if this can please be merged.  I'm not sure if this will improve the display situation in github though, which seems to treat a tab as 4 spaces.

paulusmack:
Yes, using only spaces not tabs would be more robust. I'm putting off doing that until after we get the current set of security problems solved and the next release out, though.

jkroonza:
My personal preference is using a single-tab for leading indent, allowing each person to set his/her own shift width, so I normally use (in VIM):

```
 set tabstop=4
 set shiftwidth=4
```

Then a bunch of other unrelated irrelevant settings that works for me, most relevant of which is that I've got settings to *not* indent the case portion of a select, ie:

```
select (foo) {
case bar:
    code here;
    break;
...
default:
    code here;
}
```

Where vim would normally shift both the case and code portions another shift-width length.

I think it's worthwhile to merge this in the meantime, and then after the next release I think it would make sense to have a sequence of reformat commits into whatever format you reckon is best at the end of the day.

As it stands I was using .local.vimrc (somewhat unstably, had to invoke vim from the base project folder - which I also fixed after getting editorconfig up) in ppp project folder to override the tabstop value, I took a few minutes yesterday to get editorconfig working, and then set this up.  Seems to work well in existing code, and gives others (seems vscode, vim, emacs and a few others has support for editorconfig) the benefit too.  Think it makes sense to add this so that the intended coding style (although that may change in future) is clear.

jkroonza:
Thank you!

PR #613, opened 2026-08-20
Dhcpv6plugin improvements.

jkroonza:
General bunch of patches, some of which beef up possible buffer over-reads, and also improving various bits and pieces of validation.


Neustradamus:
@jkroonza: Thanks for your improvements and thanks to @paulusmack for merging!

PR #617, opened 2026-08-20
pppd: Erase PAP credentials explicitly

grey3228:
## Summary

This change adds an internal `ppp_explicit_bzero()` helper and uses it to erase PAP credentials after authentication.

The helper calls the system `explicit_bzero()` where available and uses volatile byte stores as a portable fallback.

## Rationale

`BZERO()` expands directly to `memset()`. When it is used to clear a local buffer immediately before the buffer goes out of scope, an optimizing compiler may remove the call as a dead store.

`check_passwd()` currently relies on `BZERO()` to erase both the peer-supplied PAP password and the matching secret read from the secrets file. If those calls are removed, the credentials can remain in stack memory after authentication has completed.

## Changes

- detect the availability of `explicit_bzero()` at configure time;
- add `ppp_explicit_bzero()` with a volatile-store fallback;
- erase the received password in the PAP packet buffer;
- erase the local password and secret copies in `check_passwd()`, including early-return paths;
- erase the temporary `word` and `lsecret` buffers used by `scan_authfile()`;
- add a unit test for the new helper.

The general-purpose `BZERO()` macro is intentionally unchanged because most of its callers use it for ordinary initialization rather than erasing sensitive data.

Found by Linux Verification Center (linuxtesting.org) with SVACE.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -94,6 +94,7 @@ AC_CHECK_SIZEOF(unsigned short)
 
 # Checks for library functions.
 AC_CHECK_FUNCS([    \
+    explicit_bzero  \
     fexecve         \
     mmap            \
     logwtmp         \
--- a/pppd/auth.c
+++ b/pppd/auth.c
@@ -1532,7 +1532,7 @@ check_passwd(int unit,
 		free_wordlist(opts);
 	    if (addrs != 0)
 		free_wordlist(addrs);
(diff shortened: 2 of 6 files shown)


paulusmack:
look good, thanks

Neustradamus:
@grey3228: Thanks for your work and thanks to @paulusmack for merging!

PR #618, opened 2026-08-21
Improvements to TLS certificate verification

paulusmack:


paulusmack:
Review appreciated (particularly @enaess @jjkeijser)

Also: should the default certificate verification mode be "name" rather than "none"?

Should "suffix" mode require that the part of the certificate commonName/subject before the suffix end with a dot?

jjkeijser:
ah this is new code to me ...  
For backwards compatiblity reasons I would say that 'none' is the safest choice.
When reading the code, I see the use of `strcmp()`  (i.e. NOT `strncmp()`) - is that something we really want?

paulusmack:
> When reading the code, I see the use of `strcmp()` (i.e. NOT `strncmp()`) - is that something we really want?

We do strcmp on tls_verify_method, which is the argument the user gave to the tls-verify-method command-line option, so it will be a well-formed C string. The code used to do strcmp() on the value returned by X509_NAME_get_text_by_NID(), but now (with this PR) we do memcmp.

PR #624, opened 2026-08-28
pppd/PEAP: Guard against error returns from BIO_read/SSL_read

paulusmack:
This adds checks on the calls to BIO_read() and SSL_read() in the PEAP code for zero or negative return values.  Without these checks it is possible for peap_response() to be called with len == -1, leading to a call to memmove() with negative length and consequent memory corruption.

This can be triggered by the peer (the authenticator) sending a second PEAP start packet after the first.  The second start packet is not rejected and causes the call to BIO_ctrl_pending() to be bypassed, leading to a call to BIO_read() when no data is pending, which returns -1.  To guard further against this, we now reject start packets in phase 2, and don't skip the BIO_ctrl_pending() call on a start packet.

Thanks to the Kimi Security Team for finding and reporting this.

### DIFF
--- a/pppd/peap.c
+++ b/pppd/peap.c
@@ -597,8 +597,11 @@ int peap_process(eap_state *esp, u_char id, u_char *inp, int len)
 	len -= hlen;
 
 	if (flags & PEAP_S_FLAG_SET) {
+		if (psm->phase == PEAP_PHASE_2) {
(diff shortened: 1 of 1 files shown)


paulusmack:
@jjkeijser @enaess review appreciated

PR #629, opened 2026-09-04
plugins: handle allocation failures in radius and winbind

grey3228:
Static analysis found unchecked `malloc()` and `strdup()` results in the RADIUS and winbind plugins. Allocation failures could therefore result in NULL pointer dereferences.

This change:

- checks the base64 output buffer allocation in the winbind plugin;
- checks both RADIUS realm server-list allocations before initialization;
- checks the structure and string allocations used for `avpair` options;
- frees partially allocated RADIUS objects before invoking `novm()`.

Out-of-memory conditions remain fatal through the existing `novm()` handler.

Found by Linux Verification Center (linuxtesting.org) with SVACE.

### DIFF
--- a/pppd/plugins/radius/radius.c
+++ b/pppd/plugins/radius/radius.c
@@ -183,10 +183,18 @@ plugin_init(void)
 static int
 add_avp(char **argv)
 {
-    struct avpopt *p = malloc(sizeof(struct avpopt));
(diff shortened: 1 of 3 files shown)


paulusmack:
Looks fine. The added free() calls seem a little pointless when the process is about to terminate, but I guess they make the analyzers happier.

PR #631, opened 2026-09-10
plugins/radius: avoid unnecessary own address lookups

anphsw:
rc_own_ipaddress() lookup in rc_find_server() only needed when client/server pair specified in "servers" file. Otherwise it triggers DNS lookup via gethostbyname() each plugin call.

### DIFF
--- a/pppd/plugins/radius/config.c
+++ b/pppd/plugins/radius/config.c
@@ -464,7 +464,7 @@ static int find_match (UINT4 *ip_addr, char *hostname)
 
 int rc_find_server (char *server_name, UINT4 *ip_addr, char *secret)
 {
-	UINT4	myipaddr = 0;
+	UINT4	myipaddr = 0xFFFFFFFF;
 	int             len;
 	int             result;
 	FILE           *clientfd;
@@ -483,8 +483,6 @@ int rc_find_server (char *server_name, UINT4 *ip_addr, char *secret)
 		error("rc_find_server: couldn't open file: %m: %s", rc_conf_str("servers"));
 		return (-1);
 	}
-
-	myipaddr = rc_own_ipaddress();
 
 	result = 0;
 	while (fgets (buffer, sizeof (buffer), clientfd) != (char *) NULL)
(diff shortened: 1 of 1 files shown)


pppd/plugins/radius/config.c
```diff
@@ -512,8 +510,11 @@ int rc_find_server (char *server_name, UINT4 *ip_addr, char *secret)
 				break;
 			}
 		}
-		else /* <name1>/<name2> "paired" form */
+		else if (myipaddr) /* <name1>/<name2> "paired" form */
 		{
+			if (myipaddr == 0xFFFFFFFF) myipaddr = rc_own_ipaddress();
+			if (myipaddr == 0) continue;
```
paulusmack, line 516: Doesn't this change behaviour? Looking at the original code, if myipaddr was 0, we could still match on the else branch of the if statement below (body starting at line 528), but with your change we skip that. (Not that I understand precisely what the else branch is doing.)

paulusmack (commented):
This seems to change behaviour more than just reducing the number of calls to `rc_own_ipaddress()`. Is that deliberate (or benign)? If so it should be mentioned in the commit message.

An alternative might be to change `rc_own_ipaddress()` so it caches negative results as well as positive, e.g., initialize `this_host_ipaddr` to 0xffffffff and use that as the "not yet looked up" marker.

pppd/plugins/radius/config.c
anphsw: According to syntax, <name1/name2> pair must contain pair of valid host addresses.
Value of "0" will not match anything, so better to skip it instead of looping via such records.
Also, rc_own_ipaddress() will return 0 on failure, so it's invalid and we must not check it against records anyway.

anphsw:
> An alternative might be to change `rc_own_ipaddress()` so it caches negative results as well as positive, e.g., initialize `this_host_ipaddr` to 0xffffffff and use that as the "not yet looked up" marker.

In one hand both 0 and 0xffffffff are invalid (excluding very rare artifical cases) host addresses in terms of network communications.
In other hand, both 0 and 0xffffffff are valid (but uselsess here) hostname lookup answers.

So isn't better not to change rc_own_ipaddress() return values, but use 0xffffffff on upper level as a flag?

PR #632, opened 2026-09-11
pppd: use OS randomness for authentication challenges

### LINKED ISSUE #616: Consider using a CSPRNG for authentication challenges
pppd currently uses the `rand48` family both for public protocol values and for authentication challenges.

[`magic_init()`](https://github.com/ppp-project/ppp/blob/master/pppd/magic.c) initializes the shared `rand48` state from a host-dependent value, the current time and the process ID:

```c
seed = get_host_seed() ^ t.tv_sec ^ t.tv_usec ^ getpid();
srand48(seed);
```

The same state is then used by both `magic()` and `random_bytes()`:

```c
u_int32_t
magic(void)
{
    return (u_int32_t)mrand48();
}

void
random_bytes(unsigned char *buf, int len)
{
    int i;

    for (i = 0; i < len; ++i)
        buf[i] = mrand48() >> 24;
}
```

`magic()` is primarily used for non-cryptographic values such as LCP Magic-Numbers. Using a conventional PRNG for an LCP Magic-Number is not by itself a security issue: the value is public and its purpose is to detect looped-back links.

`random_bytes()`, however, is used to generate:

- CHAP-MD5 challenges in [`pppd/chap-md5.c`](https://github.com/ppp-project/ppp/blob/master/pppd/chap-md5.c);
- MS-CHAP and MS-CHAPv2 authenticator challenges in [`pppd/chap_ms.c`](https://github.com/ppp-project/ppp/blob/master/pppd/chap_ms.c).

There are also direct uses of `drand48()` for authentication data:

- EAP-MD5 challenge bytes in [`pppd/eap.c`](https://github.com/ppp-project/ppp/blob/master/pppd/eap.c);
- the MS-CHAPv2 Peer-Challenge in [`pppd/chap_ms.c`](https://github.com/ppp-project/ppp/blob/master/pppd/chap_ms.c).

For example, EAP-MD5 currently fills its challenge as follows:

```c
while (--challen >= 0)
    *ptr++ = (u_char)(drand48() * 0x100);
```

## Why the shared PRNG state may matter

The challenge itself is public, so confidentiality is not the requirement. Authentication challenges are instead expected to be unique and unpredictable in order to prevent replay and preplay attacks.

The `rand48` family is a linear generator with 48 bits of state. An `mrand48()` result exposes 32 bits derived from that state, leaving at most 2^16 candidate states after one observed result. Additional outputs can be used to distinguish the remaining candidates and predict subsequent `drand48()` and `mrand48()` results.

pppd transmits outputs of `magic()` to the peer as LCP Magic-Numbers before authentication. More values may be generated during LCP negotiation, for example when [`lcp_nakci()` or `lcp_reqci()`](https://github.com/ppp-project/ppp/blob/master/pppd/lcp.c) selects another Magic-Number. Because these public values and the later authentication challenges consume the same `rand48` state, the LCP exchange appears to provide an active peer with information that could be used to recover the state and predict a subsequent CHAP or EAP challenge.

This also means that the usual limitation of one PPP interface per pppd process does not necessarily eliminate the concern: the relevant state exposure and challenge generation can occur on the same link, with LCP preceding authentication.

For CHAP-MD5, the response is calculated over the identifier, shared secret and challenge. [RFC 1994, Section 2.3](https://datatracker.ietf.org/doc/html/rfc1994#section-2.3) says that each challenge should be unique and unpredictable, and specifically describes the risk of an attacker tricking a peer into responding to a predicted future challenge and later using that response to masquerade as the peer.

A possible active preplay scenario would therefore be:

1. interact with pppd during LCP and observe public outputs derived from the shared `rand48` state;
2. recover the generator state and predict a future CHAP identifier and challenge;
3. obtain a response to those predicted values from the legitimate peer by acting as an authenticator;
4. use the response when completing authentication with the real pppd instance.

I have not implemented a proof of concept and do not claim that this is exploitable in every deployment. In particular:

- pppd must be configured to authenticate its peer with one of the affected methods; a remote peer cannot normally select the local authentication policy by itself;
- an active preplay attack would require access to the legitimate peer and synchronization with the authenticator;
- CHAP-MD5, EAP-MD5 and MS-CHAPv2 are legacy methods with additional known security limitations.

Nevertheless, the use of a predictable generator for authentication challenges seems avoidable while these methods remain supported. The shared state with public LCP values also appears to make this more than a purely theoretical concern about the choice of random-number API.

## Possible direction

Would it make sense to introduce a separate secure-random API backed by an operating system CSPRNG, for example:

- `getrandom()` on Linux;
- `arc4random_buf()` where available;
- `/dev/urandom` as a fallback with complete error handling?

Authentication challenge generation could fail closed if secure random bytes cannot be obtained, rather than falling back to `rand48`.

The existing non-cryptographic PRNG could remain in use for values that do not require cryptographic unpredictability, such as LCP Magic-Numbers, packet identifiers and randomized challenge lengths. The important part would be to stop deriving authentication challenges from that state.

A possible scope would be:

1. add a secure-random abstraction suitable for the platforms supported by pppd;
2. make `random_bytes()` use it for CHAP-MD5, MS-CHAP and MS-CHAPv2 authenticator challenges;
3. replace the direct `drand48()` use for EAP-MD5 challenge contents;
4. replace the direct `drand48()` use for the MS-CHAPv2 Peer-Challenge;
5. retain the existing PRNG only for non-security-sensitive values.

I would appreciate the maintainers' opinion on whether this analysis is correct, whether the shared state with LCP Magic-Numbers is considered relevant in the supported threat model.

grey3228:
Closes #616.

`random_bytes()` and the direct `drand48()` calls that produce CHAP, EAP-MD5 and MS-CHAPv2 challenge material all draw on the same 48-bit `rand48` state that `magic()` uses for LCP Magic-Numbers. Since the Magic-Numbers are sent to the peer before authentication, an active peer is handed outputs of the same generator state that will later produce the authentication challenge. 

This series adds a separate OS-backed source for authentication challenge contents and leaves the existing PRNG in place for everything that does not need cryptographic unpredictability.

## What changes

New `auth_random_bytes()` in `pppd/auth-random.c`:

* `getrandom()` with `GRND_NONBLOCK` where available. On Linux the raw syscall is used when libc has no wrapper, so older glibc still gets the kernel interface;
* `EAGAIN` (pool not yet initialised) fails the current authentication attempt. It does **not** fall through to `/dev/urandom`, which would defeat the point;
* `/dev/urandom` is used only when the interface is absent (`ENOSYS`), opened `O_NONBLOCK | O_CLOEXEC`, with full short-read and `EINTR` handling;
* On failure the output buffer is zeroed, so a partial OS result can never reach the wire;
* Nothing in this path can block, at any point.

Call sites converted: CHAP-MD5 challenge, MS-CHAP and MS-CHAPv2 authenticator challenges, EAP-MD5 challenge contents, MS-CHAPv2 Peer-Challenge.

Left on the existing PRNG, deliberately: LCP Magic-Numbers, CHAP and EAP packet identifiers, and randomised challenge lengths.

Failures are propagated through CHAP, EAP and PEAP so that no challenge or response is transmitted and no failed MS-CHAPv2 response is added to the response cache. Without the cache guard, a retransmission of the same challenge
would resend an all-zero response from cache without consulting the RNG again.

## Policy: fail closed, with a privileged opt-out

By default a failure to obtain secure randomness fails the authentication attempt. The new privileged `allow-insecure-random` option lets an administrator accept the legacy PRNG instead, with a warning logged each time the fallback is used. It is `OPT_PRIV`, so it cannot be set from a per-peer entry in a secrets file; a remote peer cannot request or negotiate the
fallback. There is a unit test for exactly that vector.

The option and the early-boot caveat for `/dev/urandom` on older Linux kernels are documented in `pppd.8`.

## Why not `RAND_bytes()`

It was suggested in the issue thread, and I ended up not using it:

* It does not address the concern that prompted the non-blocking design. OpenSSL 3.x seeds from `getrandom()` without `GRND_NONBLOCK` by default, so `RAND_bytes()` can block until the kernel pool is initialised — which is the
  early-boot hang scenario raised in the thread;
* Builds without OpenSSL would need the OS path regardless, so using `RAND_bytes()` would mean maintaining two implementations with different failure and blocking semantics.

Using the OS interface directly in both build configurations gives one code path with one set of semantics. Happy to revisit if you would rather have `RAND_bytes()` where OpenSSL is present.

## Why not `arc4random_buf()`

It has no recoverable error return and no non-blocking flag. On FreeBSD 12 and later it is backed by `getrandom()` and blocks until the pool is seeded, so it cannot implement either half of the policy above. The rationale is recorded in a comment in `auth-random.c`.

## API note

`ChapMS2()` changes from `void` to `int` so it can report a Peer-Challenge generation failure, and `chap_ms.h` is an installed header. In practice this is a source-level change only: callers that ignore the new return value are unaffected, and on common ABIs an already-compiled plugin keeps working. It still deserves a mention in the release notes.

The `struct chap_digest_type` plugin interface is **not** changed. `chap.c` zeroes the length byte before invoking generate_challenge` / `make_response`, so an out-of-tree digest that simply writes the length behaves exactly as before; the zero length is the failure signal.

`peap_process()` gains `PEAP_AUTH_FAILED` (-2), distinct from its existing -1, so a local randomness failure does not land in the branch that NAKs to `EAPT_TLS`. A local failure must not trigger a method downgrade.

## Behaviour note

If challenge generation fails during a CHAP re-challenge on an already established link, the link is torn down rather than retried. With `GRND_NONBLOCK` this is essentially unreachable after boot, since the CRNG stays initialised; the realistic failure mode is the `/dev/urandom` path under descriptor exhaustion. Say the word if you would prefer a retry bounded by `chap-max-challenge` instead.

## Tests

Six new `check_PROGRAMS`:

* `utest_auth_random`, `utest_auth_random_device`, `utest_auth_random_syscall`, `utest_auth_random_no_header` — the real selection code compiled against scripted `getrandom`/`syscall`/`open`/`read` results, covering `EINTR`, `EAGAIN`, `ENOSYS`, permission errors, short reads, EOF, partial fills that continue into the device, and the privileged fallback. The four variants exercise the libc wrapper, the raw Linux syscall, device-only operation, and the case where the libc symbol exists without its declaring header;
* `utest_auth_protocol` — CHAP, EAP and PEAP state machines driven with failing randomness, asserting that nothing is sent or cached, plus the success and retransmission paths;
* `utest_auth_options` — the real option parser, asserting that an unprivileged source cannot enable the opt-out.

## Not covered

* Tested on Linux only. Solaris picks up `getrandom(2)` from `<sys/random.h>` on 11.3 and later and otherwise falls back to `/dev/urandom`, but I have no Solaris system to confirm this on;
* Early-boot behaviour on a real system has not been exercised.

Per the discussion in #616 this is not proposed for 2.5.4.

## Noticed but left alone

Two pre-existing issues adjacent to the code touched here, not addressed in this series:

* `pppd/eap.c`, PEAP request handling: `if (esp->es_client.ea_state = eapListen)` is an assignment, not a comparison. The branch is reachable, since `peap_process()` returns -1 on retransmit and on handshake errors.
* `peap_finish()` does not free `psm->in_buf` / `psm->out_buf`, which `peap_init()` allocates.

Happy to send either as a separate patch.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -57,7 +57,9 @@ AC_CHECK_HEADERS([  \
     stdarg.h        \
     sys/dlpi.h      \
     sys/ioctl.h     \
+    sys/random.h    \
     sys/socket.h    \
+    sys/syscall.h   \
     sys/time.h      \
     sys/uio.h       \
     time.h          \
@@ -96,6 +98,7 @@ AC_CHECK_SIZEOF(unsigned short)
 AC_CHECK_FUNCS([    \
     explicit_bzero  \
     fexecve         \
+    getrandom       \
     mmap            \
     logwtmp         \
     strerror])
--- a/pppd/Makefile.am
+++ b/pppd/Makefile.am
@@ -54,6 +54,7 @@ pppd_include_HEADERS = \
 
 # Headers to be distributed, but not installed in /usr/include/pppd
 noinst_HEADERS = \
+    auth-random.h \
     chap-md5.h \
     crypto-priv.h \
     eap-tls.h \
@@ -66,6 +67,7 @@ noinst_HEADERS = \
 
 pppd_SOURCES = \
     auth.c \
+    auth-random.c \
     ccp.c \
     chap-md5.c \
     chap.c \
--- a/pppd/Makefile.am
+++ b/pppd/Makefile.am
@@ -26,6 +26,26 @@ utest_utils_LDFLAGS =
 
 check_PROGRAMS += utest_utils
 
+utest_auth_random_SOURCES = auth_random_utest.c
+utest_auth_random_CPPFLAGS = -I$(srcdir)
+utest_auth_random_device_SOURCES = auth_random_utest.c
+utest_auth_random_device_CPPFLAGS = -I$(srcdir) -DTEST_NO_GETRANDOM
+check_PROGRAMS += utest_auth_random utest_auth_random_device
+
+utest_auth_protocol_SOURCES = auth_protocol_utest.c utils.c crypto_ms.c
+utest_auth_protocol_CPPFLAGS = -I$(srcdir) -DUNIT_TEST $(OPENSSL_INCLUDES)
+utest_auth_protocol_LDADD = libppp_crypto.la
+check_PROGRAMS += utest_auth_protocol
+
+if LINUX
+utest_auth_random_syscall_SOURCES = auth_random_utest.c
+utest_auth_random_syscall_CPPFLAGS = -I$(srcdir) -DTEST_RAW_GETRANDOM
+check_PROGRAMS += utest_auth_random_syscall
+utest_auth_random_no_header_SOURCES = auth_random_utest.c
+utest_auth_random_no_header_CPPFLAGS = -I$(srcdir) -DTEST_NO_RANDOM_H
+check_PROGRAMS += utest_auth_random_no_header
+endif
+
 pkgconfigdir   = $(libdir)/pkgconfig
 pkgconfig_DATA = pppd.pc
 
@@ -193,6 +213,13 @@ pppd_LIBS += $(SYSTEMD_LIBS)
 endif
 
 pppd_LDADD = $(pppd_LIBS)
+
+# Use the daemon's actual option tables and parser without starting a link.
+utest_auth_options_SOURCES = $(pppd_SOURCES) auth_options_utest.c
+utest_auth_options_CPPFLAGS = $(pppd_CPPFLAGS) -Dmain=pppd_main
+utest_auth_options_LDFLAGS = $(pppd_LDFLAGS)
+utest_auth_options_LDADD = $(pppd_LDADD)
+check_PROGRAMS += utest_auth_options
 
 EXTRA_DIST = \
     ppp.pam
--- a/pppd/auth-random.c
+++ b/pppd/auth-random.c
@@ -1,0 +1,130 @@
+/*
+ * Randomness for authentication challenges.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <errno.h>
+#include <fcntl.h>
+#include <string.h>
+#include <unistd.h>
+#ifdef HAVE_SYS_RANDOM_H
+#include <sys/random.h>
+#endif
+#if defined(__linux__) && defined(HAVE_SYS_SYSCALL_H)
+#include <sys/syscall.h>
+#endif
+
+#include "pppd-private.h"
+#include "auth-random.h"
+#include "magic.h"
+
+/* Use OS interfaces with explicit error returns. arc4random_buf() has no
+ * recoverable error result or nonblocking flag, so it cannot implement our
+ * policy of failing authentication when randomness is unavailable.
+ */
+/* Linux kernels can provide getrandom even when libc has no wrapper. */
+#if defined(__linux__) && defined(SYS_getrandom)
+#ifndef GRND_NONBLOCK
+#define GRND_NONBLOCK 0x0001
+#endif
+#define auth_getrandom(buf, len) syscall(SYS_getrandom, buf, len, GRND_NONBLOCK)
+#endif
+#if defined(HAVE_GETRANDOM) && defined(HAVE_SYS_RANDOM_H) && defined(GRND_NONBLOCK)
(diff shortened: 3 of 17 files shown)


paulusmack:
Thanks for noticing the `state = eapListen` typo. As for the PEAP memory leak, I have a rewrite of the PEAP code almost ready to go, which removes that code, so I'm not going to worry about it for now.

Neustradamus:
@enaess, @jkroonza: What do you think?

